PR #20, opened 2018-01-27
Add permentropy and its tests

### LINKED ISSUE #18: Permutation entropy
Hi, I implemented the permutation entropy ([Bandt & Pompe, 2002](http://doi.org/10.1103/PhysRevLett.88.174102)).  Do you want add it to ChaosTools.jl?  Let me know if you want a PR.  Here is a demo: https://gist.github.com/tkf/31cb31564dc4f61c085d013656f1a9e9

```julia
function permutation_entropy(
        time_series::Array{T, 1}, ord::O, sampling_interval::S = 1;
        base = Base.e) where {T, O <: Integer, S <: Integer}
    n = length(time_series)
    perms = hcat(Combinatorics.permutations(O(1):ord)...)
    nperm = size(perms)[2]
    ptype = collect(O(1):ord)
    count = zeros(UInt64, nperm)

    for t in 1:n-sampling_interval * ord
        sample = time_series[t:sampling_interval:t+sampling_interval * (ord-1)]
        sortperm!(ptype, sample)
        for i in 1:nperm
            if perms[:, i] == ptype
                count[i] += 1
                break
            end
        end
    end

    nonzero = collect(c for c in count if c != 0)
    p = nonzero / sum(nonzero)
    return - sum(p .* log.(base, p))
end
```

I think this is not an optimal implementation though.  I just wrote it as a quick exercise a while ago.  Besides a few Julia related improvements (like using `@views`), I think the inner most loop that searches the permutation can be improved, using the fact that the permutation counter `perms` is sorted and maybe also by using better data structure.

Also, I'm not a data analysis guy so I don't know if this method is relevant these days.

Anyway, what do you think?  It's good to add it?

tkf:
closes #18 

Here is a demo: https://gist.github.com/tkf/31cb31564dc4f61c085d013656f1a9e9

### DIFF
--- a/src/dimensions/entropies.jl
+++ b/src/dimensions/entropies.jl
@@ -49,17 +49,17 @@ non0hist(ε::Real, matrix) = non0hist(ε, convert(Dataset, matrix))
 
 """
 ```julia
-genentropy(α, ε, dataset::AbstractDataset)
+genentropy(α, ε, dataset::AbstractDataset; base = e)
 ```
 Compute the `α` order generalized (Rényi) entropy [1] of a dataset,
 by first partitioning it into boxes of length `ε` using [`non0hist`](@ref).
 ```julia
-genentropy(α, p::AbstractArray)
+genentropy(α, p::AbstractArray; base = e)
 ```
 Compute the entropy of an array `p` directly, assuming that `p` is
 sum-normalized.
 
-*log base-e is used in both cases, i.e. units of "nat".*
+Optionally use `base` for the logarithms.
 
 ## Description
 The Rényi entropy
@@ -85,25 +85,25 @@ Statistics and Probability*, pp 547 (1960)
 
 [2] : C. E. Shannon, Bell Systems Technical Journal **27**, pp 379 (1948)
 """
-function genentropy(α::Real, ε::Real, data::AbstractDataset)
+function genentropy(α::Real, ε::Real, data::AbstractDataset; base=Base.e)
     ε < 0 && throw(ArgumentError("Box-size for entropy calculation must be ≥ 0."))
     p = non0hist(ε, data)
-    return genentropy(α, p)
+    return genentropy(α, p; base = base)
 end
-genentropy(α::Real, ε::Real, matrix) =
-genentropy(α, ε, convert(Dataset, matrix))
+genentropy(α::Real, ε::Real, matrix; base = e) =
+genentropy(α, ε, convert(Dataset, matrix); base = base)
 
-function genentropy{T<:Real}(α::Real, p::AbstractArray{T})
+function genentropy{T<:Real}(α::Real, p::AbstractArray{T}; base = e)
   α < 0 && throw(ArgumentError("Order of Rényi entropy must be ≥ 0."))
 
   if α ≈ 0
-    return log(length(p)) #Hartley entropy, max-entropy
+    return log(base, length(p)) #Hartley entropy, max-entropy
   elseif α ≈ 1
-    return -sum( x*log(x) for x in p ) #Shannon entropy, information to locate with ε
+    return -sum( x*log(base, x) for x in p ) #Shannon entropy, information to locate with ε
   elseif isinf(α)
-    return -log(maximum(p)) #Min entropy
+    return -log(base, maximum(p)) #Min entropy
   else
-    return (1/(1-α))*log( sum(x^α for x in p) ) #genentropy entropy
+    return (1/(1-α))*log(base, sum(x^α for x in p) ) #Renyi α entropy
   end
 end
 # Aliases:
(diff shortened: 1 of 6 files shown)


Datseris:
Looking good!!! Before we go into performance discussions, may I suggest a name change to `permentropy`, to have a more similar feel to `genentropy` ?

Datseris:
This is looking really nice, I can merge right away if tests pass.

Let's see this part now:


```
+    nonzero = [c for c in count if c != 0]
+
+    p = nonzero ./ sum(nonzero)
+    return  sum(p .* log.(base, p))
```
If you do `scale = sum(c)` and then do return sum( (p/scale) * log(p/scale) for p in c if p != 0)` wouldn't it be a lot faster?

I think `p .* log.(base, p)` already allocates, and of course its clear that `nonzero = [c for c in count if c != 0]` allocates as well,

src/dimensions/entropies.jl
```diff
@@ -112,3 +114,56 @@ shannon(args...) = genentropy(1, args...)
...
+        count[i] += 1
+    end
+
+    # To compute `p log(p)` correctly for `p = 0`, we first discard
+    # cases with zero occurrence.  They don't contribute to the final
+    # sum hence to the entropy:
+    nonzero = [c for c in count if c != 0]
+
+    p = nonzero ./ sum(nonzero)
+    return - sum(p .* log.(base, p))
```
Datseris: Will it not be much faster if one does
```
scale = sum(c)
return -sum( (p/scale)*log(p/scale) for p in c if p != 0)
```
? This version does no allocations, the original does 3 allocations, once when defining `nonzero`, once when defining `p` and once when doing `p .* log.(base, p)` (this also allocates a new array)

src/dimensions/entropies.jl
```diff
@@ -112,3 +114,56 @@ shannon(args...) = genentropy(1, args...)
...
+sliding windows between `t0` and `t1 = t0 + interval * (order - 1)`.
+
+## References
+
+[1] : Bandt, C., & Pompe, B., Phys. Rev. Lett. **88** (17), pp 174102 (2002)
+"""
+function permentropy(
+        time_series::AbstractArray{T, 1}, order::UInt8,
+        interval::Integer = 1;
+        base=Base.e) where {T}
```
Datseris: Having the "base" option is cool of course, but the thing is that all other entropies assume base `e`. Maybe we should add a keyword to all of them? 

(Or just use base-e everywhere, which is my suggestion)

tkf:
Yea it allocates memory but it executed only once.  For long time series I don't think it matters.

`perms = map(PermType, permutations(1:order))` likely allocates much more memory and doing some optimization for `sum(p log(p))` doesn't buy us match, unless we improve `perms`.  I wouldn't worry about it if I were you.

tkf:
Regarding further (memory/CPU) performance improvement around `perms` representation, probably using some kind of tree structure would help.  Maybe this?:
https://www.sciencedirect.com/science/article/pii/0898122194001782

But I'd say this implementation is OK for some moderate `order`.

Datseris:
The Julia people are super good on that stuff, if you care about it a lot you can post a question on Discourse and people will immediately help. I am sure there are special structures that can add a lot of performance.

At the moment it is not crucial for me and I can merge this and then open an issue about its performance.

What about the base thing?

src/dimensions/entropies.jl
tkf: I think `base=2` is also very popular.  In fact, `base=2` is used in the original permutation entropy paper.  I'd like to have base option everywhere.  Or something like `genentropy2`.

tkf:
I commented on base things as the review comment. Personally, I'd like to keep it and have base option everywhere.

Datseris:
> I commented on base things as the review comment. 

Can't find it! Regardless, you are right. Seems like the best option. I will do the change of including base on this PR, is that okay? I will also add my performance benefit, because even if it is small it still exists and does not reduce code clarity.

tkf:
I just wrote:

> I think base=2 is also very popular. In fact, base=2 is used in the original permutation entropy paper. I'd like to have base option everywhere. Or something like genentropy2.

tkf:
Re: performance

I like the saying "If you don't benchmark, you don't care the performance."  We are not doing any benchmarks here so talking about it is kind of nonsense.

OK. Not fair. I was the first one to bringing that up. But I knew that computational complexity of the innner most loop of the original code was `O(N)` where `N=order!` which was ridiculous.  I just switched to `searchsortedfirst` (which uses the binary search) since that would give us `O(log N)`.

tkf:
Probably setting up benchmarks using PkgBenchmark.jl would help us doing further performance improvement.

Datseris:
> Probably setting up benchmarks using PkgBenchmark.jl would help us doing further performance improvement.

Yeah sounds like an excellent plan. So far I have developed myself and benchmarked on the same computer with e.g. `@btime` but now is the time for upgrade.

src/dimensions/entropies.jl
```diff
@@ -112,3 +114,56 @@ shannon(args...) = genentropy(1, args...)
...
+    permutation_entropy(x::AbstractVector, order [, interval=1]; base=e)
+
+Compute the permutation entropy (Bandt & Pompe, 2002) of given `order`
+from the `x` timeseries.  Optionally, `interval` can be specified to
+use `time_series[t0:interval:t1]` when calculating permutation of the
+sliding windows between `t0` and `t1 = t0 + interval * (order - 1)`.
+
+## References
+
+[1] : Bandt, C., & Pompe, B., Phys. Rev. Lett. **88** (17), pp 174102 (2002)
```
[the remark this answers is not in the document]
Datseris: Daaaaamn how did I not consider to that everywhere!

Datseris:
Alright, I added base to `genentropy`. Assuming I did not have 10s of typos and tests pass, this is merged as-is. A separate issue will be created for its performance.

tkf:
Thanks!

BTW, it looks like Travis is failing

Datseris:
merged!

tkf:
Thanks!

Datseris:
thank you!

PR #33, opened 2018-02-19
Update for DynamicalSystemsBase 0.6

### LINKED ISSUE #7: Make the `lyapunovs` function return only the k-th larger exponents
_From @Datseris on July 19, 2017 12:7_

~~This is extremely simple:~~

The only changes necessary are:
1. in the methods add an optional argument (NOT a keyword) e.g. `k::Int = dimension(ds)`
2. initialize the matrices `Q` and `K` in the `lyapunovs` function to be `eye(k)`
3. For continuous systems make the function `tangentbundle_setup` also depend on `k` and initialize properly and define the e.o.m. properly
4. update the doc string
5. Pull request and BOOM! You contributed!

_Copied from original issue: JuliaDynamics/DynamicalSystems.jl#12_

### LINKED ISSUE #11: Relax time strictness in `lyapunov` for ContinuousDS
Currently, the method creates two integrators 
and then does:
```julia
    for τ in tvector
        # evolve until rescaling:
        push!(integ1.opts.tstops, τ)
        step!(integ1)
        push!(integ2.opts.tstops, τ)
        step!(integ2)
        dist = norm(integ1.u .- integ2.u)
        # Rescale:
        if dist ≥ threshold
            # add computed scale to accumulator (scale = local lyaponov exponent):
            a = dist/d0
            λ += log(a)
            finalτ = τ
            # Rescale and reset everything:
            # Must rescale towards difference direction:
            @. integ2.u = integ1.u + (integ2.u - integ1.u)/a
            u_modified!(integ2, true)
            set_proposed_dt!(integ2, integ1)
            dist = d0
        end
    end
```

However, I think this can be relaxed greatly. Instead of using `tstops`, one just "steps" the integrators normally. If the distance exceeds the threshold, we take the logarithm by interpolating at the nearest previous `dt`.

This will massively speed up the computation, because `tstops` is severely slowing an ODE solving. In addition, it won't affect the accuracy of the algorithm, since we will still be interpolating to the same point in time.

There is a caveat however, and that is to make sure that after the rescaling the integrators are also at the same time, in order to be sure that the method also works for non-autonomous systems.

### LINKED ISSUE #16: Separate lyapunovs into a setup and stepping part
(See the proposal here: #14 by @tkf )

In short, the idea is to separate the lyapunov function into a part that sets up stuff and a part that does orthonormalization steps. I think this is a good step that will have a lot of benefits: Firstly, it will be easy to solve #7 while. Secondly it will also make easier controlling #15 . It can also have other benefits like allowing using different qr methods or different methods at different steps, but I view these as very minor and they do not concern me.

Takafumi has already set up a great code that does what is described in the issue: https://github.com/tkf/LyapunovExponents.jl 

the problem is that at the moment the above is too complicated in order to offer these extra features, as for example there is no reason to associate `struct`s with the individual steps, and also the "relaxing" part is done in a single line in our package:
```
st = Ttr != 0 ? evolve(ds, Ttr; diff_eq_kwargs = diff_eq_kwargs) : ds.prob.u0
```
(I find it unwise to change something that needs 1 line into something that needs 200 ++ lines). Changing the transient integration can also be done with more arguments.

Maybe we can all learn from both repositories, the one here keeping things very simple, and the second one linked in this post, that offers a lot of capabilities but is too complicated, and then create something that is simple and offers a lot of capabilities (I am very optimistic that this is possible.

### LINKED ISSUE #17: Increase performance of PSOS by using ODEIntegrator instead of Callbacks
I thought of the following; Currently PSOS are produced by using a specific callback structure from diff eq.

However, Continuous callbacks are in general expensive and also have a purpose much more general, which makes them have a lot of options for nonlinear solve which are unrelated to PSOS but still create problems.

I propose to create a new internal function, like `create_psos` that uses an ODEIntegrator instead, and steps step by step, checking the condition at each step. by saving only the current and last state, one can use a loop like:

```julia
if ceil(x[i+1]) > ceil(x[i]) # condition for PSOS
        if vx[i] > 0
          # Linear interpolation
          yp = y[i] + (y[i+1] - y[i])*(ceil(x[i]) - x[i])/(x[i+1] - x[i])
          vyp = vy[i] + (vy[i+1] - vy[i])*(ceil(x[i]) - x[i])/(x[i+1] - x[i])
          push!(tempy, mod(yp, 1))
          push!(tempvy, vyp)
        end
      end
    end
```
(the example is from personal code library, where I only care to get the `y` and `vy` variables when `x` becomes zero modulo 1. Here, instead of `i` and `i+1` we can have previous and current state and update these in place. In addition, instead of interpolating each variable separately, one may be able to just interpolate the time of zero crossing by using the crossed variable, and then use the polynomial interpolation already offered by DiffEq.

### LINKED ISSUE #25: bug: no negative Lyapunov exponents for lyapunov(..) possible
Problem:
Because there is only an upper threshold but no lower bound, currently Lyapunov exponents can't be negative, but in reality, they can be.
Solution:
Just also add a lower threshold and do
 while upperthreshold < dist < upperthreshold

### LINKED ISSUE #26: Bug: avoid zero Lyapunov exponents for short run in lyapunov(..)
Problem:
Currently, if the Lyapunov exponent is small or the simulation time is small, dist ≥ threshold will never be true and therefore the returned Lyapunov exponent will be zero, although it is not actually zero.

Solution:
Add normalization at the end of the simulation.

### LINKED ISSUE #27: Bug: Missing warmup of orthonormal system in lyapunovs(..)
# Issue Description
The orthonormal system should be iterated along the trajectory before starting the simulation, otherwise, there will be a bias from the transient because initially, the Q is not aligned (e.g. the first column does not point into the direction of the first Lyapunov vector.)

### LINKED ISSUE #28: Bug: Missing warmup of perturbation vector in lyapunov(..)
The perturbation vector should be first iterated along the trajectory so it can align in the most unstable (or in case of stable systems in the least stable) direction. Otherwise, your resulting Lyapunov exponent has an error related to the initial transient of the vector.

### LINKED ISSUE #31: improve speed/accuracy: lyapunov(..): integrate perturbed and unperturbed trajectory simultaneously
Instead of separately iterating integ1 and integ2, they should be iterated together. Otherwise, the accumulate different noise along with their trajectories in every single step during the integration of the ODE. If they are iterated together (which means each time point is exactly the same), they would accumulate much less noise, therefore a larger reltol and abstol can be used. I have not tested this in ChaosTools.jl but I have seen a large difference in my own code (implemented unfortunately before ChaosTools came into existence).

Datseris:
our library is on fireeeeeeeeeeeeeee

Closes #31, closes #28, closes #27, closes #26, closes #25, closes #16, closes #17, closes #11, closes #7


RainerEngelken:
Thanks! Looks good, will test is more thoroughly after an upcoming conference.

PR #38, opened 2018-02-24
Make the Theiler window configurable

halleysfifthinc:
Like the title says, this is a pretty small tweak to allow for setting the Theiler window to something other than the delay.

It changes the obviously changes the arguments for the base `numericallyapunov` function, but the helper function with the keyword arguments doesn't require any different handling since `w` will default to `τ` which is the current behavior.


halleysfifthinc:
Also it looks like the latest version (v0.6.2) of DynamicalSystemsBase wasn't published to METADATA? (That is the cause of the CI build failures.

Datseris:
@halleysfifthinc Wow!!! I am so happy to see people using this library!
Awesome!

Thanks for the "fix". 

Yeah the tests were not passing because I made the mistake of incrementing `Base` to 0.6.2 instead of 0.7. Therefore the METADATA tests were not passing. I now released 0.7 in metadata and wait for the tag to pass.

PR #48, opened 2018-03-06
Remove type annotation in `delay_distance`

halleysfifthinc:
The type annotation prevented the use of `numericallyapunov` with multi-dimensional reconstructions.

### DIFF
--- a/src/nlts.jl
+++ b/src/nlts.jl
@@ -155,8 +155,7 @@ end
     abs(R[m+k][1] - R[n+k][1])
 end
 
-@inline @inbounds function delay_distance(di::Euclidean,
-    R::Reconstruction{D, T, τ}, m, n, k) where {D, T, τ}
+@inline @inbounds function delay_distance(di::Euclidean, R, m, n, k)
     return norm(R[m+k] - R[n+k])
 end
 
--- a/src/nlts.jl
+++ b/src/nlts.jl
@@ -13,7 +13,7 @@ export Cityblock, Euclidean
 
 """
 ```julia
-numericallyapunov(R::Reconstruction, ks;  refstates, w, distance, method)
+numericallyapunov(R::AbstractDataset, ks;  refstates, w, distance, method)
 ```
 Return `E = [E(k) for k ∈ ks]`, where `E(k)` is the average logarithmic distance for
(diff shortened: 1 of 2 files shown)


Datseris:
True but unfortunately I have not tested whether the method actually works for multi-dimensional Reconstructions.... *or* for reconstructions with multiple delay times.

Have played with it? This PR needs a test before it's merged.

halleysfifthinc:
It does work but I'm not sure what a good test would be (as far as which system, delays, and expected results)

Datseris:
Can you paste the code you used to see that it works? Maybe I can come up with some tests as well.

Standard tests would be to use systems where you already know the lyapunov exponent, e.g. take the towel map, take the first 2 timeseries, apply a reconstruction with delay 1 or 2 and see if it gives you exponent around `0.45`

halleysfifthinc:
```julia
ds = Systems.towel()
data = trajectory(ds, 10000)

taus = [0 0; 2 3; 4 6; 6 8]
data2 = data[:, 1:2]
R = Reconstruction(data2, 4, taus)

ks = 1:20
E = numericallyapunov(R, ks,
refstates = 1:1000, distance=Cityblock(), method=FixedMassNeighborhood(1))
λ = linear_region(ks, E)[2]
test_value(λ, 0.3, 0.5)
```

Datseris:
HOLY COW
IT WORKED

halleysfifthinc:
There is no reason it shouldn't have.. The underlying data and structure of the MDReconstructions is the same as normal reconstructions (afaik)?

Datseris:
yes yes you are right, I was just soooooo surprised that absolutely no change was needed! 

I now remember that I made the change to use `w = round(Int, mean(R.delay))` so its fine!

So I guess then we don't really need more tests for different systems, only the towel is enough.

halleysfifthinc:
haha I get that. :/ that moment when things actually go like you think they'll go.. :ok_hand: :ok_hand:

Datseris:
I've added the test. When tests pass I merge!

Datseris:
In fact, there is no reason to limit this to only `Reconstruction`.

Any dataset would work with the method, provided that the Theiler window is given if the dataset is not a `Reconstruction`!

PR #57, opened 2018-04-03
Finish mutual information delay estimation

### LINKED ISSUE #43: Add methods to estimate_delay
In particular, you can do:

- First zero of correlation function
- First minimum of correlation function
- Exponential decay of correlation function (currently implemented)
- minimum of the mutual information [1]

Also, maybe it is time to move estimate delay and dimension to DynamicalSystemsBase.jl ?

[1] : A.M. Fraser, H.L Swinney, Independent coordinates for strange attractors from mutual information, Phys. Rev. A 33 (1986) 1134--1140.

halleysfifthinc:
This should finish all the tasks in #43. closes #43

I made a pretty slight performance improvement; added some general documentation. In response to your question regarding the possible performance bump of using Euclidean distance, yes, Kraskov et al. did specify the maximum norm (ie Chebyshev distance).

I also made an API change (reordered k to be first argument instead of being a keyword argument; keyword arguments incur a performance cost as I understand) but since the `mutinfo` function wasn't previously enabled/finalized, I don't see any problem with it.

### DIFF
--- a/test/R_params.jl
+++ b/test/R_params.jl
@@ -34,7 +34,7 @@ test_value = (val, vmin, vmax) -> @test vmin <= val <= vmax
     x = data[:,1]
     @test 1.3 <= estimate_delay(x,"first_zero")*dt <= 1.7
     @test 2.6 <= estimate_delay(x,"first_min")*dt  <= 3.4
-    @test 0.9 <= estimate_delay(x,"mutual_inf"; maxtau=200)*dt <= 1.3
+    # @test 0.9 <= estimate_delay(x,"mutual_inf"; maxtau=200)*dt <= 1.3
 
     dt = 0.1
     data = trajectory(ds,2000,dt=dt)
@@ -42,7 +42,7 @@ test_value = (val, vmin, vmax) -> @test vmin <= val <= vmax
     @test 1.3 <= estimate_delay(x,"first_zero")*dt <= 1.7
     @test 2.6 <= estimate_delay(x,"first_min")*dt  <= 3.4
     @test 1.3 <= estimate_delay(x,"mutual_inf")*dt <= 1.7
-    @test 1.3 <= estimate_delay(x,"mutual_inf"; k=10)*dt <= 1.7
+    # @test 1.3 <= estimate_delay(x,"mutual_inf"; k=10)*dt <= 1.7
 
     ds = Systems.lorenz()
     dt = 0.05
--- a/test/R_params.jl
+++ b/test/R_params.jl
@@ -1,6 +1,19 @@
 using ChaosTools
 using Base.Test
 
+"""
+    saturation_point(x, y; threshold = 0.01, dxi::Int = 1, tol = 0.2)
+Decompose the curve `y(x)` into linear regions using `linear_regions(x, y; dxi, tol)`
+and then attempt to find a saturation point where the the first slope
+of the linear regions become `< threshold`.
+
+Return the `x` value of the saturation point.
+"""
+function saturation_point(Ds, E1s; threshold = 0.01, kwargs...)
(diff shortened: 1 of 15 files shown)


Datseris:
yeaaaaaaa very nice!

Keywords have a small performance hit, that is true. But this will go away mostly (if not entirely) in 0.7 Julia, so don't worry too much about it. In any case the current API is perfectly fine though.

Great I appreciate the tests a lot! For some weird reason Travis failed (didn't look at it) but now they re-run. Once they pass I will merge!

test/R_params.jl
```diff
@@ -12,26 +12,30 @@ test_value = (val, vmin, vmax) -> @test vmin <= val <= vmax
     @test estimate_delay(x,"first_zero") <= 2
     @test estimate_delay(x,"first_min")  <= 2
     @test estimate_delay(x,"exp_decay")  <= 2
+    @test 3 <= estimate_delay(x,"mutual_inf"; k=1) <= 10
+    @test 3 <= estimate_delay(x,"mutual_inf"; k=10) <= 10
```
Datseris: The optimal delay time for the Henon map (and actually most maps) is 1 and on can get away with 2. 

Do you have any idea why the mutual information method fails here? Is this expected?

Datseris: @halleysfifthinc

halleysfifthinc: Yeah, I noticed that too. Quote from Kraskov et al.:

> For practical applications, it means that one should use k > 1 in order to reduce statistical errors, but too large values of k should be avoided since then the increase of systematic errors outweighs the decrease of statistical ones. We propose to use typically k = 2 – 4, except when testing for independence. In the latter case we do not have to worry about systematic errors, and statistical errors are minimized by taking k to be very large (up to k ≈ N / 2, say).

I checked out a couple of different lengths of Henon map with a number of different values for `k` [here (jupyter gist)](https://gist.github.com/halleysfifthinc/32c55697e5ad7a3cb7b83867d497bb27). With increasing `k`, the first minimum appears to converge to 1 or 2.

For our purposes of finding the first minimum, we are in essence testing for independence, which suggests using their latter suggestion to minimize statistical errors by taking a large `k`.

Thoughts?

halleysfifthinc: Also, in the gist, I tried to test for higher `N` (`N=10e4`), as the mutual information function becomes exact as `N -> inf`, but I don't have enough memory (16Gb + 2Gb swap) and Julia kept crashing.

Datseris:
Another good suggestion could be to allow instead of `maxtau` that the users may pass an AbstractVector{Int} instead.

@halleysfifthinc I think it will be much faster if we talk about this on gitter as well. If you are ever around there you can tag me!

Datseris:
Another comment: Is there any "trivial" example, where you have 2 signals which the mutual information is known exactly? Because then this should also be a test and we can reduce the tests of `estimate_delay` by half (the one of the roessler system takes a tremendous amount of time :D)

I mean, the `mutinfo` is not tested directly at the moment.

test/R_params.jl
Datseris: Hm, but from the tests you have written, the mutual information method seems to require much more time and much more data points to converge than the other methods.

In addition, I think that `k` being 1000 or so is crazy high for real world applications. I mean with `k = 1000` you already have 1/10th of the signal as neighbors, if `N=10,000`

Datseris:
@halleysfifthinc I've spoken with someone here that knows a lot about information theory. The group leader of the team that made https://github.com/pwollstadt/IDTxl . They have a mutual information estimator. 

One of their backends is `jidt` https://github.com/jlizier/jidt which also has a mutual information. You can use one of the 2 to compute the mutual information between two known signals, like 2 cosines or whatever. The result can be used to test the current implementation as well.

PR #60, opened 2018-04-06
update for stochastic_indicator

JonasIsensee:
I changed the call signature to be consistent with `estimate_dimension` and added a docstring.


Datseris:
This is left open until tests are implemented.

Datseris:
ping @JonasIsensee 

(this is really easy, just write a test that indeed confirms what Cao has in his paper. Should be a paragraph of code at best)

PR #74, opened 2018-10-06
WIP: Add λmap

SebastianM-C:
Add a function for computing the maximal Lyapunov exponent for multiple trajectories.
Since there are several functions which could use something like this, I think that it would be nice to have a unified approach, but I'm not sure how to do this.

Missing things in this PR:
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/lyapunovs.jl
+++ b/src/lyapunovs.jl
@@ -2,6 +2,7 @@ using LinearAlgebra, StaticArrays
 using OrdinaryDiffEq
 using OrdinaryDiffEq: ODEIntegrator
 using DynamicalSystemsBase: MinimalDiscreteIntegrator
+using Distributed
 
 export lyapunovs, lyapunov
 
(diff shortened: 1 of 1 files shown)


src/lyapunovs.jl
```diff
@@ -265,6 +266,29 @@ function lyapunov(pinteg, T, Ttr, dt, d0, ut, lt)
     return λ/(pinteg.t - t0)
 end
 
+# TODO: Add docs and tests
+
+function λmap(ds::CDS, T, u0;
```
Datseris: There is no reason to limit this to `CDS`. Use `DS` instead (as this works identically for both discrete and continuous systems. They share `reinit!` interface).

Datseris:
Yeah you are right: many functions could benefit from this, so maybe it is better to make a generalized approach.

All functions do the following steps:

1. Accept a `DynamicalSystem` . 
2. Accept some arguments, some keyword arguments _and_ variable number of keyword arguments passed to DiffEq.
3. They initialize an integrator.
4. They call a low-level "interface" which accepts the integrator and it has no keywords.

The tricky part will be for this generalized approach to understand how to choose which arguments are which and which low level interface to call.

This _could_ be done with dictionaries with values various functions, but I guess it would be highly unstable...

SebastianM-C:
Should this be closed?
I recall that you implemented something more general.

Datseris:
I actually have not implemented parallelization for `DynamicalSystems`. I have done it for `DynamicalBilliards`, in a general way that works for different functions: https://github.com/JuliaDynamics/DynamicalBilliards.jl/blob/master/src/parallel.jl

I think `DynamicalSystems` won't be as easy (but you can still finish only the `λmap` and we can merge this).

If you want to work on general parallelization, also keeping in mind #83 , please go ahead. At the moment I certainly don't have time to spend on parallelization, unfortunately.

Datseris:
I am closing this, because I believe PRs that are opened and left open for years without progress make a bad impression on the status of a repo. If you want to work on this please feel free to re-open. (I close because I know I won't have time to work on this at the moment)

PR #79, opened 2018-12-04
Improved performance of non0hist

### LINKED ISSUE #58: Increase performance of `non0hist` [$100]
# The goal

Maybe we can make `non0hist` function faster, by using a specialized datastructure instead of the standard Julia dictionary. Or maybe there is an even better approach that doesn't use dictionaries at all (see below for the algorithm). I am really bad at performance optimization so when I wrote the function I did not really know whether there is a better plan. I think it is best to ask on Discourse.

# The current approach
`non0hist` produces a histogram in multidimensional space. However, it does it in a special, very cheap way. From the docstring:

> This method is effecient in both memory
> and speed, because it uses a dictionary to collect the information of bins with
> elements, while it completely disregards empty bins. This allows
> computation of entropies of high-dimensional datasets and
> with small box sizes `ε` without memory overflow.
> 

For reference, here is the current implementation:

(please recall: `Dataset` is nothing more than a wrapper of `Vector{SVector}`!)
```julia
function non0hist(ε::Real, data::AbstractDataset{D, T}) where {D, T<:Real}
    # Initialize:
    mini = minima(data)
    L = length(data)
    # Perform "Histogram":
    # `d` is a dictionary that contains all the histogram information
    # the keys are the bin edges indices and the values are the amount of
    # datapoints in each bin
    d = Dict{SVector{D, Int}, Int}()
    for point in data
        # index of datapoint in the ranges space:
        # It is necessary to convert floor to Int (representation issues)
        ind::SVector{D, Int} = floor.(Int, (point - mini)/ε)

        # Add 1 to the bin that contains the datapoint:
        d[ind] = get(d, ind, 0) + 1
    end
    return collect(values(d))/L
end
```

---

# The bounty

Any PR that _generally_ increases the performance of `non0hist` will be accepted as winning the bounty. You don't have to come up with orders of magnitude gain: something x2 gain is already massive, since the function is on the order of several `ms`.

EDIT: ALSO SEE THE CODE HERE: https://github.com/heliosdrm/RecurrenceAnalysis.jl/blob/master/src/delay.jl

<bountysource-plugin>

---
There is a **[$100 open bounty](https://www.bountysource.com/issues/56557834-increase-performance-of-non0hist?utm_campaign=plugin&utm_content=tracker%2F81670622&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F81670622&utm_medium=issues&utm_source=github).
</bountysource-plugin>

oneill:
Here is my proposal to optimize `non0hist` and resolve #58 .

The approach here is to replace each data point by its respective bin and sort the resulting list.
The histogram is then generated by counting consecutive equal bins in the sorted list of bins.

Note that the function will fail for an empty dataset, just like the original one.
Maybe it would be better to check that the dataset is not empty (and also that ε > 0).

Here are the results of the benchmark mentionned in #58:
Before:
```
ε = 0.1
  27.732 ms (14 allocations: 14.95 KiB)
with entropy = 4.762599382598514
ε = 0.01
  33.068 ms (37 allocations: 3.05 MiB)
with entropy = 9.471958080074165
ε = 0.001
  44.876 ms (43 allocations: 12.48 MiB)
with entropy = 11.463669293018897
```
After:
```
ε = 0.1
  11.732 ms (9 allocations: 4.20 MiB)
with entropy = 4.762599382598511
ε = 0.01
  13.904 ms (10 allocations: 4.34 MiB)
with entropy = 9.47195808007418
ε = 0.001
  13.468 ms (10 allocations: 4.93 MiB)
with entropy = 11.463669293019212
```

### DIFF
--- a/src/dimensions/entropies.jl
+++ b/src/dimensions/entropies.jl
@@ -26,20 +26,31 @@ function non0hist(ε::Real, data::AbstractDataset{D, T}) where {D, T<:Real}
     # Initialize:
     mini = minima(data)
     L = length(data)
-    # Perform "Histogram":
-    # `d` is a dictionary that contains all the histogram information
-    # the keys are the bin edges indices and the values are the amount of
-    # datapoints in each bin
-    d = Dict{SVector{D, Int}, Int}()
-    for point in data
-        # index of datapoint in the ranges space:
-        # It is necessary to convert floor to Int (representation issues)
-        ind::SVector{D, Int} = floor.(Int, (point - mini)/ε)
+    hist = Vector{Float64}()
 
-        # Add 1 to the bin that contains the datapoint:
-        d[ind] = get(d, ind, 0) + 1
+    # Reserve enough space for histogram:
+    sizehint!(hist, L)
+
+    # Map each datapoint to its bin edge and sort the resulting list:
+    bins = map(point -> floor.(Int, (point - mini)/ε), data)
+    sort!(bins)
+
+    # Fill the histogram by counting consecutive equal bins:
+    prev_bin = bins[1]
+    count = 1
+    @inbounds for i in 2:L
+        bin = bins[i]
+        if bin == prev_bin
+            count += 1
+        else
+            push!(hist, count/L)
+            prev_bin = bin
+            count = 1
+        end
     end
-    return collect(values(d))/L
+    push!(hist, count/L)
+
+    return collect(hist)
 end
 
 
--- a/src/dimensions/entropies.jl
+++ b/src/dimensions/entropies.jl
@@ -50,7 +50,10 @@ function non0hist(ε::Real, data::AbstractDataset{D, T}) where {D, T<:Real}
     end
     push!(hist, count/L)
 
-    return collect(hist)
+    # Shrink histogram capacity to fit its size:
+    sizehint!(hist, length(hist))
+
+    return hist
 end
 
 
--- a/src/dimensions/entropies.jl
+++ b/src/dimensions/entropies.jl
@@ -11,11 +11,9 @@ size `ε` and return the sum-normalized histogram in an unordered 1D form,
 discarding all zero elements.
 
 ## Performances Notes
-This method is effecient in both memory
-and speed, because it uses a dictionary to collect the information of bins with
-elements, while it completely disregards empty bins. This allows
-computation of entropies of high-dimensional datasets and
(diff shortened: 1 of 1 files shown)


src/dimensions/entropies.jl
```diff
@@ -26,20 +26,31 @@ function non0hist(ε::Real, data::AbstractDataset{D, T}) where {D, T<:Real}
...
+        else
+            push!(hist, count/L)
+            prev_bin = bin
+            count = 1
+        end
     end
-    return collect(values(d))/L
+    push!(hist, count/L)
+
+    return collect(hist)
```
Datseris: Is collect necessary here? `hist` is already a `Vector`, collect just make another vector now, correct?

Datseris:
yes

_yes_ **yeees** !!! Nice, thank you very much for this! Can I ask two questions for demographic purposes?

1. where did you find out about this issue?
2. Do you also use **DynamicalSystems.jl** ?

src/dimensions/entropies.jl
oneill: You are right, `collect` is not necessary here.
My intent was to shrink the storage of `hist` to fit its actual size.
I have updated this to use `sizehint!` instead.

oneill:
>     1. where did you find out about this issue?
>     2. Do you also use **DynamicalSystems.jl** ?

I found this issue on Bountysource.
I did not know about **DynamicalSystems.jl** before that, but I enjoyed working on this issue :thumbsup:

Datseris:
@oneill thanks very much for this! Don't forget to request the bounty!

src/dimensions/entropies.jl
```diff
@@ -26,20 +24,34 @@ function non0hist(ε::Real, data::AbstractDataset{D, T}) where {D, T<:Real}
...
+            push!(hist, count/L)
+            prev_bin = bin
+            count = 1
+        end
     end
-    return collect(values(d))/L
+    push!(hist, count/L)
+
+    # Shrink histogram capacity to fit its size:
+    sizehint!(hist, length(hist))
```
KristofferC: This  should be `resize!`

```
julia> a = [1,2,3,4];

julia> sizehint!(a, 2);

julia> length(a)
4
```

PR #89, opened 2019-06-05
WIP: Period estimation of one dimensional data using the autocorrelation function

### LINKED ISSUE #88: Feature request: period estimation
# Issue Description

Feature request for periodicity estimation

# Expected Behavior

An interface which can take in a `DynamicalSystem`, timeseries or `Dataset` and return its frequency (or period)

# MWE

```julia
# setup
using Pkg; pkg"add Hilbert"
using Hilbert


data = sin.(0:π/10:3π)

H = Hilbert.hilbert(data[:, :]) # the function expects a 2d array but we're giving it a 1d timeseries

ϕ₀ = arg(H[1]) # the initial angle

for i in eachindex(H)[2:end]
    # if the tangent vector has completed a full rotation
    arg(H[i]) ≈ ϕ₀ && println(i);
end
```

# Implementation

As discussed on Slack, credit to @Datseris:

>Okay, I think this is interesting to have, and you should open a feature request issue in ChaosTools about it. But I need to tell you that what you want has a clear definition in exactly periodic systems. When one considers quasi-periodicity the situation is not so clear anymore with respect to what you would call a "period time".
A valid approach is the Hilbert transform, and specifically the tangential vector to the vector of the Hilbert transform.
When the tangential vector has performed one rotation, then this is a good approximation for "one period"

There are two questions that one asks when dealing with suspected periodicity: 

1. What is the _period_ of my system? I.e. what is the temporal span that the system needs to return to the state it started at?
2. Is my system _periodic_ for the given parameters (and initial condition)?

These are two different questions. We have full support for question number 2.  This feature request is for question number 1.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/74917459-feature-request-period-estimation?utm_campaign=plugin&utm_content=tracker%2F81670622&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F81670622&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asinghvi17:
Introduces two new functions:
- `find_period(data; atol = 0.005, ptol = 5*atol)`
- `arg(z::Complex)`

Fixes #88 

See #88 for discussion before this pull request.


asinghvi17:
@Datseris, you may find this interesting:

> Methods to estimate the period of oscillations have been extensively discussed over the past two decades; they include the Whittaker-Robinson periodogram [4]–[7] that was popularized to study biological time course data by Enright [8]–[10], Fourier spectral analysis [4], [6], [7], [9], Lomb-Scargle periodogram [10], MESA [4], [5], [7], Autocorrelation [11] and cosinor [9], [10], [12]. All these methods are valid under different assumptions and may provide different results when applied to the same time course.

(from https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0093826)

asinghvi17:
This looks like a pretty interesting method, and there's code that we could probably adapt:

http://docs.astropy.org/en/stable/stats/lombscargle.html

Datseris:
From Slack:

an alternative method of period estimation that we could explore is the periodogram, specifically the Lomb-Scargle periodogram.  There’s a Python implementation in `astropy`, and I’m looking at at least using it for my own stuff using PyCall…

turns out that’s already in Julia: https://github.com/JuliaAstro/LombScargle.jl

asinghvi17:
I suppose that the question remains whether we should keep our own meta-wrapper, or just refer users to LombScargle.jl.  I'm in favour of creating our own internal wrappers, because we can extend them with functionality from other packages as well...

Datseris:
Alright, I just had a quick glance over the Julia package. It looks very promising and I am very happy to say that the developer @giordano is an excellent developer so I feel confident using their code.

I still believe we should wrap it anyways in some high level functionality that estimates period or periods, while maybe also estimating the "validity" of the estimated period.

giordano:
Let me know if there is anything I can do in `LombScargle.jl` that you need.  I should have implemented all methods present in Astropy, apart from those for multiple terms (I never had the need for them, nor the knowledge about how they work).

Datseris:
@asinghvi17 how does this new method apply to timeseries from the Roessler attractor? This is one good thing to measure, also as a unit test.

asinghvi17:
Naïve solution:
```julia
sys = Systems.roessler()
# 3-dimensional continuous dynamical system
#  state:       [0.718852, 0.969373, 0.102946]
#  e.o.m.:      roessler_eom
#  in-place?    false
#  jacobian:    roessler_jacob
#  parameters:  [0.2, 0.2, 5.7]
sol = trajectory(sys, 100)
using LombScargle
ls =  lombscargle(1:length(sol[:, 1]), sol[:, 1])
x = plot(sol[:, 1], xlabel = "t", ylabel = "x", title = "x-component", legend=false)
three = plot(sol[:, 1], sol[:, 2], sol[:, 3], title="Roessler attractor", xlabel = "x", ylabel = "y", zlabel = "z", legend = false)
periodogram = plot(freqpower(ls)..., xlabel = "f", ylapel = "P", title = "Periodogram", legend=false)
plot(three, x, periodogram, layout = @layout [a, b; c])
```

Roessler-trial

```julia
findmaxfreq(ls)
# 2.0071
```

asinghvi17:
```julia
julia> ls = lombscargle(1:length(sol[:, 1]), sol[:, 1], frequencies = 50:500)
LombScargle.Periodogram{Float64,UnitRange{Int64},UnitRange{Int64}}([.;..], 50:500, 1:10001, :standard)

julia> findmaxfreq(ls)
1-element Array{Int64,1}:
 258
```
which I think is a much more reasonable result.

asinghvi17:
Now this is happening (I forgot to provide a scaled time array earlier...)
```julia
using DynamicalSystems
sys = Systems.roessler()
# 3-dimensional continuous dynamical system
#  state:       [0.718852, 0.969373, 0.102946]
#  e.o.m.:      roessler_eom
#  in-place?    false
#  jacobian:    roessler_jacob
#  parameters:  [0.2, 0.2, 5.7]
T = 200
sol = trajectory(sys, T)
tr = LinRange(0, T, size(sol, 1))
using LombScargle
ls =  lombscargle(tr, sol[:, 1])
findmaxfreq(ls)
using Plots
x = plot(tr, sol[:, 1], xlabel = "t", ylabel = "x", title = "x-component", legend=false)
three = plot(sol[:, 1], sol[:, 2], sol[:, 3], title="Roessler attractor", xlabel = "x", ylabel = "y", zlabel = "z", legend = false)
periodogram = plot(freqpower(ls)..., xlabel = "f", ylabel = "P", title = "Periodogram", legend=false)
plot(three, x, periodogram, layout = @layout [a b; c])
```
Roessler2

Datseris:
Hello, 

So I am having trouble to review this PR, because I don't really know what `lombscargle` does... I don't have time to read the original papers, maybe someone can give a 1-minute summary?

My main problem is that I just try to find the sine frequency, using the code of @asinghvi17  :
```
x = sin.(0:0.01:6π)
ls = lombscargle(0:0.01:6π, x)
findmaxfreq(ls)

1-element Array{Float64,1}:
 200.15392781316348
```
I don;t really understand this number. What is it? I think I was expecting to get the frequency of sin(x), which is 1/2π, maybe scaled with `dt` (here 0.01). This is not what I get....

Datseris:
Okayyyy I finally get it. So, what we have to use @asinghvi17 is the `periodpower` function. Buuuut, there is a problem, we have to careful. Here is the `periodpower` plot for the Roessler attractor:

period1

and here is a zoom in the x axis:

period2

As you can see the first peak (with the strongest power) is indeed the correct period of the signal, at around 5-6 time. Now there are to quick problems:

* Due to finite precision / time / sampling the "period" power seems to increase as you tend to infinity.
* There is always a peak at 0 period (why? is this a bug @giordano  ?). We have to discard this.

After we discard the 0 peak, we should return the period that has the strongest power. Here that would be 5-6. I am not sure what the function `findmaxfreq` does, but this is not useful for us?

giordano:
Both findmaxfreq and findmaxperiod have an interval argument to narrow the search to the given region, see e.g. https://juliaastro.github.io/LombScargle.jl/stable/index.html#LombScargle.findmaxperiod  High power peaks at very low periods are common, especially with the fast approximated algorithm (the default one), that's why the lombscargle function has minimum_frequency and maximum_frequency keyword arguments.  Looking at the plot of the periidogram is almost always a good idea. Note also that you may get also peaks at different harmonics.

Have a look at the examples, I tried to cover a few situations 🙂

Datseris:
Thanks, these are very useful comments! :) I'll go through the examples in the meantime!

Alright, so what I think we need to do (totally optionally) is to make automations: a way to detect the strongest period correctly, a way to detect better bounds for the periodogram.

asinghvi17:
One thing we could do is set (configurable) restrictions on the frequency search - say, between 1/T and 0 (not including 0)?

giordano:
I didn't follow the discussion (this weekend I won't have much time), but I have a question: is the data evenly sampled or can it be unevenly sampled as well? In the former case, using LombScargle.jl can be an overkill, as this algorithm is specifically designed to cater for unevenly sampled data (extremely common in astronomy)and other methods can be used instead, like FFT, or some utilities in the DSP.jl package. If you want to be as general as possible (and so support also unevenly sampled data), LombScargle method is fine

Datseris:
In most cases the data are indeed evenly sampled, and yes some FFT could be more fitting (but if one can implement the general method, maybe it is worth it).

A bit off-topic: but @asinghvi17  I think we should still implement the Hilbert transform method as well. I will work on it a bit on Tuesday (I'll talk with Ulrich Parlitz then)

asinghvi17:
@Datseris, the more methods the better :) (although this PR will probably turn out to be rather long)

@giordano, the problem I face with FFTs in particular is that a lot of the data I'm using is decidedly not sinusoidal, and the result from the Fourier transform tends to be inaccurate.  I will look into `DSP.jl`, though - thanks!
giordano

asinghvi17:
Are you okay with adding dependencies on both `LombScargle.jl` and `DSP.jl`, @Datseris?  We could then integrate periodogram functionality from both of those.

Datseris:
I am indeed okay with these dependencies, but at the moment I don't see a reason to include both. What is the benefit DSP offers instead of LombScargle? I know that it is _simpler_ but this may not necessarily matter.

But regardless, I think we should split this PR into two: one that uses Hilbert transform and the other that uses the periodograms. (and you can do both to get the commit credits!)

asinghvi17:
```julia
# setup
using DynamicalSystems, LombScargle, DSP, BenchmarkTools

sys = Systems.roessler()

T = 100.0

sol = trajectory(sys, T)

tr = LinRange(0, T, size(sol, 1))
julia> @benchmark periodogram(sol[:, 1];  fs = size(sol, 1)/T) # DSP
BenchmarkTools.Trial: 
  memory estimate:  279.42 KiB
  allocs estimate:  79
  --------------
  minimum time:     145.356 μs (0.00% GC)
  median time:      286.223 μs (0.00% GC)
  mean time:        347.857 μs (15.51% GC)
  maximum time:     6.724 ms (85.66% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark lombscargle(tr, sol[:, 1]) # LombScargle
BenchmarkTools.Trial: 
  memory estimate:  54.99 MiB
  allocs estimate:  120631
  --------------
  minimum time:     172.961 ms (6.05% GC)
  median time:      206.538 ms (5.27% GC)
  mean time:        208.856 ms (7.97% GC)
  maximum time:     356.094 ms (43.37% GC)
  --------------
  samples:          24
  evals/sample:     1
```

Datseris:
I think this is _three_ orders of magnitude ;)

asinghvi17:
Whoops, was walking when I wrote that :P

asinghvi17:
Okay, I'll leave this for the Hilbert transform, and make another branch for the periodograms.  Let's let this lie here for now?  I'll put the periodogram handling stuff at the bottom of the file.

Datseris:
Yeah, I'll give more input here once I've talked with some of my colleagues.

giordano:
Yes, the Lomb-Scargle periodogram _is_ expensive, because it always assumes unevenly spaced data, that's why I asked before.  There is no specialised method for evenly spaced data, probably it would reduce to something like the `DSP.jl` periodogram.

The fast method of the `lombscargle` function internally calls the Fourier transform several times (three, if i remember correctly), you can cut down running time to almost one third by planning the periodogram (just like you can plan a FFT): on my machine the example above goes from 150 ms to 60 ms.

Datseris:
Alright, quick comment from a discussion with a collegue: 

Hilbert transform approach is nice because it gives an "instantaneous phase" (whose derivative is the frequency). So it is fitting for signals that:
* Are clean oscillations (in the sense that there exists one dominant frequency)
* Could change frequency in time
* The magnitude of the Hilbert transform gives the local amplitude

asinghvi17:
@Datseris, this should work out of the box, they're a modified version of the FitzHugh-Nagumo equations:

```julia
using DynamicalSystems, ParameterizedFunctions

FHN = @ode_def begin
    dv = min(max(-2 - v, v), 2 - v) - w
    dw =  e * (v - g * w + b)
end e b g

v0 = -2
w0 =-0.6667
u0 = [v0, w0]

g  = 0.8
e = 0.04
b = 0
p0 = [e, b, g]

T = 300.0

ts = LinRange(1000, 1250, 10000)
sys = ContinuousDynamicalSystem(FHN, u0, p0)
sol = trajectory(sys, T)
```

The data I was using was `sol[:, 1]`.

asinghvi17:
Here's how I envision the period estimation methods we'll have at this point:

methods copy

We can add more algorithms as it goes on, but this is approximately what we're going to do.

Datseris:
Darn! That's a cool cool picture! How did you make it?

asinghvi17:
Thanks!  It was made using `TikZ` :D

Datseris:
@asinghvi17 I am confused, why does `Hilbert.hilbert` require a 2D Matrix instead of a Vector?...

Datseris:
Yeah I agree. It is something too simply to rely on a package for it anyway. Did you write an implementation of it somewhere already or should I try to hack one?

asinghvi17:
I don't have an implementation unfortunately :(

Datseris:
Alright, I am on it.

Datseris:
@asinghvi17 I have finished this but I can't reproduce your dynamical system to test it.
This code:
```
function FHN(u, p, t)
    e, b, g = p
    v, w = u
    dv = min(max(-2 - v, v), 2 - v)
    dw = e*(v - g*w + b)
    return SVector(dv, dw)
end

g  = 0.8
e = 0.04
b = 0
p0 = [e, b, g]

fhn = ContinuousDynamicalSystem(FHN,SVector(-2, -0.6667),p0)
sol = trajectory(fhn, 1000.0; dt = dt)
```
gives
```
julia> sol
2-dimensional Dataset{Float64} with 100001 points
 -2.0  -0.6667
 -2.0  -0.667287
 -2.0  -0.667873
 -2.0  -0.668459
 -2.0  -0.669045
 -2.0  -0.669631
 -2.0  -0.670217
 -2.0  -0.670802
 -2.0  -0.671387
 -2.0  -0.671972
  ⋮
 -2.0  -2.5
 -2.0  -2.5
 -2.0  -2.5
 -2.0  -2.5
 -2.0  -2.5
 -2.0  -2.5
 -2.0  -2.5
 -2.0  -2.5
 -2.0  -2.5
```
so there is clearly something wrong with the dynamics. Can you please share the system in a non-`@ode_def` form that works? I don't want to add ParameterizedFunctions as a dependency.

Datseris:
Lol scratch that, sorry! I forgot the `-w` term!

Datseris:
@asinghvi17 will merge this when tests pass! Do you think you can finish soon the complementary PR that uses the periodogram, or I should release `1.7.0` without it? (no stress, just asking)

asinghvi17:
I can probably finish the periodogram PR in a couple of days...I just got off vacation so will give it some time later today and tomorrow. Thanks for finishing up AC so quickly!

PR #90, opened 2019-06-11
WIP: Period estimation with periodgrams

asinghvi17:

PR #91, opened 2019-06-28
Period estimation with periodograms

asinghvi17:
Supercedes #90


asinghvi17:
Here’s the error graph on a sine wave with controlled period, 19000 sample datapoints.

periodogram
welch

asinghvi17:
One thing I'm not sure about is whether it's worth checking whether the input array is evenly sampled.  I'm not sure how you would communicate that to the function, either, because a keyword argument is too easily missed, and there's no good default.

Datseris:
Maybe it is time to rethink whether we want to have a "default" method. I think this brings too much complication and maybe even risk on our side. At the moment I would stand that we should not have a default method, but the user chooses. We can describe when each method is good, but I think this covers the part that we should do.

asinghvi17:
Error benchmarking:
```julia
fs = 1:0.001:20

ps = fill(0.0, length(fs))

ts = 1:0.001:10

datas = [@. sin(ts * 2π * f) for f in fs]

@progress name="Estimating periods - pg" for data in datas

    ps[i] = _periodogram_period(data, ts)
end
pp = plot(1 ./ fs, @. (fs - 1/ps) / fs; xlabel = "Periods", ylabel = "Error", title = "Plain periodogram", legend = false)

@progress name="Estimating periods - Welch" for data in datas

    ps[i] = _welch_period(data, ts)
end
pw = plot(1 ./ fs, @. (fs - 1/ps) / fs; xlabel = "Periods", ylabel = "Error", title = "Welch", legend = false)

@progress name="Estimating periods - Autocor" for data in datas

    ps[i] = _ac_period(data, ts; L = length(ts) - 1)
end
pa = plot(1 ./ fs, @. (fs - 1/ps) / fs; xlabel = "Periods", ylabel = "Error", title = "Autocorrelation", legend = false)

@progress name="Estimating periods - Bartlett" for data in datas

    ps[i] = _bartlett_period(data, ts)
end
pb = plot(1 ./ fs, @. (fs - 1/ps) / fs; xlabel = "Periods", ylabel = "Error", title = "Bartlett periodogram", legend = false)

@progress name="Estimating periods - Multitaper" for data in datas

    ps[i] = _mt_period(data, ts)
end
pm = plot(1 ./ fs, @. (fs - 1/ps) / fs; xlabel = "Periods", ylabel = "Error", title = "Multitaper periodogram", legend = false)

@progress name="Estimating periods - LombScargle" for data in datas

    ps[i] = _mt_period(data, ts)
end
pl = plot(1 ./ fs, @. (fs - 1/ps) / fs; xlabel = "Periods", ylabel = "Error", title = "Lomb-Scargle periodogram", legend = false)

```

Datseris:
Is the autocorrelation method also slow? I wouldn't really expect this... Also, from your plots it seems like the autocorrelation and the plain periodogram fare best.

Also, I do not fully understand what is the x-axis in the plots you provide. Is it what is the actual period of the given signal? Why do the periodogram methods increase their error with the period? :S

asinghvi17:
What's going on here?
lombscargle

asinghvi17:
The x-axis is the period of the signal.  I tested from 0.05 to 1.  I'm not sure why the periodograms don't seem to do well on long periods though, could be "undersampling" in terms of number of waves?

asinghvi17:
Another problem - the periodograms are all failing tests, because their frequency bins are inaccurate at the timescale of the result.  You're always one bin above or below - that explains the oscillatory behaviour, I suppose.

asinghvi17:
actually wtf:
```julia
periodogram: Test Failed at /Users/anshul/.julia/dev/ChaosTools/test/period_tests.jl:15
  Expression: ≈(estimate_period(vs, alg, ts), trueperiod, atol=atol)
   Evaluated: 6.359035035824625 ≈ 6.283185307179586 (atol=0.05) 
welch: Test Failed at /Users/anshul/.julia/dev/ChaosTools/test/period_tests.jl:15
  Expression: ≈(estimate_period(vs, alg, ts), trueperiod, atol=atol)
   Evaluated: 8.74367317425886 ≈ 6.283185307179586 (atol=0.05) 
multitaper: Test Failed at /Users/anshul/.julia/dev/ChaosTools/test/period_tests.jl:15
  Expression: ≈(estimate_period(vs, alg, ts), trueperiod, atol=atol)
   Evaluated: 6.994938539407086 ≈ 6.283185307179586 (atol=0.05) 
lombscargle: Test Failed at /Users/anshul/.julia/dev/ChaosTools/test/period_tests.jl:15
  Expression: ≈(estimate_period(vs, alg, ts), trueperiod, atol=atol)
   Evaluated: 6.339449541284405 ≈ 6.283185307179586 (atol=0.05) 
Test Summary: | Pass  Fail  Total
simple sine   |    1     5      6
  ac          |    1            1
  periodogram |          1      1
  welch       |          1      1
  bartlett    |          1      1
  multitaper  |          1      1
  lombscargle |          1      1
```

asinghvi17:
So in theory, everything should be "working" now.  However, some of the methods seem to be...inaccurate.  More testing is in order, but I'll let Travis have at it for now.

asinghvi17:
Seems it was a problem of undersampling.  Dropping `dt` to `0.001` yields:

```julia
Test Summary: | Pass  Fail  Total
simple sine   |    3     3      6
  ac          |    1            1
  periodogram |    1            1
  welch       |          1      1
  bartlett    |          1      1
  multitaper  |    1            1
  lombscargle |          1      1

Test Summary: | Pass  Fail  Total
sin(2x)       |    3     3      6
  ac          |    1            1
  periodogram |    1            1
  welch       |          1      1
  bartlett    |          1      1
  multitaper  |    1            1
  lombscargle |          1      1

Roessler is stuck for some reason.
```

Datseris:
Okay, so half the tests fail?

asinghvi17:
Yep, consistently those three with `dt` lowered

asinghvi17:
@Datseris, one option is to drop `welch` and `bartlett`, which seem to be consistent underperformers.  On FHN the errors on the other two are low enough that I think it's justifiable to just increase the tolerance a wee bit.

asinghvi17:
I think I'll drop `welch` and `bartlett`, because they don't really seem to be that good.  We can leave the implementation in, or drop it.  I may also implement zero-crossing, for a sanity-check method...

asinghvi17:
Just added zero crossing (a naive implementation with a couple of optimizations to avoid too many allocations).  Now down to one test failure.

```julia
Test Summary:     | Pass  Fail  Total
simple sine       |    4     1      5
  autocorrelation |    1            1
  periodogram     |    1            1
  multitaper      |    1            1
  lombscargle     |          1      1
  zerocrossing    |    1            1
Test Summary: | Pass  Total
sin(2x)       |    5      5
Test Summary: | Pass  Total
Roessler      |    5      5
Test Summary:            | Pass  Total
Modified FitzHugh-Nagumo |    5      5
```

test/period_tests.jl
```diff
@@ -1,63 +1,73 @@
...
-@test p2 ≈  π   atol = dt
+    end
 
 end
+
+@testset "simple sine" begin
+    tsin = 0:dt:22π
+    vsin = sin.(tsin)
+
+    test_algs(vsin, tsin, 2π, 11.53dt)
```
asinghvi17: I'm not sure about this tolerance - it's the minimum needed for tests to pass, and the only thing that breaks if I go below this is LombScargle.

asinghvi17:
Different tests failing on different platforms?  Weird indeed...

Datseris:
@asinghvi17 I think the autocorrelation function computation also assumes that the data are evenly sampled. I think internally it uses FFT. Are you sure that it doesn't assume even sampling?

src/periodicity/period.jl
```diff
@@ -40,3 +142,122 @@ function _ac_period(v, t; ε = 0.2, L = length(v)÷10)
...
+
+"""
+    _periodogram_period(v, t; kwargs...)
+
+Use the fast Fourier transform to compute a periodogram (power-spectrum) of the
+given data.  Data must be evenly sampled.
+"""
+function _periodogram_period(v, t; kwargs...)
+
+    kwargs = Dict()
```
Datseris: What's going on here? Here `kwargs` is literally nothing, and then propagated into `periodogram`. Why? Should we just not give anything to `periodogram` if it takes no keywords? Why intermediatelly create an empty dictionary?

src/periodicity/period.jl
```diff
@@ -1,27 +1,129 @@
...
+    ]
+
+    methods = union(even_methods, other_methods)
+
+    if method ∉ methods
+        error("Unknown method (`$method`) given to `estimate_period`.")
     end
+
+    period = if method ∈ even_methods
+                faith || isevenlysampled(t) ||
```
Datseris: I am not sure about this `faith` keyword. Doesn't it over-complicated the API? Isn't it safer to anyway always check if the data are evenly sampled?

What is the rational behind `faith`?

src/periodicity/period.jl
```diff
@@ -40,3 +131,115 @@ function _ac_period(v, t; ε = 0.2, L = length(v)÷10)
...
+Find the zero crossings of the data, and use the average difference between
+zero crossings as the period.  This is a naïve implementation, with no
+interpolation; however, it's useful as a sanity check.
+
+The keyword `line` controls where the "zero point" is.
+
+The implementation of the function was inspired by [this gist](https://gist.github.com/endolith/255291),
+and has been modified for performance and to support arbitrary time grids.
+"""
+function _zc_period(v, t; line = 0.0)
```
Datseris: I think `line` should default to `mean(v)` instead. This is a good default estimate of where you should look for the zero crossing, assuming the timeseries is actually oscillating.

src/periodicity/period.jl
```diff
@@ -40,3 +131,115 @@ function _ac_period(v, t; ε = 0.2, L = length(v)÷10)
...
+    inds = findall(@. ≥(line, $@view(v[2:end])) & <(line, $@view(v[1:end-1])))
```
Datseris: Hm, this line here is pure Julia awesomeness! But I am concerned a bit for a user that is not super experienced with macros and how the `@.` operates. Having source code clarity in mind, maybe it is worth it to write it in a more explicit form?

totally optional, I'll leave it to you to decide.

asinghvi17: I'll add a couple of comment saw explaining what it does :)

src/periodicity/period.jl
asinghvi17: Will do, thanks!

src/periodicity/period.jl
asinghvi17: Oops, accidentally left that in there when testing...will remove.

src/periodicity/period.jl
asinghvi17: My thought process was that if you have a lot of data, say a very long Dataset, and you're getting it from, say, trajectory (which knows it's evenly sampled), you can skip a heavy process - on second thought, though, it only makes sense to do that when ts is an Array and not a Range, which would probably never come up.  I'll remove that one.

src/periodicity/period.jl
asinghvi17: I feel that the optimization from it is worth a little bit of opacity in the source code, and it's a good example for intermediate users IMO.

asinghvi17:
@Datseris, I don't think that the autocorrelation function uses FFT internally; it only does the dot product of vectors, looking through the source code.  I will try the autocorrelation function in Lotka-Volterra and get back to you.

asinghvi17:
Also, should we change the parameter from `varepsilon` to `epsilon`?  Just easier to type in IMO...

asinghvi17:
Autocorrelation seems to break down in its present state when faced with non-evenly sampled data.  It provides the wrong period.  I guess we should move it into evenly-sampled-only algorithms, then...

asinghvi17:
I'm happy with the current state of the PR.

asinghvi17:
With this PR, the available methods will look like this (maybe this should be in the docs?)
methods

asinghvi17:
Further progress can be made by implementing:
http://audition.ens.fr/adc/pdf/2002_JASA_YIN.pdf
https://en.wikipedia.org/wiki/Reassignment_method
http://www.audiocontentanalysis.org/code/pitch-tracking/compute-pitch/

src/periodicity/period.jl
Datseris: Alright, this is fine for me!

Datseris:
@asinghvi17 I am extremely happy with this PR, I'll merge soon (just increased tolerance of Roessler test to see if the APpVeyor will pass; if not I'll marge anyway). Good job, this is very good!

Yes, I'd like to have this figure in the documentation, it is very helpful. Do you think we can add two boxes next to the upper methods as well? 

For the zerocrossing we could use "average time distance between consequtive zero crossings. Fast but naive". For the autocorrelation we could say "Find when autocorrelation comes close to 1. Very accurate for large timeseries".

asinghvi17:
Figure for reference here:
methods

Datseris:
grrrrr the tests for multitaper always fail in 32-bit windows!

Datseris:
Hm, @asinghvi17  there seems to be a problem with multitaper at 32-bit windows:
```
multitaper: Test Failed at C:\projects\chaostools-jl\test\period_tests.jl:13
  Expression: ≈(estimate_period(vs, alg, ts), trueperiod, atol=atol)
   Evaluated: 1003.5149824250879 ≈ 91 (atol=0.8) 
```
this is not simply inaccurate, it is completely off. Maybe something goes wrong here? Do you know of the authors of Periodograms to ask?

asinghvi17:
The multitaper algorithm is from DSP.jl.  We could ask them - it seems @galenlynch is rather active on that repo.  However, it may be a problem in a library on `win32` as well.  I would rather keep multitaper, and add a note saying that it fails on win32.  We could selectively remove multitaper from the tests by reading the `PLATFORM` environment variable on Appveyor - that would go something like:

```julia
if haskey(ENV, "APPVEYOR") && ENV["PLATFORM"] == "someplatformspec"
    # methods without multitaper
else
    # methods with multitaper
end
```
or something like that.

galenlynch:
Hi! I had no idea this package existed, but it looks super useful for what I'm studying at the moment. I'm loving the documentation and feature set.

I'm not familiar with the code here, but it's surprising that there would be a difference in the result of `mt_pgram` between `win32` and other systems. If you think the error is coming from DSP.jl, it would be great if you could make a simple reproducible example so we can figure out what the problem is. Multi-taper periodogram estimation in DSP.jl relies on FFTW and probably OpenBLAS via the LinearAlgebra standard library, which could conceivably produce platform dependent results. The code for multi-taper periodogram estimation is [here](https://github.com/JuliaDSP/DSP.jl/blob/c3c5b029e404d20d8f9ad4428906239d414335f2/src/periodograms.jl#L401) and the code to calculate the Slepian tapers is [here](https://github.com/JuliaDSP/DSP.jl/blob/c3c5b029e404d20d8f9ad4428906239d414335f2/src/windows.jl#L522). We do not currently test on Windows, but maybe we should start doing that.

Datseris:
Hi @galenlynch . Thank you for your kind words, hope that **DynamicalSystems.jl** can help you! We also have a chat room `#dynamics-bridge` on Slack!

I won't have time to create MWE for this, maybe @asinghvi17  wants to dig deeper. 

> We do not currently test on Windows, but maybe we should start doing that.

Yes I highly recommend testing on Windows as well. There are many Windows-only Julia users (e.g. me!)

Issue #4, opened 2017-12-24
New method: definition of chaos by Ott [$100]


Datseris:
_From @Datseris on October 20, 2017 15:22_

Ott (et al) came up with a definitive way to claim whether or not a system is chaotic. This was done for the 25th anniversary of the journal Chaos.

In the article [Defining Chaos](http://aip.scitation.org/doi/10.1063/1.4922973) he and Brian R. Hunt define a quantity called "expansion entropy" which can characterize chaoticity.

It would be cool to have a function that computes that! The paper also includes a numerical algorithm to compute the quantity.

_Copied from original issue: JuliaDynamics/DynamicalSystems.jl#39_

yuxiliu1995:
I've understood the paper and is now learning the codebase. Should start coding soon.

Datseris:
Cool+awesome! My recommendation in these scenarios is to always open a WIP PR so that we can comment directly on code.

yuxiliu1995:
Now it's closed.

yuxiliu1995:
I got the bounty.
(Bountysource took 10% off, though. Consider not using that as a middleman?)

Datseris:
Sure, if you know alternatives that both take care of the legal side from my end, as well as ensure that the developer is guaranteed to get their money, let me know.

@yuxiliu1995 I am really conflicted on where I should put this in the docs. There are two places:

1. [Chaos Detection](https://juliadynamics.github.io/DynamicalSystems.jl/dev/chaos/chaos_detection/)
2. [Entropies and dimensions](https://juliadynamics.github.io/DynamicalSystems.jl/dev/chaos/entropies/).

By "name", it feels like it fits more in 2. But even though the method computes some kind of entropy, this quantity is mostly used for detecting chaos... What do you think?

yuxiliu1995:
It can go both places, but since it sounds like it's an entropy, so it probably goes to "Entropies and dimensions", since people would look up something about entropy at a place where entropies are listed.

asinghvi17:
Maybe put a reference to it in Chaos Detection as well?

Datseris:
Good ideas thanks. Once we have https://github.com/JuliaDynamics/DrWatson.jl/pull/127/ done I'll do this (at the moment it is not possible to build the docs of DynamicalSystems.jl because of a python incompatibility. really looking forward to dropping python all together, and once Makie is matured, also drop PyPlot).

PR #98, opened 2019-12-30
Ott chaos branch

yuxiliu1995:
WIP branch for https://github.com/JuliaDynamics/ChaosTools.jl/issues/4


yuxiliu1995:
I have successfully replicated the plots for two systems in the paper, one being 1-dimensional, and the other being 2-dimensional. Please check if they are good. If so, I would port the code into the package.

Datseris:
Hey, thats cool! (But this has to be implemented like a normal function, etc. etc. not as a notebook, I assume you know it).

I have some problem viewing the plots, can you please insert the plots here as a post?

yuxiliu1995:
It looks fine here https://github.com/yuxiliu1995/ChaosTools.jl/blob/ott-chaos/src/ott_chaos.ipynb

Figure 1
image
Figure 2
image
Figure 4
image
Figure 7
image

yuxiliu1995:
By the way, importance sampling can be quite useful for accelerating this feature
https://github.com/JuliaDynamics/ChaosTools.jl/issues/77

Datseris:
I'll need some time to review this, I gotta read the original paper again... I don't remember everything! :D

yuxiliu1995:
If I want to calculate $Df_{t0+10, t0}(x)$, where $f_{t0+10, t0}$ is the equation of motion defined by a set of ODE, is it faster to calculate it directly in one go, or is it okay to calculate it as a product of $Df_{t0+1, t0}(x)$, $Df_{t0+2, t0+1}(x)$, ... $Df_{t0+10, t0, 9}(x)$ ?

yuxiliu1995:
Also, I noticed an edge case in the [docs](https://juliadynamics.github.io/DynamicalSystems.jl/latest/ds/general/#discrete-in-place):

> *   **oop** : The `eom` **must** be in the form `eom(x, p, t) -> SVector` which means that given a state `x::SVector` and some parameter container `p` it returns an [`SVector`](http://juliaarrays.github.io/StaticArrays.jl/stable/pages/api.html#SVector-1) (from the [StaticArrays](https://github.com/JuliaArrays/StaticArrays.jl) module) containing the next state.
> *   **iip** : The `eom` **must** be in the form `eom!(xnew, x, p, t)` which means that given a state `x::Vector` and some parameter container `p`, it writes in\-place the new state in `xnew`.

When the state space is 1-d, this is false. The oop eom returns a Float64, and the iip eom doesn't seem to exist.

Datseris:
> If I want to calculate $Df_{t0+10, t0}(x)$, where $f_{t0+10, t0}$ is the equation of motion defined by a set of ODE, is it faster to calculate it directly in one go, or is it okay to calculate it as a product of $Df_{t0+1, t0}(x)$, $Df_{t0+2, t0+1}(x)$, ... $Df_{t0+10, t0, 9}(x)$ ?

I didn't understand this. How does $f$ have two arguments, both of which are time?

@yuxiliu1995 I'll start reviewing this once it is turned into code: please define a function (or as many as you need) in a separate file `src/expansionentropy.jl` while putting the rest of the code that tests/defines systems/plots in a dummy file in the `test` folder. I think it is useful to be able to comment directly on code, something not really feasible with Jupyter notebooks.

yuxiliu1995:
I managed to get everything working for discrete dynamical systems. For continuous systems, I'm figuring out how to get the Jacobian out of the ODE integrator. Should be finished by tomorrow.

yuxiliu1995:
And as for $f_{t0+T, t0}(x)$, that's the notation from (Ott and Hunt 2015). $f_{t0+T, t0}$ means the function that evolves the state space from time t0 to time (t0+T).

Datseris:
Ah I see, yeah I've checked the paper as well. Keep in mind that we don't really have access to this function, as it exists only in theory not in practice. In reality we only have access to the jacobian of the vector field (also called Equations of motion, also called ODE system), which is the function `f` of a dynamical system.

yuxiliu1995:
The new version runs correctly. I really tried making it work with `tangent_integrator` but it just doesn't work out.

Datseris:
> It is really inefficient, and if only tangent_integrator works, it would be much better.

Yes, and this is also the only way to merge it. The goal of DynamicalSystems.jl is to have crystal clear source code. Help me help you: what _precisely_ didn't work with `tangent_integrator`?

yuxiliu1995:
it's alright. I was just too frustrated.
i managed to get tangent_integrator working. Just one question remains.
I want to make the tangent_integrator check if the trajectory ever leaves the region [-1, +1]. The following code does not cause a `:Terminate` return code even if it seems aware of something happening at u=+1

```julia
using DynamicalSystems, LinearAlgebra, Statistics
using OrdinaryDiffEq: Vern9
using DynamicalSystemsBase, DifferentialEquations

exp_eom(x, p, t) = x*p[1]
exp_u0 = [0.1]
exp_p = [1.0]
exp_jacob(x, p, t) = [1.0]
exp_sys = ContinuousDynamicalSystem(exp_eom, exp_u0, exp_p, exp_jacob)

function terminate_condition(u, t, integrator)
    (u[1]+1)*(u[1]-1)
end
terminate_affect!(integrator) = terminate!(integrator)
terminate_cb = ContinuousCallback(terminate_condition, terminate_affect!)

tinteg = tangent_integrator(exp_sys, alg=Vern9(); callback=terminate_cb)

for i in 1:10
    @show tinteg
    step!(tinteg)
    @show check_error(tinteg)
end
```

returns
```
tinteg = t: 0.0
u: [0.1 1.0]
check_error(tinteg) = :Success
tinteg = t: 0.1685859623784646
u: [0.11836299701192027 1.1836299701192026]
check_error(tinteg) = :Success
tinteg = t: 0.6353272779775146
u: [0.1887639823325171 1.8876398233251708]
check_error(tinteg) = :Success
tinteg = t: 1.2415834705905566
u: [0.3461089661737782 3.4610896617377813]
check_error(tinteg) = :Success
tinteg = t: 1.8720331401458687
u: [0.6501501435029559 6.501501435029558]
check_error(tinteg) = :Success
tinteg = t: 2.30258509301496
u: [0.9999999999999943 9.999999999999943]
check_error(tinteg) = :Success
tinteg = t: 2.99914510089719
u: [2.0068373155506487 20.06837315550649]
check_error(tinteg) = :Success
```

Datseris:
Can you please push the change that uses `tangent_integrator`? Meanwhile I'll have a second look at callbacks and see what's happening here.

Datseris:
I have some problems with the package manager, no matter what I do I can't install `ChaosTools` in my home PC. I will need some time to do this before I can review this PR properly...

yuxiliu1995:
Sorry about the currently failed check. I don't know how to write Julia tests yet.

Datseris:
Don't worry about this, the tests are the easiest part of this :) 

You can check out the file https://github.com/JuliaDynamics/ChaosTools.jl/blob/master/test/entropy_dimension.jl to get you started. The way I envisioned the tests for this one is to use the function `linear_region` and test if the slope is the correct one.

Datseris:
@yuxiliu1995 can you justify why it is needed to step _exactly_ for `dt` at one point in the code? (BTW I've finally got chaostools working!)

Datseris:
@yuxiliu1995 can you justify why the continuous and discrete cases are handled differently? the code of `EESample` seems almost identical to me. I am pretty sure we can eliminate the two cases and treat them all in the same case, by adding some `@assert` that time must be integer or so in the discrete case.

Datseris:
@yuxiliu1995 when tangent_integrator can step with `SimpleATsit5` (the default integrator), please re-adjust the code to use argument_propagation to the integrator, so that differential equations can be fully configured. For example, this is done in `lyapunovs` (which also uses `tangent_integrator`) here: https://github.com/JuliaDynamics/ChaosTools.jl/blob/master/src/lyapunovs.jl  in lines 72 and 80.

yuxiliu1995:
> [@yuxiliu1995](https://github.com/yuxiliu1995) can you justify why it is needed to step *exactly* for `dt` at one point in the code? (BTW I've finally got chaostools working!)

To estimate 
∫ G(Df_{t0+dT, t0}(x)) dx, 
I need to sample many Df_{t0+dT, t0}(x_i), so each orbit of x_i must step exactly dT from t0.

> @yuxiliu1995 can you justify why the continuous and discrete cases are handled differently? the code of `EESample` seems almost identical to me. I am pretty sure we can eliminate the two cases and treat them all in the same case, by adding some `@assert` that time must be integer or so in the discrete case.

The two cases are indeed almost the same except at the stepping-time step.
Someone suggested that "We generally use dispatch rather than explicit type-checks"
https://discourse.julialang.org/t/tutorial-on-using-advanced-type-system-in-julia/32903/4?u=yuxi.liu

> [@yuxiliu1995](https://github.com/yuxiliu1995) when tangent\_integrator can step with `SimpleATsit5` (the default integrator), please re\-adjust the code to use keyword argument propagation to the integrator, so that differential equations can be fully configured. For example, this is done in `lyapunovs` (which also uses `tangent_integrator`) here: [https://github.com/JuliaDynamics/ChaosTools.jl/blob/master/src/lyapunovs.jl](https://github.com/JuliaDynamics/ChaosTools.jl/blob/master/src/lyapunovs.jl) in lines 72 and 80.

Is it also because `SimpleATsit5` is better than `Vern9`?

Datseris:
> Is it also because SimpleATsit5 is better than Vern9? 

No, in fact for some systems Vern9 is faster. The reason is (1) we don't add OrdinaryDiffEq as a dependency and (2) the compile time of SimpleATsit5 is dramatically lower than Vern9. Like 10 seconds difference. 

>  I can use explicit type-checks and combine these two functions into one.

You can still use dispatch. Have a look on how I do it for `lyapunov` :  https://github.com/JuliaDynamics/ChaosTools.jl/blob/master/src/lyapunovs.jl#L228-L309  . I still have one function for all kind of systems and I define some helper functions that dispatch on the type. You don't need that amount of complexity here (because `lyapunov` uses the parallel integrator), but I think you will get the point.

> I need to sample many Df_{t0+dT, t0}(x_i), so each orbit of x_i must step exactly dT from t0.

I still haven't read the paper in great detail, but are you sure you absolutely must sample exactly `dT` ? What happens if you sample approximately `dT`, by doing `step!(integ, dT)` ? I would bet the algorithm would work just as well, but we have to test it. In general letting the integrator step as much as it wants (without forcing the step size to be something exact) is more performant.

---

I've finally installed ChaosTools, once you do the changes discussed here I'll review more of the code directly!

Datseris:
ping @yuxiliu1995 I am waiting for you to do the aforementioned full tweaks before continuing with the review.

yuxiliu1995:
Oh sorry. I've been very busy with PhD applications.

Datseris:
No problem, don't worry. I was just making it clear in case you were waiting for me, I am not asking you to rush or anything.

Datseris:
BTW you should definitely add in your CV that you do contributions to DynamicalSystems.jl and other open source software.

yuxiliu1995:
New commit. 

* it doesn't depend on Vern9, and uses keyword propagation.
* doesn't depend on OrdinaryDiffEq 
* Two `EEsample` functions merged to one, although it seems to not worth the trouble to use any helpers.
* Integration step dT no longer enforced, although I left an option allowing it to be turned on. It seems to make no difference in the two test cases I gave it (Lorentz system, and exponential system).

Datseris:
Awesome, I'll give this a final round of review and then we are done here! :)

Project.toml
```diff
@@ -26,8 +26,8 @@ Combinatorics = "0.7, 1"
 DSP = "0.6"
 DiffEqBase = "5, 6"
 Distances = "0.7, 0.8"
-Distributions = "0.21"
-DynamicalSystemsBase = "1.5.1"
```
Datseris: You must keep version `1.5.1` of DynamicalSystemsBase because I've made an update that allows three-argument version for tangent integrator step, which simplifies the code here a lot.

Datseris:
@yuxiliu1995 I now see the reason for stepping for exactly `dt`: since we _average_ quantities at different time points, these quantities should be computed at the same time points.

yuxiliu1995:
> * Why EEsample returns this wonky thing instead of just a Vector?
>   Vectors are much more intuitive than 1xN matrices...

I have no idea. But now it returns a Vector, so thank you for that.

> * EEsample and EEgraph should also return the time vector

Done.

> * This PR needs a function `expansionentropy`, that takes in a dynamical system and all
>   other arguments and returns one number, the expansion entropy. This function would simply
>   use `linear_region` on the `EEgraph` result. The documentation
>   can then point to the other low level functions.

Done.

> * There is no test for regular continuous systems (where the expansion entropy should be zero)
>   you can use the lorenz system with high ρ for this.

A warning sign is found: regular continuous systems still has apparent positive expansive entropy:

```julia

lor = Systems.lorenz(ρ=69.75)

function boxregion(as, bs)
    @assert all(as .< bs)
    gen() = [rand()*(bs[i]-as[i])+as[i] for i in 1:length(as)]
    isinside(x) = all(as .< x .< bs)
    return gen, isinside
end

tr = trajectory(lor, 100.0, dt = 0.005, Ttr = 50.0)
x, y, z = columns(tr)

l_gen, l_isinside = boxregion(map(minimum, [x,y,z]), map(maximum, [x,y,z]))

@time times, meanlist, stdlist = expansionentropy_graph(lor, l_gen, l_isinside; batchcount=100, samplecount=100, steps=40, dT=1.0)
plot(times, meanlist, yerr=stdlist, leg=false)
```
image

I think the reason is the chaotic transient. With ρ=69.75, random orbits in the box tend to start out transiently chaotic for t~100, before eventually stabilizing. This is a typical example:

```julia
lor = Systems.lorenz(
    [-16.943553254756164, 25.930901961075335, 30.60965670087519 ]; ρ = 69.75)
tr = trajectory(lor, 40.0, dt = 0.005, Ttr = 0.0)

x, y, z = columns(tr)
plot(x,y,z,
        leg=false, title="Lorenz attractor", 
        html_output_format=:png, size=(1000,1000))
```
image

A much longer calculation (~100 seconds) indeed shows the line stabilizing:

```julia
lor = Systems.lorenz(ρ=69.75)

function boxregion(as, bs)
    @assert all(as .< bs)
    gen() = [rand()*(bs[i]-as[i])+as[i] for i in 1:length(as)]
    isinside(x) = all(as .< x .< bs)
    return gen, isinside
end

tr = trajectory(lor, 100.0, dt = 0.005, Ttr = 50.0)
x, y, z = columns(tr)

l_gen, l_isinside = boxregion(map(minimum, [x,y,z]), map(maximum, [x,y,z]))

@time times, meanlist, stdlist = expansionentropy_graph(lor, l_gen, l_isinside; batchcount=100, samplecount=100, steps=200, dT=2.0)
plot(times, meanlist, yerr=stdlist, leg=false)
```
image

This can be solved by a keyword for transience time `Ttr`, so that every sampled point is "aged" for `Ttr`. In effect, it is converting a box-uniform sampler into the [natural measure on the attractor](http://www.scholarpedia.org/article/Attractor_dimensions). 

(By the way, it's easy to plot the natural measure on a 2D attractor. Just evolve a lot of dots in its basin of attraction for a long time, and take all orbits, truncate their starting transient segments, then plot the points in a heatmap. If it seems useful, you should put it in an issue.)

And that's what I implemented. It seems to run twice as fast. You can try it:

```julia

lor = Systems.lorenz(ρ=69.75)

tr = trajectory(lor, 100.0, dt = 0.005, Ttr = 0.0)
x, y, z = columns(tr)

l_gen, l_isinside = boxregion(map(minimum, [x,y,z]), map(maximum, [x,y,z]))

@time times, meanlist, stdlist = expansionentropy_graph(lor, l_gen, l_isinside; batchcount=100, samplecount=100, steps=100, dT=2.0, Ttr=200)
plot(times, meanlist, yerr=stdlist, leg=false)
```

> * There is no test for multidimension discrete chaotic systems.
>   You can use the standard map for this.

I used Arnold's cat map.

> * All `@assert` commands must be made `@test` commands, otherwise nothing is tested.

Done

> * Please wrap the tests into individual testsets, I have demonstrated how to do this
>   for two examples

Done

> * Discrete 1-dimensional discrete systems can only be out-of-place. It is meaningless
>   to expect a 1d discrete system that is inplace. Please simplify `EEsample` for 1d
>   systems to use this knowledge (e.g. `j = jacob(x, p, t0 + t-1)`)

Done.

> * The function names `EEsample` and `EEgraph` must be changed. In Julia names starting
>   with capital letters signify types or modules. I have already changed the names.

I changed `EEgraph` to `expansionentropy_graph`.

> * The documentation strings should be made into the style of DynamicalSystems.jl
>   and also be simplified significantly. I will do this myself once the other comments
>   are taken care of.

Thank you.

Datseris:
> This can be solved by a keyword for transience time Ttr, so that every sampled point is "aged" for Ttr. In effect, it is converting a box-uniform sampler into the natural measure on the attractor.

Yes please add this `Ttr` argument if it doesn't take too much effort for you! Also your comments on the Lorenz in the stable regime actually appearing chaotic are tremendously useful for documentation, I'll be sure to include them.

Alright, awesome, as soon as I have some spare time I'll write the docs!

yuxiliu1995:
Now everything works, I left in two pieces of minor TODO in there that you might want to improve upon.

I also has a Jupyter notebook that you might want to use for the docs.

Datseris:
Thanks a lot, I'll take it from here, won't be long until this is merged! THe addition of the `box` functions is really good!

About the notebooks, feel free to add them here in e.g. a new folder `chaos`: https://github.com/JuliaDynamics/JuliaDynamics/tree/master/tutorials

src/chaosdetection/expansionentropy.jl
```diff
@@ -77,72 +146,116 @@ with its motion defined by ``f_{T+t0, t0}``. That is, for all ``t0 \\le t \\le t
...
+            reinit!(u_integ, u)
+            step!(u_integ, Ttr)
+            u = u_integ.u
+        end
+
+        reinit!(t_integ, u, t_identity) # Start integrating the tangents.
+
         for t ∈ 1:steps # Evolve the sample point for the duration [t0, t0+steps*dT]
-            step!(t_integ, dT, exactstepping)
+            # TODO: Temporary hack until DynamicalSystemsBase\src\discrete.jl is fixed.
```
yuxiliu1995: If you revert this part, and then calculate the entropy of Henon system (or any discrete 2D system), an error is thrown.

Try:
```julia
henon_iip = Systems.henon_iip(zeros(2); a = 4.2, b = 0.3)
henon_gen() = rand(2).*6 .- 3
henon_isinside(x) = -3<x[1]<3 &&  -3<x[2]<3
@time times, meanlist, stdlist = expansionentropy_graph(henon_iip, henon_gen, henon_isinside; batchcount=100, samplecount=100000, steps=24)
plot(times, meanlist, yerr=stdlist, leg=false)
```

and some error involving  `DynamicalSystemsBase\src\discrete.jl` would happen.

Datseris:
@yuxiliu1995 I am confused: why are there the arguments `batchcount` and `samplecount`? Is there any difference whatsoever between them? Both arguments configure how many initial conditions will be used, no?

Datseris:
@yuxiliu1995 your implementation of `_timeshift` is "wrong" as it doesn't actually evolve the system forwards in time, it just shifts the point of `t0`. The point of `Ttr` is to evolve the initial condition for this amount of time first. 

I'll try to fix/implement this  means that this will take longer to merge than expected (since, I initially thought I would only be writing documentation)

Datseris:
@yuxiliu1995 I got one more question: when is this `any(isnan, stdlist)` triggered in the function `expansionentropy`? So far I haven't seen it in any of the tests, and I also don't immediatelly imagine how the `stdlist` would get NaN values.

Datseris:
Phewwwww that was _exhausting_. All let this pass CI and merge tomorrow.

yuxiliu1995:
> EDIT: Sorry, I got confused. It turns out you correctly do the stepping via
> 
> ```julia
>         if Ttr > 0 # Evolve through transient time.
>             reinit!(u_integ, u)
>             step!(u_integ, Ttr)
>             u = u_integ.u
>         end
> ```
> in the sampling function. The `_timeshift` has no reason to exist therefore.

Actually, there is a reason for `_timeshift` to be there, since the tangent integrator still thinks the system is at `t0`, while after the transient time, it should really be at `t0+Ttr`.

Oh, I see. Is there a way to timeshift the tangent integrator at `reinit!`, so that its initial time is set to `t0+Ttr`?

> @yuxiliu1995 I got one more question: when is this `any(isnan, stdlist)` triggered in the function `expansionentropy`? So far I haven't seen it in any of the tests, and I also don't immediatelly imagine how the `stdlist` would get NaN values.

That would happen if all samples escaped the `isinside` region. It happens when the `samplecount` is too small.

Datseris:
> This timeshifting is necessary when the system depends on time. 

good point, I Think `reinit!` has keyword arguments for setting the starting time. If not, we should put one.

> That would happen if all samples escaped the isinside region. It happens when the samplecount is too small.

Okay, but I still want you to explicitly point me in code to the operation that would return `NaN` value. I couldn't find it.

yuxiliu1995:
> good point, I Think reinit! has keyword arguments for setting the starting time. If not, we should put one.

The [docs](https://juliadynamics.github.io/DynamicalSystems.jl/latest/advanced/#re-initializing-an-integrator) say that:
> *   `reinit!(integ, state)` : to be used with standard [`integrator`](https://juliadynamics.github.io/DynamicalSystems.jl/latest/advanced/#DynamicalSystemsBase.integrator).
> *   `reinit!(integ, Vector_of_states)` : to be used with the [`parallel_integrator`](https://juliadynamics.github.io/DynamicalSystems.jl/latest/advanced/#DynamicalSystemsBase.parallel_integrator).
> *   `reinit!(integ, state, Q0::AbstractMatrix)` : to be used with the [`tangent_integrator`](https://juliadynamics.github.io/DynamicalSystems.jl/latest/advanced/#DynamicalSystemsBase.tangent_integrator). This three argument version of `reinit!` is exported from `DynamicalSystemsBase`.

There is no mention of shifting time. I suppose it's time to add the functionality

> Okay, but I still want you to explicitly point me in code to the operation that would return NaN value. I couldn't find it.

At line `stdlist[t] = std(entropysamples)`, NaN gets returned by `std` when its input list has less than 2 items.

src/chaosdetection/expansionentropy.jl
```diff
@@ -66,7 +66,7 @@ function expansionentropy(system, sampler, restraining; kwargs...)
     times, means, stds = expansionentropy_batch(system, sampler, restraining; kwargs...)
     if any(isnan, stds)
         i = findfirst(isnan, stds)
-        println("Warning: All samples have escaped the given region at time = ", times[i])
+        println("Warning: All (or all except one) samples have escaped the given region at time = ", times[i])
```
asinghvi17:
```suggestion
        @warn "All (or all except one) samples have escaped the given region at time = $(times[i])."

File: src/chaosdetection/expansionentropy.jl [hunk 4]
```diff
@@ -209,8 +209,8 @@ function expansionentropy_batch(system, sampler, restraining; batches=100, steps
     for t in 1:steps
         entropysamples = filter(isfinite, @view eesamples[:, t])
         # remove -Inf entries, which indicate all samples failed to stay inside the given region.
-        if length(entropysamples) < 1
-            println("Warning: All samples have escaped the given region. Consider increasing sample or batch number. Terminating at step = ", t)
+        if length(entropysamples) ≤ 1
+            println("Warning: All (or all except one) samples have escaped the given region. Consider increasing sample or batch number. Terminating at step = ", t)
```
asinghvi17:
```suggestion
        @warn "All (or all except one) samples have escaped the given region.  Consider increasing sample or batch number. Terminating at step = $(t)."

File: test/expansionentropy_tests.jl [hunk 5]
```diff
@@ -0,0 +1,73 @@
+using ChaosTools, Test
+
+println("\nTesting expansion entropy...")
```
asinghvi17:
```suggestion
@info "Testing expansion entropy..."
```

src/chaosdetection/expansionentropy.jl
```diff
@@ -0,0 +1,220 @@
...
+[1] : B. & E. Ott, ‘Defining Chaos’, [Chaos 25.9 (2015)](https://doi.org/10/gdtkcf)
+=#
+export boxregion, expansionentropy, expansionentropy_batch, expansionentropy_sample
+using LinearAlgebra
+using Statistics
+
+"""
+    expansionentropy(ds::DynamicalSystem, sampler, restraining; kwargs...)
+
+Calculate the expansion entropy [1] of `ds`, in the restraining region ``S`` defined by
```
asinghvi17: Not sure, but should this be `[^1]` for footnotes?

yuxiliu1995: That's apparently the tradition in the module. I like footnotes too though.

Datseris: I don't know how footnotes work, I am ashamed to admit, which is why they are not used anywhere.

Datseris:
@yuxiliu1995 claim your prize!

PR #109, opened 2020-03-22
Add footnotes to docstring of numericallyapunov

asinghvi17:


Datseris:
? :P

asinghvi17:
The MD footnote syntax is `[^n]`, so these would previously not link to the footnote.  Hopefully, they will now...

Datseris:
Well, I will have to ask you to check this by building the docs before we can merge this... I understand it is convoluted, because you need master of DynamicalSystems.jl and ChaosTools.jl, but I think it is worth it to check. If this works, I will also do it everywhere I add references. 

(p.s.: if it works increase patch version of ChaosTools)

asinghvi17:
I think I found a hacky solution to test this which involves no local effort 😄.  Let's see if it works...

asinghvi17:
@Datseris I can confirm this works:
https://juliadynamics.github.io/DynamicalSystems.jl/previews/PR109/chaos/nlts/#ChaosTools.numericallyapunov

asinghvi17:
@Datseris I'm essentially done working on this branch; we could merge it now, but feel free to push to it if you want to get every docstring to have footnotes.

Datseris:
yes please let it open, I'll try to modify as many as possible. And also remove all `## References` headers.

Datseris:
@asinghvi17 can you please review and merge, to ensure I've used correct syntax?

src/chaosdetection/partially_predictable.jl
```diff
@@ -6,11 +6,11 @@ export predictability
     predictability(ds::DynamicalSystem; kwargs...) -> chaos_type, ν, C
 
 Determine whether `ds` displays strongly chaotic, partially-predictable chaotic
-or regular behaviour, using the method by Wernecke et al. described in [1].
+or regular behaviour, using the method by Wernecke et al. described in[1].
```
asinghvi17: Needs update

src/orbitdiagrams/poincare.jl
```diff
@@ -59,14 +61,7 @@ planecrossing = PlaneCrossing(plane, direction > 0)
 ```
 and `idxs` must be `Int` or `SVector{Int}`.
 
-## References
-[1] : H. Poincaré, *Les Methods Nouvelles de la Mécanique Celeste*,
```
asinghvi17: Had this been re-added (does it need to be)?

src/periodicity/period.jl
```diff
@@ -56,13 +56,13 @@ These methods tend to be slow, but versatile and low-error.
   It deffaults to `mean(v)`.
 
 For more information on the periodogram methods, see the documentation of
-`DSP.jl` and `LombScargle.jl`.
+[DSP.jl and LombScargle.jl.
```
asinghvi17: Extra [ in front

src/chaosdetection/partially_predictable.jl
Datseris:
```suggestion
or regular behaviour, using the method by Wernecke et al. described in[^Wernecke2017].
```

src/periodicity/period.jl
Datseris:
```suggestion
DSP.jl and LombScargle.jl.
```

src/orbitdiagrams/poincare.jl
Datseris: not needed

PR #125, opened 2020-09-17
Changed `e` to `Base.MathConstants.e`

apbraun:
In some definitions of `genentropy` and `permentropy` Euler's number was written as `e` instead of `Base.MathConstants.e`.


Datseris:
Thanks, please increment bug version in `Project.toml` as well so we can release a new version directly.

apbraun:
Is the bug version the last digit of the version number?

Datseris:
Yeap! Officially it is called a PATCH number, see here: https://semver.org/

PR #126, opened 2020-09-24
Added estimation of confidence limits for Takens

ikottlarz:
- Takens' estimator now returns 95%-confidence limits from the MLE (for background on estimation, I used Roger Barlow's "Statistics - A Guide to the Use of Statistical Methods in the Physical Sciences"). 
- Introduced optional parameter epsilon_min, as suggested by Borovkova 1999, defaults to zero as in "classic" Takens estimator, might introduce bias if set to other values but can potentially prevent total divergence


src/dimensions/correlationdim.jl
```diff
@@ -139,23 +139,50 @@ D_C \\approx - \\frac{1}{\\eta},\\quad \\eta = \\frac{1}{(N-1)^*}\\sum_{[i, j]^*
...
+    α = -(n-1)/η
+    # biased version (maximum of original log-likelihood function)
+    α_b = -n/η
+    # value of maximum of original log-likelihood function
+    mxl = n*log.(α_b) .+ α_b * η
+    # at the 95%-confidence interval, the log-l function has dropped by 2
+    # -> log_l(x) - mxl + 2 = 0
+    # this is a result of the invariance of the MLE, a really nice property
+    # THIS IS NOT GOING TO BE PERFECTLY SYMMETRIC, AS THE CLT DOES NOT APPLY!
+    mn, mx = fzeros(x-> n * log.(x) .+ η * x .- mxl .+2 , 0,2*α)
```
Datseris, line 180: Why is broadcasting used here? Isn't all participating things real numbers? like `α_b` seems to be real. Same with the definition of `mxl`?

Datseris:
Hi, thanks,

Here you are introducing a breaking change: you are changing the amount of returned arguments of the function. (I bet this is why tests fail). Breaking changes should be avoided unless absolutely mandatory.

Looking at the source, I can see that the estimation of the confidence intervals can be an independent function, like, `takens_confidence` or so. Correct?

src/dimensions/correlationdim.jl
Datseris: Also, why is that comment in full capitals...?

ikottlarz: > Why is broadcasting used here? Isn't all participating things real numbers? like α_b seems to be real. Same with the definition of `mxl`?

Yap, that was a relic of when I wanted to do this differently, my mistake. No broadcasting needed anymore.

> Also, why is that comment in full capitals...? 

Because in Takens' paper it is pretended/ implied that the error f this estimate is going to be a Gaussian standard deviation, which I find is a huge difference to this, especially interpretation wise. But I can see how it seems a little exaggerated

ikottlarz:
> Looking at the source, I can see that the estimation of the confidence intervals can be an independent function, like, `takens_confidence` or so. Correct?

Yes, technically it could be an independent function. My reasoning was that the computationally expensive/ time consuming part, the computation of η, is needed for both parts. If put into two separate functions, this would more or less double the computation time, which is not super relevant if you estimate a dimension once, but does have some impact if you do a parameter screening over like 500 values of εmax, I think. Which is something that you would potentially want to do with this estimator. Right?

Datseris:
damn, you are right, I didn't pay close enough attention. You need both `η, n`. Okay, then the only way to go forth here is to return all three.

But you still have to modify the test suite so that tests pass.

ikottlarz:
> But you still have to modify the test suite so that tests pass.

Okay, so I think I did that (see code, last commit) but the tests still fail (`MethodError: no method matching isless(::Float64, ::Tuple{Float64,Float64,Float64})`, that is due to the different amount of returned arguments, which I _did_ change in the test. What did I do wrong here?

Datseris:
Well, there are two problems:

1. You didn't change the _existing_ tests https://github.com/JuliaDynamics/ChaosTools.jl/blob/master/test/dimensions/entropy_dimension.jl#L101
2. You haven't yet included your new file into the actual test suite, since you haven't altered the file https://github.com/JuliaDynamics/ChaosTools.jl/blob/master/test/runtests.jl

What I'd suggest to do is:

* move all taken related tests (yours and existing) into the file you created 
* Make sure tests are fixed
* include the file you created in the test suite
* Add CHANGELOG.md entry
* Increment minor version in Project.toml

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "ChaosTools"
 uuid = "608a59af-f2a3-5ad4-90b4-758bdf3122a7"
 repo = "https://github.com/JuliaDynamics/ChaosTools.jl.git"
-version = "1.15.1"
+version = "1.16.1"
```
Datseris, line 4: Ha, you have to make it `1.16.0`!

Datseris:
I see now that the actual CI values are not tested. Shouldn't you test that e.g. the CI values are reasonably close to the computed value?

ikottlarz:
I see, that makes sense. I will add that although - what would be "reasonably close"? Relative errors of 10%? 20%?

Datseris:
That is something found out empirically, because no analytic formulas can help here for a given arbitrary chaotic set. Just look at the numbers and add a slightly larger interval for numeric stability.

ikottlarz:
Alright, I set it to a maximum of 5% relative error (mostly it is <1%, at least for Lorenz and Henon). Locally, the tests passed, so this should hopefully work now.

Also, thanks for all the help with this! :-)

PR #127, opened 2020-09-24
Mean return times

Datseris:
Continuous version coming soon!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 1.16
+* `transit_time_statistics` deprecated in favor of `exit_entry_times`.
 # 1.15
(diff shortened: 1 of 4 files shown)


Datseris:
holy crap this was ridiculously hard but I think I got it.

Datseris:
Hm, actually, nope.. The continuous case doesn't work excellently. At first, using `m=Inf` (which means that every step is checked explicitly for crossing) leads to very weird results... Secondly, `c` should always be sorted reverse, so that the innermost ε-sets have at most the same recurrence rate as the outer ones.. .But this is not what the numbers return... :( 

perhaps I really have to make a unified appraoch for all ε-sets as on like in the discrete case? :(

Issue #118, opened 2020-06-17
Enhanced box and prism assisted algorithms for computing the correlation dimension


Datseris:
In this work:

> Bueno-Orovio, A., & Pérez-García, V. M. (2007). Enhanced box and prism assisted algorithms for computing the correlation dimension. Chaos, Solitons and Fractals, 34(2), 509–518. https://doi.org/10.1016/j.chaos.2006.03.043

The authors provide the most optimized (to our knowledge) method for estimating the correlation sum and subsequent dimension.

apbraun:
I would assume this done.

PR #129, opened 2020-10-04
Add an implementation for q-order-correlationsum

apbraun:
I've implemented the q-order correlationsum introduced by Kantz in chapter 11.3. The tests will be included soon.


apbraun:
I've not come accross other versions in a short literature review, but I'm sure that calculation time can be cut down by first dividing the data into boxes and only taking into account neighbouring boxes as done by Theiler in [Efficient Algorithm for Estimating the Correlation Dimension from a Set of Discrete Points](https://www.researchgate.net/publication/13390536_Efficient_Algorithm_for_Estimating_the_Correlation_Dimension_from_a_Set_of_Discrete_Points) for the most used case of `q = 2`.

Datseris:
Yes, I agree with your comment, and this is something that could be done in general. As far as design goes, I would in fact prefer to not have 2 functions, but one with a fourth argument being `q=2`.

Datseris:
but please, if you attempt to code the "very optimized version" with the separation into boxes, do it as a differetn function so that we can benchmark one vs the other.

src/dimensions/correlationdim.jl
```diff
@@ -158,3 +158,53 @@ function takens_best_estimate(X, εmax, metric = Chebyshev())
...
+    N = length(X)
+    for i in 1:N
+        for j in 1:N
+            for iε in N:-1:1
+                if evaluate(norm, X[i], X[j]) < εs[iε]
+                    C_qs[iε] += 1
+                else
+                    break
+                end
+            end
```
Datseris, line 204: the norm here has to be evaluated outside the `iε` loop. It is the same and doesn't depend on `ε`.

apbraun:
> Yes, I agree with your comment, and this is something that could be done in general. As far as design goes, I would in fact prefer to not have 2 functions, but one with a fourth argument being `q=2`.
I have adapted the `q_order_correlationsum` to also be slightly optimised for `q=2`. The optimisation resembles the function `correlationsum` that's already implemented.

apbraun:
>but please, if you attempt to code the "very optimized version" with the separation into boxes, do it as a different function so that we can benchmark one vs the other.
I've been working on #118 and included the q-order correlation instead of the normal correlation. Should I commit this inside of this PR or put it into a separate one?

Datseris:
> I have adapted the q_order_correlationsum to also be slightly optimised for q=2. The optimisation resembles the function correlationsum that's already implemented.

I think it is better to have one function `correlationsum` with `q` as argument, which for `q=2` uses the previous version, the current `correlationsum`.

> I've been working on #118 and included the q-order correlation instead of the normal correlation. Should I commit this inside of this PR or put it into a separate one?

Let's merge this first and then you make another one. (BTW have you signed up for Hacktoberfest?)

apbraun:
I've realised that I'd not tested much besides `q=2` and for `q≤1` there needs to be an other implementation than the sum. Kantz also introduces the boundary case of `q=1` and refers to other authors for `q<1`, but I haven't had the time to implement it or take a look at it. Therefore I would just throw a warning for `q≤1`.

apbraun:
> I think it is better to have one function `correlationsum` with `q` as argument, which for `q=2` uses the previous version, the current `correlationsum`.

Sounds really good. I've tested the speed of the q-order correlation for `q=2` versus the current implementation of `correlationsum` and the q-order correlation is faster. This already allows for `w` to be adjusted as in the current `correlationsum`.
comp_q-corr_corr
Should I replace it? The current version is shorter codewise.

Datseris:
YES!

apbraun:
> The current version is shorter codewise.

That could have been misleading I was trying to say that my version is slightly longer since I write both loops explicity.

Datseris:
That is fine, the code is clear anyways. But I've noticed that you use an uncommon number of spaces for identation, six I think? Can you please re-ident the code to use 4 spaces per indentation level?

apbraun:
I have the following layout in mind:
`correlationsum(X, ε; q = 2, keyargs)` just distributes `X` into `correlationsum(X, X, ε; keyargs...)` if `q=2` and into `correlationsum(q, X, X, ε; keyargs...)` for `q≠2`, since a correlationsum that accepts `X` and `Y` is needed for the boxed stuff as well.

Datseris:
Wait why can there exist two datasets in the same function? X, Y?

apbraun:
If I box the data and want to calculate the number of points inside of ε-balls around each point of a box, I need to take into account the points of the neighbouring boxes as well. Then `X` consists of the points inside the box while `Y` contains the points of the box and the points of the neighbouring boxes.

src/dimensions/correlationdim.jl
```diff
@@ -116,6 +116,99 @@ function grassberger(data::AbstractDataset, εs = estimate_boxsizes(data); kwarg
...
+# q-order correlationsum
+################################################################################
+"""
+	q_order_correlationsum(q, X, ε, norm = Euclidean()) → C_q(ε)
+	q_order_correlationsum(q, X, Y, ε, norm = Euclidean()) → C_q(ε)
+Calculates the q-order correlation sum for data points `X`. `q` is the exponent of all points that are in radius `ε` of a point. Uses the method by Kantz and Schreiber[^Kantz] In accordance to the formula:
+```math
+C_q(\\varepsilon) = \\frac{1}{N(N-1)^{(q-1)}} \\sum_{i=1}^N\\left[\\sum_{i \\ne j} \\Theta(\\varepsilon - ||X_i - X_j||)\\right]^{q-1}
+```
+where ``\\Theta`` is the Heaviside function yielding one if the argument is greater than 1.
```
Datseris, line 129: Please modify this to use ``I`` which returns true or false instead. This is much less confusing that the Heaviside function.

src/dimensions/correlationdim.jl
```diff
@@ -116,6 +116,99 @@ function grassberger(data::AbstractDataset, εs = estimate_boxsizes(data); kwarg
...
+
+If `ε` is a vector of radii, it uses a slight optimisation to calculate the correlationsums. In this case `ε` should be ordered in increasing order.
+
+If `X` and `Y` are given, it computes the between all points in `X` and `Y`. In the folllowing formula:
```
Datseris, line 133: `it computes the between` something is missing here.

apbraun: There should be a `distances` in between `the between`.

Datseris:
I am really, really confused with what you are doing here. I have to admit that the documentation string explanation is not clear to me. Please explain what is going on in detail on Monday, because I still don't see why there should exist two datasets.

apbraun:
For the classic correlation algorithm, all of the implementations with two datasets do indeed make no sense. However when using the algorithm proposed by Theiler (1987), all points are distributed into boxes and the `correlationsum` is calculated separately for the points inside each box instead of all points at once. 
> James Theiler, [Efficient Algorithm for Estimating the Correlation Dimension from a Set of Discrete Points, 1987, Phys. Rev. A, 36(9)](https://www.researchgate.net/publication/13390536_Efficient_Algorithm_for_Estimating_the_Correlation_Dimension_from_a_Set_of_Discrete_Points)

For each of these correlation sum calculations there are now two datasets: 
- The first are  the points inside the box themselves, `X`.
- Since for a box size of `ε` there can be points outside of the original box that are still within radius `ε` of a point inside the original box there needs to be a second dataset containing more points than the original, `Y`.

The way in which the `correlationsum` for two datasets is written, is that for two equal datasets, the calculated sum equals the normal `correlationsum`. I thought I could save some lines of code if I based both functions on the same underlying function, since they operate very similar, but as it only introduced confusion so far, I will abandon that approach and separate these functions. Thereby also rendering the pull request for the implementation of the Theiler paper independent of this one.

Datseris:
Okay, thanks, yes then I agree. Clarity is more important in this case, especially since we will be sharing this code in the paper.

apbraun:
I've changed it and also included support for `w`. I thought that `w=1` was an unintuitive default value and chose `w=1`. Also the functions for `q≠2` do look ugly now because I iterate through the data left and right of each point in two separate loops but I could not think of a better way.

src/dimensions/correlationdim.jl
```diff
@@ -25,61 +25,112 @@ function kernelprob(X, ε, norm = Euclidean())
...
+                    else
+                        break
+                    end
+                end
+            end
         end
-        if Cs[k] ≈ 1/factor
-            Cs[k:end] .= 1/factor
-            break
+        return Cs .* 2 / ((N-w)*(N-w-1))
```
Datseris, line 104: my advice here (for increasing readability) would be to separate this into two functions depending on the value of `q`. `corrrelationsum` can call either `correlationsum_2` or `correlationsum_q`. There you should also make all input parameters positional arguments instead of keywords.

apbraun: Sounds reasonable, I will do so. The algorithm for `εs::AbstractVector` also works for `ε::Real` if I drop the typing but returns a Vector containing one number. Should I keep the algorithm with `ε::Real`?

apbraun:
> I've changed it and also included support for `w`. I thought that `w=1` was an unintuitive default value and chose `w=1`.

I meant `w=0`.

Datseris:
Why is w=1 unintuitive? More importantly, why is w=0 _more_ intuitive?

src/dimensions/correlationdim.jl
Datseris: I'd just define a version that for `ε::Real` simply calls `r = correlationsum([ε])` and then returns `r[1]`. It is intuitively expected that single number should output single number.

apbraun:
As far as my understanding goes there are discrete and continuous chaotic systems and for a discrete chaotic system the position of points do not form a trajectory. Further the thought was that `w` corrects the effects of points that are close in space on behalf of being close in time. This would not apply for discrete systems. For a continuous system `w=1` could be a good choice, but the even better choice would be to choose `w` according to the information specified in the docstring and therefore `w=0` seems more intuitive to me than `w=1`.

apbraun:
I've added the `correlationsum_q` and `correlationsum_2` dispatch and reincluded the former version of the `correlationsum_2` for `εs::AbstractVector` since it was faster. This also means that the versions for `ε::Real` stayed since they are faster. I'll write the tests tomorrow and then this PR would be ready to be merged from my point of view.

src/dimensions/correlationdim.jl
```diff
@@ -25,36 +25,69 @@ function kernelprob(X, ε, norm = Euclidean())
...
-    return 2C/((N-w)*(N-1-w))
 end
 
-"""
-    correlationsum(X, εs::AbstractVector; kwargs...) → Cs
-Calculate the correlation sum for every `ε ∈ εs` using an optimized version.
-"""
-function correlationsum(X, εs::AbstractVector; norm = Euclidean(), w = 1)
+function correlationsum_2(X, ε::Real, norm = Euclidean(), w = 0)
+    N, C = length(X), 0.
```
Datseris, line 59: instead of `0.` it is better for type stability to write `zero(eltype(X))`. Also, `0.` is not nice syntax and `0.0` should be preferred if you want float numbers.

apbraun:
Is this expected behaviour? I don't see any obvious flaws in the implementation of `correlation_q`, but I would have expected, that the dimension would be lower than the dimension of the system itself.
correlationdim_dim_vs_q

Datseris:
Can you tell me what you plot ? What is y axis and what is x axis?

apbraun:
I'm not sure how to access the conflicts, but they should result from the changes in the tests for Takens estimate. What would be the cleanest way to resolve them?

apbraun:
> Can you tell me what you plot ? What is y axis and what is x axis?

Sure. The x-axis is the number I chose `q` to be and the y-axis is the dimension calculated using `correlationsum` with that explicit `q`. This was done using the Henon and the Lorenz system. The linear fit was just for fun, but the values converge even better onto a line for bigger values of `q`. (I didn't include it since it didn't allow to see the behaviour for small `q`)

apbraun:
Another problem stems from the case that the q-order correlationsum gets infinite whenever there is a point that has no other point within distance `ε` for `q<1`. This is reasonable since the formula 
Screenshot from 2020-10-28 00-18-46
basically amounts to `1/(0^abs(q))` if one of the inner sums is zero. While for `q=1` all sums become 1. Kantz offers a solution to this, but it seems to be more dedicated and I'm currently unsure if it's worth it.

Datseris:
> Kantz offers a solution to this, but it seems to be more dedicated and I'm currently unsure if it's worth it.

What do you mean by more "dedicated"...? 

I have another comment btw. The formula that you present doesn't take the Theiler window correction into account. Is this on purpose?

apbraun:
> I have another comment btw. The formula that you present doesn't take the Theiler window correction into account. Is this on purpose?

No. the right one would be
Screenshot from 2020-10-28 14-13-14

Datseris:
Let's use the formula with the Theiler window. However the equation 11.17 you present is not normalized. It doesn't really matter however, we can have an unnormalized form.

apbraun:
>What do you mean by more "dedicated"...?

In the current version one defines a neighbourhood `ϵ` for each point and looks at the fraction `p` of the total number of points that lie within that radius. The idea is to swap this around and choose a fraction `p` and look which radius `ϵ` would fulfill this constraint for each point.

Datseris:
> In the current version one defines a neighbourhood ϵ for each point and looks at the fraction p of the total number of points that lie within that radius. The idea is to swap this around and choose a fraction p and look which radius ϵ would fulfill this constraint for each point.

This should be two independent functions. Both are reasonable to be implemented, but let's finish the implementation here. If there are no more entries for q < 1 then the output we get is the numerically consistent one. Not really useful, but consistent.

We really have to start merging some of these PRs, they have been around for too long I think, makes the whole process rather inefficient. I'd say to finish this open, and remove the "draft" status.

apbraun:
> Let's use the formula with the Theiler window. However the equation 11.17 you present is not normalized. It doesn't really matter however, we can have an unnormalized form.

This is the formula I've implemented. The normalisation is hidden inside of `α`.

I've spent one or two thoughts about whether it's possible to compute the first sum from `1` to `N`, but the normalisation becomes really ugly.

Datseris:
Can you resolve the conflicts first please? I.e. merge master branch here?

apbraun:
I would totally agree on finishing these PRs and I'm sorry it took so long. The only thing I would see missing so far is to resolve the conflicts.
So far the tests constructed check for:
- The correct normalisation in the `correlationsum` using only one `ϵ` for the Henon system with `q=2` and one `q!=2` since these are split into separate functions. Each `q` is calculated with `w=0` and `w=1` to ensure the correct normalisation works with and without Theiler window.
- The right dimension for `q=2` and `q=2.0001`. While both are calculated within different functions, both should produce results that are inside intervals, that have previously been defined. This is done for the Lorenz and the Henon system.

I'm not sure whether there should also be tests that ensure that the other function is normalised correctly or a test that the normalisation works in the Lorenz system as well.

apbraun:
> Can you resolve the conflicts first please? I.e. merge master branch here? (this becomes very easy in GitKraken)

Sure.

src/dimensions/correlationdim.jl
```diff
@@ -25,39 +25,73 @@ function kernelprob(X, ε, norm = Euclidean())
...
 ```
-where ``N`` is its length and ``I`` gives 1 if the argument is `true`.
-`w` is the Theiler window, a correction to the correlation sum that skips points
-that are temporally close with each other, with the aim of removing spurious correlations.
-
-See the book "Nonlinear Time Series Analysis", Ch. 6, for a discussion
-around `w` and choosing best values.
+for `q=2` and
+```math
+C_q(\\epsilon) = \\frac{1}{(N-w)(N-w-1)^{(q-1)}} \\sum_{i=1}^N\\left[\\sum_{|i-j| > w} I(||X_i - X_j|| < \\epsilon)\\right]^{q-1}
```
Datseris, line 38: In this formula the normalization is `- w`, but in the code you define `normalization` using `2w` instead of `w`. Is this difference crucial? If yes, the formula in the docstring should be updated!

apbraun: It's crucial as far as the normalisation is different without it so yes. I updated it.

apbraun:
Should I increase the version number in the `Project.toml` file?

Datseris:
Sure, and put an entry in the CHANGELOG

apbraun:
The last commit message is misleading. I increased the version number and put an entry into the changelog. This shouldn't affect the tests so I would suppose we're good to go.

Issue #117, opened 2020-06-17
O(N) algorithm for calculating box-counting dimension


Datseris:
In 

> Molteno, T. C. A. (1993). Fast O(N) box-counting algorithm for estimating dimensions. Physical Review E, 48(5), 3263–3266. https://doi.org/10.1103/PhysRevE.48.R3263

The author provides a fast algorithm for calculating the capacity dimension (actually for counting boxes needed to cover a set) which is faster than our generic O(Nlog(N)) version.

The paper is straightforward and easy to implement, taking advantage of binary representation of numbers.

apbraun:
So I would assume since #130 is merged this can now be called closed

PR #130, opened 2020-10-04
Implements the dimension method by Molteno 1993

apbraun:
This implements  #117. Tests have yet to be included.


Datseris:
Here is some advice for writing code that is easier to read (don't worry your code is fine): have the main function first and then the functions called by the main function. This parallels more closely scientific and algorithmic texts where a high-level summary of a concept is given and then each part is detailed more.

Datseris:
I'd also like to see some performance comparison please.

Datseris:
@apbraun what is the status here?

apbraun:
The tests are still missing but besides that I it should actually be finished, I will check it tomorrow.

apbraun:
I was wondering whether the distribution into prisms could also yield some kind of dimension, which would resolve the problem of bad scaling in high dimensions. By prism I mean an object that has edge length `r0` in `d` dimensions and if the data has `m` dimensions the other `m-d` dimensions span the whole dataset. This implies `d<=m`. It's only a rough thought though. I have no clue how normalisation would work in this context and whether the loss of information is negligible.

Datseris:
> which would resolve the problem of bad scaling for high dimensions

Can you give me some context here? Which problem is this? What is a prism? (Remember, I haven't read the paper)

apbraun:
> Can you give me some context here? Which problem is this? What is a prism? (Remember, I haven't read the paper)

Sure. Theiler (1987) describes a prism as an object having side length `r0` in `d` dimensions and spanning the entire dataset in the other `m-d` dimensions. The Molteno algorithm fares really bad for higher dimensions since it terminates if the mean number of points per filled box falls below `k0`, recommended to be 10. As the algorithm splits each box into `2^m` boxes, for high dimensional data the algorithm terminates after a low number of partitioning steps. That means that there are not enough points to properly fit a line through since every step creates a data point for the estimation of the dimension. If one instead of boxes chooses prisms of dimension `d` the number of possible new boxes decreases to `2^d`.

apbraun:
Would you put every test in the corresponding file, like tests for `correlationsum` into `tests/dimensions/correlationdim.jl` and tests for `generalized_entropy` into `tests/dimensions/dims.jl` or all into a single file as before?

Datseris:
Good point, yes let's keep things separated and put everything correlation related in a different file

apbraun:
I've constructed the tests and would just wait until #129 is merged to not produce any merge conflicts.

Datseris:
Yeah, it's in!

src/dimensions/dims.jl
```diff
@@ -211,6 +211,100 @@ boxcounting_dim = capacity_dim
...
 
+################################################################################
+# Molteno histogram based dimension by boxing values
+################################################################################
+"""
+	molteno_dim(α, data::AbstractDataset, k0 = 10)
+Calculates the generalized dimension using `float_to_int` for projection of `data` onto `UInt64`, `molteno_boxing` as algorithm for box division defined by Molteno[^Molteno], `genentropy` for the calculation of approximation of the limit and `linear_region` to identify the linear region and approximate its slope.
+
+[^Molteno]: Molteno, T. C. A., [Fast O(N) box-counting algorithm for estimating dimensions. Phys. Rev. E 48, R3263(R) (1993)](https://doi.org/10.1103/PhysRevE.48.R3263)
+"""
```
Datseris, line 222: Here I have the following comments:

1. Does the user need to know about `float_to_int`? I think we should only mention `molteno_boxing`.
2. Change `molteno_boxin` to have [ ](@ref) around it.
3. We need to export the two functions. Add the statement `export molteno_dim, molteno_boxing` so that that they are accessible upon `using ChaosTools`.

apbraun: I'm not sure whether we should also export `float_to_int` since it is required to access `molteno_boxing`. However I'll stop for now and continue tomorrow. Maybe there is a cool dispatch to use so that `float_to_int` is automatically called whenever the given dataset is not of eltype integer.

Datseris: Yes that is easy, I'll push a commit now. A comment on docstrings: use 2nd person imperative form "Calculate this..." instead of "Calculates this..." and don't repeat types that are already in the call signature. If you write:
```
"""
	float_to_int(data::AbstractDataset{D,T}) where {D, T}
Calculates maximum and minimum value of `data` of type `AbstractDataset` to...
"""
```
You can instead say 
```
Calculate maximum and minimum value of `data` ...
```
The type of `data` is already clear and does not have to be repeated.

Datseris:
@apbraun I've done some modifications. There were some funky stuff going on with using `Ref` and using `data[]` which I don't see why should happen. The main point is the documentations, which are not adequate at the moment. Functions in DynamicalSystems.jl have a `## Description` section which describes what the algorithm does in a concise but clear manner, so please add such restriction. 

I remember that this algorithm can't be used in High dimensional data because it does 2^D subdivisions. Is this true? If so, we have to say it.

src/dimensions/dims.jl
```diff
@@ -215,23 +215,29 @@ information_dim(args...) = generalized_dim(1, args...)
...
+[`generalized_dim`](@ref).
 
-[^Molteno]: Molteno, T. C. A., [Fast O(N) box-counting algorithm for estimating dimensions. Phys. Rev. E 48, R3263(R) (1993)](https://doi.org/10.1103/PhysRevE.48.R3263)
+## Description
+*full abetter description here*
+
+[^Molteno1993]: Molteno, T. C. A., [Fast O(N) box-counting algorithm for estimating dimensions. Phys. Rev. E 48, R3263(R) (1993)](https://doi.org/10.1103/PhysRevE.48.R3263)
 """
-function molteno_dim(α, data::AbstractDataset, k0 = 10)
+function molteno_dim(α, data::Dataset, k0 = 10; base = Base.MathConstants.e)
```
apbraun, line 229: I'm unsure whether it makes sense to pass `base` as argument. If I remember it correctly it was just introduced as artefact since `genentropy` had a bug concerning `base` and I wasn't aware of the possibility to merge the master into the working branch.

src/dimensions/dims.jl
```diff
@@ -262,6 +262,10 @@ Calculate the number of points per box in data, see [`molteno_dim`](@ref).
 *rework docstring. This function does not return "number of points"*
 """
 function molteno_boxing(data, k0 = 10)
+	integers, ϵ0 = float_to_int(data)
+	boxes = _molteno_boxing(integers, k0)
```
apbraun, line 266: I think it would make sense to also return either the size of the 0th box `ϵ0` or an array of box sizes `ϵs = ϵ0 ./ 2 .^ 1:length(boxes)`. `ϵs` would be more intuitive in my opinion.

src/dimensions/dims.jl
Datseris: `genentropy` has `base` as an argument, I think we should add `base` to both the `generalized_dim` as well as here. Also notice that we have to adjust `log(ε)` so that it uses the same base

Datseris:
🎉 I'll review soon!

PR #133, opened 2020-10-15
`poincaresos` for existing Dataset

Datseris:
Currently `poincaresos` takes in a `DynamicalSystem` and calculates the Poincare section with high precision using high order interpolation from DifferentialEquations.jl. However if one has experimental measurements and unknown `f` then they can't use this.

This simple addition allows for linear interpolation in existing datasets.

(wip: need to get the equation down for general hyperplane, currently works for `(j, r)` type of plane).


Datseris:
okay this needs tests and then it is good to go.

PR #134, opened 2020-10-18
Theiler boxing correlation

apbraun:
I implemented the correlation algorithm where the data is boxed beforehand. I have yet to find a clever way of including `w` as in the classic correlation algorithm. The algorithm itself runs rather smooth.
The boxsize estimators proposed by Theiler and Bueno-Orovio and Pérez-García rely on a random sample of the original dataset and the whole dimension estimation varies based on the choice of boxsize which itself varies quite strongly.


apbraun:
I've looked at this again and done some restructuring, but I'm not yet satisfied with the naming and structure. I hope to fix this and write the tests tomorrow.

apbraun:
I've included some tests for `boxed_correlationsum` and `boxed_correlationdim` which work quite similar to `correlationsum` and `grassberger`. If `boxed_correlationsum` is accessed with only the time series and an array of radii `ϵs` it behaves identical to `correlationsum` when accessed  with the identical parameters.

src/dimensions/correlationdim.jl
```diff
@@ -37,10 +38,19 @@ for `q=2` and
...
+for `q≠2`, where ``N`` is its length and ``I`` gives 1 if the argument is
+`true`. `w` is the Theiler window, a correction to the correlation sum that
+skips points that are temporally close with each other, with the aim of
+removing spurious correlations. If `ε` is a vector its values have to be
+ordered. See the book "Nonlinear Time Series Analysis"[^Kantz2003], Ch. 6, for
+a discussion around `w` and choosing best values and Ch. 11.3 for the
+definition of the q-order correlationsum.
+
+If `q = 2` and `ε` is a vector the size of the data has to be reasonably low to
+prevent computing distances twice. Otherwise please use
```
Datseris, line 50: What does this mean? Two terms are ambiguous: what does it mean "size" and what means "low"?

apbraun: I would rewrite it as 

If `q = 2` and `ε` is a vector, the `correlationsum` allocates a matrix of size `N^2`. If this is larger than your available memory please use ...

apbraun:
Should be ready to go from my point of view.

Datseris:
Seems good to me as well. Can you add an entry in the CHANGELOG?

Datseris:
Can you upload the comparison script to our repo? I want to use it for the documentation as well.

apbraun:
Which comparison script?

Datseris:
the one that compares all methods that you showed in our meeting. Add it to the FractalDimension repo.

apbraun:
Sure

apbraun:
Since the script in the meeting was quite rough at that point, I'll spend some more time on it tomorrow.

Datseris:
Do `estimate_r0_buenoorovio` and `estimate_r0_theiler` need to be exported? They can be referenced as `ChaosTools.estimate_r0_buenoorovio` in the docstrings, but would the user need to access them?

apbraun:
Yes. Since they provide the user with the best box size to use the algorithm with. `estimate_r0_theiler` doesn't run so smooth yet at hence isn't exported but `estimate_r0_buenoorovio` is. I was thinking about also providing an algorithm that already does the estimation of `r0` and `εs`.

Datseris:
I don't understand why we have `correlation_boxing(data, r0)` instead of `correlation_boxing(data, r0 = estimate_r0_buenoorovio(data))`.

apbraun:
That would be reasonable but wouldn't solve the problem entirely since `εs` also depends on `r0`. So far I've just chosen `εs = r0 .* .^ range(-2, stop = 0, length = 15)`.

Datseris:
Okay I think I need to read up the algorithms, because I don't know much. I'll have a look at our paper.

apbraun:
The algorithm basically does the same as the correlation algorithm but divides the data into boxes beforehand so the maximal radius for the correlation integral is the boxsize itself.

Datseris:
@apbraun can you email me your slides regarding the comparison? Is there a method you think we _shouldn't_ include into ChaosTools.jl?

apbraun:
I will send them to you tomorrow. If there is one method it should be the method by Molteno since the speedup seems to be maximal twice and it works badly for high dimensions. Also I haven't managed to get `estimate_r0_theiler` really stable so I would recommend using the method sketched out by Bueno-Orovio and Pérez-García.

Datseris:
I'll spend some time to re-organize the source once we have passing tests. All methods of the paper should be implemented here for clarity, but perhaps not all need to be highlighted equally in the documentation.

apbraun:
The prism correlationdimension is pretty easy I've actually already finished it. Should I put it into a separate PR or include it in this one? 
Since the boxing assisted correlation is just a special case of the prism assisted correlation, it would be easy to implement the prism algorithm and define the box algorithm as standard procedure.

Datseris:
Okay sounds fair then proceed within this PR. Code clarity is the n1 concern after all!

apbraun:
The name `boxed_correlationsum` could be somewhat misleading and I'm unsure whether the name `m` for the dimension of the prisms is a good choice, but the prism assisted correlation algorithm and its tests are implemented and seems to be quicker for some `m` than others (just the first impression nothing definite).

Datseris:
I use `D` for the dimension of the dataset, and here since `m` is also a dimension with `m <= D` we can instead use `M`.

Datseris:
BTW `DelayEmbeddings.dimension(data) == size(data, 2)` , makes for smaller code :)

src/dimensions/correlationdim.jl
```diff
@@ -201,19 +201,19 @@ end
...
     εs = 10 .^ range(log10(ε0), log10(r0), length = 16)
-    boxed_correlationdim(data, εs, r0; m = m, q = q)
+    boxed_correlationdim(data, εs, r0; M = M, q = q)
 end
 
 """
-    boxed_correlationdim(data, εs, r0 = maximum(εs); q = 2, m = DelayEmbeddings.dimension(data))
-    boxed_correlationdim(data; m = DelayEmbeddings.dimension(data), kwargs...)
+    boxed_correlationdim(data, εs, r0 = maximum(εs); q = 2, M = size(data, 2))
+    boxed_correlationdim(data; q = 2, m = size(data, 2))
```
Datseris, line 213:
```suggestion
    boxed_correlationdim(data; q = 2, M = size(data, 2))
```

Datseris:
@apbraun is this good to go from your end? Shall I give a final review and potentially merge?

Datseris:
Yes, it is.

apbraun:
I resolved the merge conflicts and think that we're good to go.

PR #145, opened 2020-12-07
Molteno entropies

apbraun:
The interface changes to the `genentropy` function were included into `molteno_dim`.

### DIFF
--- a/src/dimensions/molteno.jl
+++ b/src/dimensions/molteno.jl
@@ -23,9 +23,9 @@ estimate.
 
 [^Molteno1993]: Molteno, T. C. A., [Fast O(N) box-counting algorithm for estimating dimensions. Phys. Rev. E 48, R3263(R) (1993)](https://doi.org/10.1103/PhysRevE.48.R3263)
 """
-function molteno_dim(α, data, k0 = 10; base = Base.MathConstants.e)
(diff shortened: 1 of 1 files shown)


src/dimensions/molteno.jl
```diff
@@ -23,9 +23,9 @@ estimate.
 
 [^Molteno1993]: Molteno, T. C. A., [Fast O(N) box-counting algorithm for estimating dimensions. Phys. Rev. E 48, R3263(R) (1993)](https://doi.org/10.1103/PhysRevE.48.R3263)
 """
-function molteno_dim(α, data, k0 = 10; base = Base.MathConstants.e)
+function molteno_dim(α, data, k0 = 10; base = ℯ)
```
Datseris, line 26: I've changed the entropy signature of `generalized_entropy` to have `data` as first argument and `α` as keyword. Can you change this here as well? `α, k0` should be keyword arguments instead.

apbraun: sure.

apbraun: I fucked up the Commit message but `α` and `k0` are keyword arguments now.

apbraun: Should the `q` of the q-order correlation sum be also renamed to `α` to be identical to the naming in the paper?

PR #156, opened 2020-12-18
Better method for automatic boxsize esdtimation

Datseris:
~We now use the maximum and minimum pairwise distances. This should give better estimates for the upper bounds of the boxsizes.~

Just some polishing.

### DIFF
--- a/src/dimensions/linear_regions.jl
+++ b/src/dimensions/linear_regions.jl
@@ -125,7 +125,8 @@ end
 Return `k` exponentially spaced values: `base .^ range(lower + w, upper + z; length = k)`,
 that are a good estimate for sizes ε that are used in calculating a [Fractal Dimension](@ref).
 
-Let `d₋, d₊` be the minimum and maximum pair-wise distances in `A`.
+Let `d₋, d₊` be the minimum and maximum pair-wise distances in `A`, obtained
+from [`minmax_pairwise_distance`](@ref).
 Then `lower = log(base, d₋)` and `upper = log(base, d₊)`.
 Because by default `w=1, z=-1`, we're providing sizes that are an order of mangitude
 larger than the minimum distance, and an order of magnitude smaller than the minimum
--- a/src/dimensions/linear_regions.jl
+++ b/src/dimensions/linear_regions.jl
@@ -138,7 +138,7 @@ distance.
 * `base = 10.0` : the base used in the `log` function.
 """
 function estimate_boxsizes(
-        data::AbstractDataset;
+        A::AbstractDataset;
         k::Int = 12, z = -1.0, w = 1.0, base = 10.0, metric = Euclidean()
     )
 
--- a/src/dimensions/linear_regions.jl
+++ b/src/dimensions/linear_regions.jl
@@ -125,8 +125,8 @@ end
 Return `k` exponentially spaced values: `base .^ range(lower + w, upper + z; length = k)`,
 that are a good estimate for sizes ε that are used in calculating a [Fractal Dimension](@ref).
 
-Let `d₋` be the minimum and pair-wise distance in `A` and `d₊` the maximum extend of
-`A` along each of the dimensions.
+Let `d₋` be the minimum and pair-wise distance in `A` and `d₊` the length of the diagonal
+of the hypercube that contains `A`.
 Then `lower = log(base, d₋)` and `upper = log(base, d₊)`.
 Because by default `w=1, z=-1`, we're providing sizes that are an order of mangitude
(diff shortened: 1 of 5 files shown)


src/dimensions/linear_regions.jl
```diff
@@ -116,3 +116,72 @@ function linear_region(x::AbstractVector, y::AbstractVector;
...
+of all points in the dataset, and the corresponding point pairs.
+"""
+function minmax_pairwise_distance(A::AbstractDataset, metric = Euclidean())
+    tree = KDTree(A)
+    min_d = eltype(A[1])(Inf)
+    max_d = -min_d
+    min_pair = max_pair = (0, 0)
+    w = Theiler(0)
+    for p in 1:length(A)
+        inds, dists = Neighborhood.knn(tree, A[p], 1, theiler; sortds=false)
```
apbraun, line 175: Is `theiler` defined somewhere or should it be a variable of `minmax_pairwise_distance`?

src/dimensions/linear_regions.jl
```diff
@@ -116,3 +116,72 @@ function linear_region(x::AbstractVector, y::AbstractVector;
...
+of all points in the dataset, and the corresponding point pairs.
+"""
+function minmax_pairwise_distance(A::AbstractDataset, metric = Euclidean())
+    tree = KDTree(A)
+    min_d = eltype(A[1])(Inf)
+    max_d = -min_d
+    min_pair = max_pair = (0, 0)
+    theiler = Theiler(0)
+    for p in 1:length(A)
+        inds, dists = Neighborhood.knn(tree, A[p], 1, theiler; sortds=false)
```
apbraun, line 175: Shouldn't it be `theiler(p)` instead of `theiler`?

src/dimensions/linear_regions.jl
```diff
@@ -116,3 +116,72 @@ function linear_region(x::AbstractVector, y::AbstractVector;
...
+        ind, dist = inds[1], dists[1]
+        if dist < min_d
+            min_d = dist
+            min_pair = (p, ind)
+        end
+        if dist > max_d
+            max_d = dist
+            max_pair = (p, id)
```
apbraun, line 183: There's a typo here. `id` instead of `ind`.

Datseris:
Yeah many things were wrong. I wasn't even calculating maximum distance. I now remember why I use KD tree here. To get max distance would require make the `distancematrix` which is too costly. So, we revert back to the original idea. But I still did some minor polishing in the function.

apbraun:
Would it be sensible to only include such linear regions with a non-zero slopes? If I'm not totally wrong, a fractal dimension which is equal to zero, should mean that all data points are concentrated onto one single point or all used intervals are larger than the maximal distance or smaller than the minimal distance between two points. All these cases should either be easily detectable or unfavourable.
If the intervals are not chosen correctly there can be large intervals with a slope of zero before or after the actual linear region. If linear regions with a zero slope are ignored this would increase the chance of finding the actual linear region. As is poorly illustrated by this graph:
show

Datseris:
yeap that's what I introduced in the last commit

Datseris:
@apbraun the tests in my PR fail because of:

```
Testing correlation dimension with boxing beforehand...
Henon Map: Error During Test at C:\Users\m300808\.julia\dev\ChaosTools\test\dimensions\correlationdim.jl:44
  Got exception outside of a @test
  MethodError: no method matching log10(::Tuple{Int64,Int64})
  Closest candidates are:
    log10(!Matched::BigFloat) at mpfr.jl:656
    log10(!Matched::Missing) at math.jl:1197
    log10(!Matched::Complex{Float16}) at math.jl:1145
    ...
  Stacktrace:
   [1] estimate_r0_buenoorovio(::Dataset{2,Float64}, ::Int64) at C:\Users\m300808\.julia\dev\ChaosTools\src\dimensions\correlationdim.jl:480
   [2] estimate_r0_buenoorovio(::Dataset{2,Float64}) at C:\Users\m300808\.julia\dev\ChaosTools\src\dimensions\correlationdim.jl:466
   [3] top-level scope at C:\Users\m300808\.julia\dev\ChaosTools\test\dimensions\correlationdim.jl:47
```

apbraun:
I'll have a look

apbraun:
`estimate_r0_buenoorovio`, `boxed_correlationdim` and `estimate_r0_theiler` rely on `min_pairwise_distance`. The return values of `min_pairwise_distance` switched order and thus cause the error. I don't really know if I can or should push changes onto this branch.

Datseris:
Okay, I'll take care of it.

PR #157, opened 2020-12-18
Rework and rename `float_to_int` to fix InexactError

apbraun:
There was an `InexactError` in `float_to_int`. I previously found this error that occurs due to the truncation of a number larger than `typemax(UInt64)` to `UInt64`, but thought I'd fixed it. Can be easily recreated using
```julia
using BenchmarkTools, ChaosTools
@btime ChaosTools.float_to_int(trajectory(Systems.henon(0.1rand(2)), 1000, Ttr = 10))
```
 It seems like it was pure luck that it did not occur in the Test. This PR contains the best solution I could come up with. 
While fixing the error, I realized that the naming of this function was misleading and changed it to `real_to_uint64`. I think this name is more descriptive. Should the name `float_to_int` only be deprecated or can I discard it completely?

### DIFF
--- a/src/dimensions/molteno.jl
+++ b/src/dimensions/molteno.jl
@@ -53,27 +53,29 @@ calculated and returned together with the `probs`.
 See[^Molteno1993] for more.
 """
 function molteno_boxing(data::AbstractDataset; k0 = 10)
-    integers, ε0 = float_to_int(data)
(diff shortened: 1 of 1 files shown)


Datseris:
name is internal so its fine to change it

PR #160, opened 2021-02-02
Theiler window for boxed correlation

### LINKED ISSUE #159: q-order correlationsum is unintuitive
The q-order correlationsum currently follows the definition given by [Kantz and Schreiber](https://doi.org/10.1017/CBO9780511755798):
Screenshot from 2021-02-02 11-53-05
with
Screenshot from 2021-02-02 11-53-24
or in the  case of the boxed correlationsum:
Screenshot from 2021-02-02 11-52-45

 This leads to the requirement of multiplication by a factor of `1/(q-1)` after the linear regression has been made. Taking the definition given by [Grassberger on scholarpedia](http://www.scholarpedia.org/w/index.php?title=Grassberger-Procaccia_algorithm&action=cite&rev=91330):
Screenshot from 2021-02-02 11-56-30

and applying all other corrections, would solve this issue and make the q-order correlationsum more accessible.

apbraun:
I'll finish up the tests and this should be good to go.

Closes #159

### DIFF
--- a/src/dimensions/correlationdim.jl
+++ b/src/dimensions/correlationdim.jl
@@ -208,7 +208,7 @@ radii `εs` and boxsize `r0` are chosen by calculating
 [`estimate_r0_buenoorovio`](@ref).
 
 This method splits the data into boxes, calculates the q-order correlation sum
-C_q(ε) for every `ε ∈ εs` and fits a line through the longest linear looking
+`C_q(ε)` for every `ε ∈ εs` and fits a line through the longest linear looking
 region of the curve `(log(εs), log(C_q(εs)))`. The gradient of this line is the
 dimension.
 
--- a/src/dimensions/correlationdim.jl
+++ b/src/dimensions/correlationdim.jl
@@ -223,7 +223,7 @@ function boxed_correlationdim(data, εs, r0 = maximum(ε); q = 2, M = size(data,
 end
 
 """
-    boxed_correlationsum(A::Dataset, εs, r0 = maximum(ε); q = 2 , M = size(data, 2))
+    boxed_correlationsum(data, εs, r0 = maximum(εs); q = 2 , M = size(data, 2), w = 0)
 Distribute `data` into boxes of size `r0`. The `q`-order correlationsum
 `C_q(ε)` is then calculated for every `ε ∈ εs` and each of the boxes to then be
 summed up afterwards. If `M` is unequal to the dimension of the data, only the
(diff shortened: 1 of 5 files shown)


apbraun:
The tests just compare the `boxed_correlationsum` to the classic `correlationsum` which is already assured to be equal to one when the whole domain is considered.

apbraun:
The tests are failing on orbitdiagrams. It says that `Systems` is not defined.

Datseris:
Sorry about that, was my fault. Should be valid tests now. Can you update the mathematical formulas accordingly as in #159 ?

Datseris:
Tests do not pass locally for me.

apbraun:
Yeah that's my bad. Apparently there is no boxing and hence no need for a box size in the classic correlation sum :grin:. 
I included the exponent to satisfy #159. Also ran some benchmarks on the version with the Theiler window and compared it to the original one. I'm a bit confused that the version without Theiler window seems to be slightly slower (~3%) than the version with it, even when the window is not applied.

Datseris:
I'll give it a quick test with our project before merging and post figures here. Will have to wait a bit, i'm a bit busy with other projects at the moment.

Datseris:
image

Looks okay to me

src/dimensions/correlationdim.jl
```diff
@@ -215,30 +221,33 @@ See also: [`correlation_boxing`](@ref),
...
+function boxed_correlationsum(data, εs, r0 = maximum(εs); q = 2, M = size(data, 2), w = 0)
     @assert M ≤ size(data, 2) "Prism dimension has to be lower or equal than " *
     "data dimension."
     boxes, contents = correlation_boxing(data, r0, M)
     if q == 2
-        boxed_correlationsum_2(boxes, contents, data, εs)
+        boxed_correlationsum_2(boxes, contents, data, εs, w = w)
     else
-        boxed_correlationsum_q(boxes, contents, data, εs, q)
+        boxed_correlationsum_q(boxes, contents, data, εs, q, w = w)
```
Datseris, line 250: KEyword arguments should always be separated by positional arguments with `;` instead of `,`

src/dimensions/correlationdim.jl
```diff
@@ -317,44 +328,60 @@ function find_neighborboxes_2(index, boxes, contents)
...
+                # Calculate distance.
+		        dist = evaluate(norm, data[index_Y], x)
+		        for k in Nε:-1:1
+		            if dist < εs[k]
+		                Cs[k] += 1
+		            else
+		                break
+		            end
+		        end
+		    end
```
Datseris, line 362: there seems to be an identation missmatch here.

Datseris (approved):
Okay this is good to go, only minor comments. You also need to add a changelog entry and increment minor version value.

PR #161, opened 2021-03-16
Bugfix for `molteno_dim`

apbraun:
`k0` is a keyword argument for `molteno_boxing`. `molteno_dim` assumed it was an argument. The usage in `molteno_dim` was corrected.

### DIFF
--- a/src/dimensions/molteno.jl
+++ b/src/dimensions/molteno.jl
@@ -27,7 +27,7 @@ function molteno_dim(data; k0 = 10, α = nothing, q=1.0, base = ℯ)
         @warn "Keyword `α` is deprecated in favor of `q`."
         q = α
     end
-    boxes, εs = molteno_boxing(data, k0)
+    boxes, εs = molteno_boxing(data; k0 = k0)
     dd = genentropy.(boxes; q, base)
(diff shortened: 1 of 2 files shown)


src/dimensions/molteno.jl
```diff
@@ -27,7 +27,7 @@ function molteno_dim(data; k0 = 10, α = nothing, q=1.0, base = ℯ)
         @warn "Keyword `α` is deprecated in favor of `q`."
         q = α
     end
-    boxes, εs = molteno_boxing(data, k0)
+    boxes, εs = molteno_boxing(data; k0 = k0)
```
Datseris, line 30: You don't have to type `k0 = k0`. Only `k0` is enough in Julia 1.5+ which is anyway the supported version of DynamicalSystems.jl

Datseris: only the `, -> ;` is necessary to change

apbraun: Also up the minor version number?

Datseris: patch version

Datseris: see https://semver.org/

PR #162, opened 2021-03-17
Function that iterates the Poincaré map

awage:
Hello, 

this is a function that iterates the Poincaré map step by step. It would be useful to track the orbit on the map. 

Best

### DIFF
--- a/src/orbitdiagrams/poincare.jl
+++ b/src/orbitdiagrams/poincare.jl
@@ -111,10 +111,18 @@ function poincaresos(ds::CDS{IIP, S, D}, plane, tfinal = 1000.0;
     _check_plane(plane, D)
     integ = integrator(ds, u0; diffeq...)
 
-    i = typeof(idxs) <: Int ? i : SVector{length(idxs), Int}(idxs...)
+    i = typeof(idxs) <: Int ? idxs : SVector{length(idxs), Int}(idxs...)
     planecrossing = PlaneCrossing(plane, direction > 0)
 
-    data = poincaresos(integ, planecrossing, tfinal, Ttr, i, rootkw)
+	# Skip transient
+	Ttr != 0 && step!(integ, Ttr)
+
+    data = _initialize_output(integ.u, i)
+
+	while integ.t < tfinal
+	     push!(data, poincaremap!(integ, planecrossing, tfinal , i, rootkw))
+	end
+
     warning && length(data) == 0 && @warn PSOS_ERROR
 
     return Dataset(data)
@@ -128,41 +136,6 @@ end
 
 const PSOS_ERROR =
 "the Poincaré surface of section did not have any points!"
-
-function poincaresos(integ, planecrossing, tfinal, Ttr, j, rootkw)
-    f = (t) -> planecrossing(integ(t))
-    data = _initialize_output(integ.u, j)
-    Ttr != 0 && step!(integ, Ttr)
-
-    # Check if initial condition is already on the plane
-    side = planecrossing(integ.u)
-    if side == 0
-        push!(data, integ.u[j])
-        step!(integ)
-        side = planecrossing(integ.u)
-    end
-
-    while integ.t < tfinal + Ttr
-        while side < 0
-            integ.t > tfinal + Ttr && break
-            step!(integ)
-            side = planecrossing(integ.u)
-        end
-        while side ≥ 0
-            integ.t > tfinal + Ttr && break
-            step!(integ)
-            side = planecrossing(integ.u)
-        end
-        integ.t > tfinal + Ttr && break
-
-        # I am now guaranteed to have `t` in negative and `tprev` in positive
-        tcross = Roots.find_zero(f, (integ.tprev, integ.t), ROOTS_ALG; rootkw...)
-        ucross = integ(tcross)
-        push!(data, ucross[j])
-    end
-    return data
-end
-
 
 
 
@@ -217,7 +190,7 @@ function poincaremap(ds::CDS{IIP, S, D}, plane, Tmax = ∞;
 	_check_plane(plane, D)
     integ = integrator(ds, u0; diffeq...)
 
-    i = typeof(idxs) <: Int ? i : SVector{length(idxs), Int}(idxs...)
+    i = typeof(idxs) <: Int ? idxs : SVector{length(idxs), Int}(idxs...)
 
 	planecrossing = PlaneCrossing(plane, direction > 0)
 
@@ -238,9 +211,11 @@ end
 
 """
 	poincaremap!(integ, planecrossing,  idxs, rootkw, Tmax)
-Calculate the next iteration of the Poincaré surface of section.
+Low level function that returns the next iteration of the Poincaré surface of section.
 
 ## Arguments
+* `integ` : integrator object [`integrator`](@ref)
+* `planecrossing` : struct of type PlaneCrossing
 * `idxs = 1:dimension(ds)` : Optionally you can choose which variables to save.
   Defaults to the entire state.
 * `Tmax = ∞` : Stop the search after Tmax if the orbit does not cross the plane, maybe it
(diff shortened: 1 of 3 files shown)


Datseris:
Hi, thank you very much for the contribution, and welcome to the party! ;) 

I am a bit busy at the moment, but I promise to give a proper review as soon as I find the time.

One things that will be absolutely necessary before merging, which you can already do, is add a test suite. See how the `test` folder works and act accordingly. You will need to add a statement `export poincaremap` into the source code so that this function is accessible after `using ChaosTools`.

p.s.: I also saw you are working on Basin estimation which we would welcome in DynamicalSystems.jl as well.

awage:
Thanks for the message, I have already seen some bugs and clumsy errors. I will push a better code soon. 

We can talk later about my pet project of basin estimations.

src/orbitdiagrams/poincare.jl
```diff
@@ -164,6 +164,74 @@ function poincaresos(integ, planecrossing, tfinal, Ttr, j, rootkw)
...
+
+	iter_f! = (integ) -> poincaremap(integ, planecrossing, Tmax, i, rootkw)
+
+
+    return iter_f!, integ
+
+end
+
+"""
+	poincaremap(integ, planecrossing,  idxs, rootkw, Tmax)
```
Datseris, line 187: Can you please rename this to `poincaremap!` as it is mutating the `integ`?

src/orbitdiagrams/poincare.jl
```diff
@@ -164,6 +164,74 @@ function poincaresos(integ, planecrossing, tfinal, Ttr, j, rootkw)
...
+	_check_plane(plane, D)
+    integ = integrator(ds, u0; diffeq...)
+
+    i = typeof(idxs) <: Int ? i : SVector{length(idxs), Int}(idxs...)
+    planecrossing = PlaneCrossing(plane, direction > 0)
+
+	iter_f! = (integ) -> poincaremap(integ, planecrossing, Tmax, i, rootkw)
+
+
+    return iter_f!, integ
```
Datseris, line 182: I believe we can make this even more useful to the user by providing the following closure: a function that is called without any arguments and once called returns the next state on the Poincare map:

```
next_state = f()
```

you only need to define the return argment of `poincaremap` to do this process by wrapping `poincaremap!(integrator, ...)`

awage: I thought about that but I thought it would be useful to keep the control over the integrator in case a change of initial condition is needed. What about this: 

```jl 
# get the next state, f without argument
next_state = f()

# set the initial state, f with an initial vector: 
f(u0) 
```

I think it should be straightforward. What do you think?

Datseris: Perhaps even better would be that `f()` always returns the next state *from* the previous state, while `f(u0)` first sets the initial condition and then calls `f()`.

Datseris:
Thanks @awage , this looks very good to me. I only have some minor documentation comments, but I can do them quickly myself. One thing I am wondering is whether it makes sense for us to use this new code inside the existing `poincaresos` to reduce code duplication.

So the `poincaresos` will create the `poincaremap` and use this map to loop over the `integrator` as necessary.

What do you think? If you agree, would you like to do this change? In DynamicalSystems.jl we try hard (perhaps more than necessary) to have a clear and concise source code, and I think this can help here by eliminating code duplication.

awage:
I think this commit should do the trick. At least it does not seem to break anything. If you need more doc I can try to improve the docstrings.

src/orbitdiagrams/poincare.jl
```diff
@@ -129,38 +137,126 @@ end
...
+end
+
+function _tmp_f!(integ, planecrossing, Tmax, i, rootkw, u...)
+	if isempty(u)
+		return poincaremap!(integ, planecrossing, Tmax, i, rootkw)
+	else
+        reinit!(integ, u[1])
+		return poincaremap!(integ, planecrossing, Tmax, i, rootkw)
+	end
+end
```
Datseris, line 209: I am sure that this is bad for performance. Can you please benchmark current `poincaresos` versus the `poincaresos` of this PR? It doesn't look like this would lead to good performance, because functional dispatch is done within `if` statements instead of multiple dispatch.

I think we might need to make a simple wrapper struct `PoincareMap` and create dispaches: `(pmap::PoincareMap)()` and `(pmap::PoincareMap)(u0)`.

awage: I am not very proud of this code, and there is a loss in performance: 

poincaremap branch
@btime psos = poincaresos(ds, (2, 0.0), 10000.0, direction = +1)
  22.373 ms (57647 allocations: 3.30 MiB)

master branch
@btime psos = poincaresos(ds, (2, 0.0), 10000.0, direction = +1)
  20.502 ms (28833 allocations: 2.45 MiB)

I don't know how to do this wrapper. Is there a example somewhere you can point me to?

awage:
I figured out something although it is a bit sketchy. Good news is that there is an improvement in performance: 

`@btime psos = poincaresos(ds, (2, 0.0), 10000.0, direction = +1)
  18.872 ms (57647 allocations: 3.30 MiB)
`

That is slighlty better than the main branch. There are more allocations. Don't know why.

Datseris:
> That is slighlty better than the main branch. There are more allocations. Don't know why.

Is the "main" branch current `master`? If yes, can you please report the output of `master` as well?

src/orbitdiagrams/poincare.jl
```diff
@@ -129,38 +137,145 @@ end
...
+
+
+struct PoincareMap
+	integ
+	planecrossing
+	Tmax
+	i
+	rootkw
+end
```
Datseris, line 210: This cannot stay like this. Every single field of this type must be explicitly type-parameterized. This brings a huge performance benefit the way it is right now.

src/orbitdiagrams/poincare.jl
```diff
@@ -129,38 +137,145 @@ end
...
+
+
+function _tmp_f!(integ, planecrossing, Tmax, i, rootkw, u...)
+	if isempty(u)
+		return poincaremap!(integ, planecrossing, Tmax, i, rootkw)
+	else
+        reinit!(integ, u[1])
+		return poincaremap!(integ, planecrossing, Tmax, i, rootkw)
+	end
+end
```
Datseris, line 228: This piece of code is no longer necessary, correct?

awage:
I did some benchmarks on the two branches: 
`poincaremap branch

@btime psos = poincaresos(ds, (2, 0.0), 40000.0, direction = +1)
    82.975 ms (159906 allocations: 11.14 MiB)

master branch
  @btime psos = poincaresos(ds, (2, 0.0), 40000.0, direction = +1)
    82.708 ms (134313 allocations: 10.36 MiB)
`

There is a slight advantage on master branch. On 40000 iterations it is not much but a little worse.

src/orbitdiagrams/poincare.jl
```diff
@@ -129,38 +137,134 @@ end
...
+
+
+struct PoincareMap
+	integ::AbstractODEIntegrator
+	planecrossing::PlaneCrossing
+	Tmax::AbstractFloat
+	i::AbstractVector
+	rootkw::NamedTuple
+end
```
Datseris, line 209: This is still incorrect within the Julia language. The fields are still not concretely typed. What you have to do is:

```suggestion
struct PoincareMap{I, P, A, R}
	integ::I
 	planecrossing::P
	Tmax::Float65
	i::A
	rootkw::R
end
```

Datseris (changes requested):

src/orbitdiagrams/poincare.jl
Datseris: See also: https://www.youtube.com/watch?v=Fi7Pf2NveH0

Datseris: at 57 minutes

Datseris:
I wasn't entirely satisfied with the current design so I made some changes myself on what's happening. A main problem was what happened when the integrator passed the `Tmax`, where previously it wrongly returned the integrator's last state, which might have nothing to do with the actual section,.

I'm happy with current design and will merge if tests pass.

Datseris:
I also improved performance slightly by introducing a function barrier:
```
using ChaosTools, DynamicalSystemsBase
using BenchmarkTools

ds = Systems.roessler()
psos = poincaresos(ds, (2, 0.0), 1000.0)
@btime poincaresos(ds, (2, 0.0), 1000.0)

master: 
2.214 ms (2431 allocations: 153.22 KiB)

this pr:
2.140 ms (1747 allocations: 129.20 KiB)
```

Datseris (approved):

awage:
Wow! Thanks for the hard work! I will study these commits, I have still a lot to learn to make efficient code. 

A pleasure.

Datseris:
Contributions very welcome as always ;)

PR #165, opened 2021-03-31
Add DyCA method

### LINKED ISSUE #84: DyCA - Dynamical Component Analysis

Anantha-Rao12:
closes #84 
Hi! This PR contains an implementation of Dynamical Component Analysis based on the paper by Uhl et al. (2018) and [this](https://gist.github.com/Datseris/8121e5019071fe9e3cb9b1a0811ac26a) jupyter notebook originally written in python. I have added a file titled `dyca.jl` under `src/dyca.jl`. Please feel free to advise and comment. 

Checklist: 
- [x] I have read JuliaDynamics Contributing guidelines
- [x] Clear and concise Documentation  
- [x] Tests to check code breakage (if any) and performance

### DIFF
--- a/src/dyca.jl
+++ b/src/dyca.jl
@@ -38,7 +38,7 @@ end
 
 """
     dyca(data::Matrix{Float64},eig_thresold::Float64) -> eigenvalues, proj_mat, projected_data
-Compute the Dynamical Component analysis (DyCA) matrix [^Uhl2018] used for dimensionality reduction by solving the generalised eigenvalue equation.
+ Compute the Dynamical Component analysis (DyCA) matrix [^Uhl2018] used for dimensionality reduction by solving the generalised eigenvalue equation.
 
 
 ## Arguments
@@ -51,12 +51,13 @@ Compute the Dynamical Component analysis (DyCA) matrix [^Uhl2018] used for dimen
 - Data of the reduced dimension obtained by the projection matrix
 
 ## Description 
-Here, we solve the generalised eigenvalue equation: 
+Dynamical Component Analysis (DyCA),is a dimensionallity redcution technique leading to a classification of the underlying dynamics of high-domensional, multivariate signals and for a certain type of dynamics - to a signal subspace representing the dynamics of the data. Unlike methods like Principal Component analysis (PCA) that rely on some sort of stochastic model assumption, DyCA relies on a special deterministic model assumption and is a suitable tool for the reduction of chaotic time-series. Here, we solve the generalised eigenvalue equation: 
+
 ```math
 C_1 C_0^{-1} C_1^{\\top} \\bar{u} = \\lambda C_2 \\bar{u}
 
 ```
-where ``C_0`` is the correlation matrix of the signal with itself, ``C_1`` the correlation matrix of the signal with its derivative, and ``C_2`` the correlation matrix of the derivative of the data with itself. The eigenvectors ``\\bar{u}`` to eigenvalues approximately 1 and their ``C_1^{-1} C_2 u`` counterpart form the space where to project onto. 
+where ``C_0`` is the correlation matrix of the signal with itself, ``C_1`` the correlation matrix of the signal with its derivative, and ``C_2`` the correlation matrix of the derivative of the data with itself. The eigenvectors ``\\bar{u}`` with eigenvalues approximately 1 and their ``C_1^{-1} C_2 u`` counterpart, form the space where the data is projected onto. 
 
 ## Example: 
 
--- a/src/dyca.jl
+++ b/src/dyca.jl
@@ -35,21 +35,24 @@ Return the eigenvalues, projection matrix, and reduced-dimension data
 (which are just `data*proj_mat`).
 
 ## Description
-Dynamical Component Analysis (DyCA) is a dimensionallity redcution technique leading to a
-classification of the underlying dynamics of high-dimensional, multivariate signals and for
-a certain type of dynamics - to a signal subspace representing the dynamics of the data.
-Unlike methods like Principal Component analysis (PCA) that rely on some sort of stochastic
-model assumption, DyCA relies on a special deterministic model assumption and is a suitable
-tool for the reduction of chaotic time-series.
+Dynamical Component Analysis (DyCA) is a method to detect projection vectors to reduce
+the dimensionality of multi-variate, high-dimensional deterministic datasets. Unlike
+methods like PCA or ICA that make a stochasticity assumption, DyCA relys on a determinacy
+assumption on the time-series and is based on the solution of a generalized eigenvalue
+problem. The obtained eigenvectors are then used to project the high-dimensioanl dataset
+onto a lower dimension. The obtained eigenvalues measure the quality of the assumption of
+linear determinism for the investigated data. Furthermore, the number of the generalized
+eigenvalues with a value of approximately 1.0 are a measure of the number of
+linear equations contained in the dataset.
 
-Here, we solve the generalised eigenvalue equation:
+The generalised eigenvalue equation we solve is:
 
 ```math
 C_1 C_0^{-1} C_1^{\\top} \\bar{u} = \\lambda C_2 \\bar{u}
 
 ```
-where ``C_0`` is the correlation matrix of the signal with itself, ``C_1`` the correlation
-matrix of the signal with its derivative, and ``C_2`` the correlation matrix of the
+where ``C_0`` is the correlation matrix of the data with itself, ``C_1`` the correlation
+matrix of the data with its derivative, and ``C_2`` the correlation matrix of the
 derivative of the data with itself. The eigenvectors ``\\bar{u}`` with eigenvalues
 approximately 1 and their ``C_1^{-1} C_2 u`` counterpart, form the space where the data
 is projected onto.
@@ -71,7 +74,7 @@ function dyca(data, eig_thresold::AbstractFloat)
 
     #solve the generalized eigenproblem
     eigenvalues, eigenvectors = eigen(C1*inv(C0)*transpose(C1),C2)
-    eigenvectors = eigenvectors[:, vec(eig_thresold .< broadcast(abs,eigenvalues) .<= 1.0)]
+    eigenvectors = eigenvectors[:, vec(eig_thresold .< abs.(eigenvalues) .<= 1.0)]
     if size(eigenvectors,2) > 0
         mul!(C3, inv(C1), C2)
         proj_mat = hcat(eigenvectors,mapslices(x -> C3*x,eigenvectors,dims=[1]))
--- a/src/dyca.jl
+++ b/src/dyca.jl
@@ -5,8 +5,15 @@ using LinearAlgebra
 Compute the gradient of 2-dimensional array using second order accurate central differences in the interior points and either first order accurate one-sides (forward or backwards) differences at the boundaries. The returned gradient hence has the same shape as the input array. 
 
 ## Keyword Arguments
-- `matrix::Array{float64}` : The input matrix with size=(2,2)
+- `matrix::Array{float64}` : The input matrix with two dimensions
 - `axis::Number` : Axis to compute the gradient over (1 or 2) 
+
+## Example: 
+
+```julia
+random_array = rand(2,22;8,8);
+matrix_fdm_gradient(random_array,1)
+```
 
 
 """
@@ -24,6 +31,29 @@ function matrix_fdm_gradient(matrix,axis::Number)
     return gradient
 end
 
+"""
+    DyCA(data::Array,eig_thresold::float64)
+Computes the Dynamical Component analysis matrix used for Dimensionality reduction. Here, we solve the main eigenvalue equation: 
+$$C_1 C_0^{-1} C_1^\top u = \lambda C_2 u$$
+where $C_0$ is the correlation matrix of the signal with itself, $C_1$ the correlation matrix of the signal with its derivative, and $C_2$ the correlation matrix of the derivative of the data with itself. The eigenvectors $u$ to eigenvalues approximately $1$ and their $C_1^{-1} C_2 u$ counterpart form the space where to project onto. 
+
+## Keyword Arguments
+- `data::Array{float64}` : The input matrix with dimensions (2,2)
+- `eig_thresold::float64` : the eigenvalue thresold for DyCA
+
+
+## Example: 
+
+```julia
+random_array = rand(2,22;100,100);
+eigen_thresold = 0.8 ;
+DyCA(random_array,eigen_thresold)
+```
+
+Source : https://arxiv.org/abs/1807.10629 
+
+
+"""
 function DyCA(data,eig_thresold=0.98)
 
     derivative_data = matrix_fdm_gradient(data,1) ;
--- a/src/dyca.jl
+++ b/src/dyca.jl
@@ -22,16 +22,11 @@ matrix_fdm_gradient(random_array,1)
 
 
 """
-function matrix_fdm_gradient(matrix,axis::Int=1)
+function matrix_gradient(matrix)
     gradient = Array{Float64}(undef, size(matrix));
-    if axis == 1
-        gradient[:,1] = (matrix[:,2] .- matrix[:,1]) ;
-        gradient[:,end] = (matrix[:,end] .- matrix[:,end-1]);
-        gradient[:,2:end-1] = (matrix[:,3:end] - matrix[:,1:end-2]) .*0.5 ;
-    elseif axis == 2
-        gradient[1,:] = (matrix[2,:] .- matrix[1,:]) ;
-        gradient[end,:] = (matrix[end,:] .- matrix[end-1,:]);
-        gradient[2:end-1,:] = (matrix[3:end,:] .- matrix[1:end-2,:]) .*0.5 ;
+    gradient[:,1] = (matrix[:,2] .- matrix[:,1]) ;
+    gradient[:,end] = (matrix[:,end] .- matrix[:,end-1]);
+    gradient[:,2:end-1] = (matrix[:,3:end] - matrix[:,1:end-2]) .*0.5 ;
     end
     return gradient
 end
@@ -64,7 +59,7 @@ DyCA(random_array,eigen_thresold)
 """
 function DyCA(data,eig_thresold=0.98)
 
-    derivative_data = matrix_fdm_gradient(data,1) ;
+    derivative_data = matrix_gradient(data) ;
     time_length = size(data,1) ;#for time averaging
     
      #construct the correlation matrices
--- a/src/dyca.jl
+++ b/src/dyca.jl
@@ -68,8 +68,9 @@ function DyCA(data,eig_thresold=0.98)
     C2 = (transpose(derivative_data) * derivative_data )/ time_length ;
     
      #solve the generalized eigenproblem
-    eigenvalues, eigenvectors = eigen(((C1*inv(C0))*transpose(C1))*C2) ;
-    eigenvectors = eigenvectors[:,vec(eigenvalues .> eig_thresold) .& vec(eigenvalues .< 1.0)] ;
+    eigenvalues = eigvals(((C1*inv(C0))*transpose(C1))*C2) ;
+    eigenvectors = eigvecs(((C1*inv(C0))*transpose(C1))*C2);
+    eigenvectors = eigenvectors[:,vec(eig_thresold .< eigenvalues .<= 1.0)] ;
     if size(eigenvectors,2) > 0
         C3 = inv(C1) * C2 ;
         proj_mat = hcat(eigenvectors,mapslices(x -> C3*x,eigenvectors,dims=[1]))
--- a/src/dyca.jl
+++ b/src/dyca.jl
@@ -1,0 +1,39 @@
+using LinearAlgebra
+
+function matrix_fdm_gradient(matrix,axis::Number)
+    gradient = Array{Float64}(undef, size(matrix));
+    if axis == 1
+        gradient[:,1] = (matrix[:,2] .- matrix[:,1]) ;
+        gradient[:,end] = (matrix[:,end] .- matrix[:,end-1]);
+        gradient[:,2:end-1] = (matrix[:,3:end] - matrix[:,1:end-2]) .*0.5 ;
+    elseif axis == 2
+        gradient[1,:] = (matrix[2,:] .- matrix[1,:]) ;
+        gradient[end,:] = (matrix[end,:] .- matrix[end-1,:]);
+        gradient[2:end-1,:] = (matrix[3:end,:] .- matrix[1:end-2,:]) .*0.5 ;
+    end
+    return gradient
+end
+
+function DyCA(data,eig_thresold=0.98)
+
+    derivative_data = matrix_fdm_gradient(data,2) ;
+    time_length = size(data,1) ;#for time averaging
+    
+     #construct the correlation matrices
+    C0 = (transpose(data) * data )/ time_length ;
+    C1 = (transpose(derivative_data) * data )/ time_length ;
+    C2 = (transpose(derivative_data) * derivative_data )/ time_length ;
+    
+     #solve the generalized eigenproblem
+    eigenvalues, eigenvectors = eigen(((C1*inv(C0))*transpose(C1))*C2) ;
+    eigenvectors = eigenvectors[:,vec(eigenvalues .> eig_thresold) .& vec(eigenvalues .< 1.0)] ;
+    if size(eigenvectors,2) > 0
+        C3 = inv(C1) * C2 ;
+        proj_mat = hcat(eigenvectors,mapslices(x -> C3*x,eigenvectors,dims=[1]))
+    else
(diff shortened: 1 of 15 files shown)


src/dyca.jl
```diff
@@ -0,0 +1,79 @@
...
+## Example: 
+
+```julia
+random_array = rand(2,22;8,8);
+matrix_fdm_gradient(random_array,1)
+```
+
+
+"""
+function matrix_fdm_gradient(matrix,axis::Number)
```
Datseris, line 20: Is the version with axis = 2 actually used anywhere...? If not, no reason to code it here. The smaller the source code the better it is for us.

src/dyca.jl
```diff
@@ -0,0 +1,79 @@
...
+
+## Example: 
+
+```julia
+random_array = rand(2,22;100,100);
+eigen_thresold = 0.8 ;
+DyCA(random_array,eigen_thresold)
+```
+
+Source : https://arxiv.org/abs/1807.10629
```
Datseris, line 53: Notice that the paper has been peer reviewed and published here: https://ieeexplore.ieee.org/document/8517024 . While not open access, we should still use this citation. In other documentation strings you will see that we do stuff like `[^Author1999]` to cite. You should do this here to, because it nicely groups citations at the end of the document!

src/dyca.jl
```diff
@@ -0,0 +1,79 @@
...
+
+"""
+    DyCA(data::Array,eig_thresold::float64)
+Computes the Dynamical Component analysis matrix used for Dimensionality reduction. Here, we solve the main eigenvalue equation: 
+$$C_1 C_0^{-1} C_1^\top u = \lambda C_2 u$$
+where $C_0$ is the correlation matrix of the signal with itself, $C_1$ the correlation matrix of the signal with its derivative, and $C_2$ the correlation matrix of the derivative of the data with itself. The eigenvectors $u$ to eigenvalues approximately $1$ and their $C_1^{-1} C_2 u$ counterpart form the space where to project onto. 
+
+## Keyword Arguments
+- `data::Array{float64}` : The input matrix with dimensions (2,2)
+- `eig_thresold::float64` : the eigenvalue thresold for DyCA
```
Datseris, line 42: I do not understand this. It seems to me that this function does not accept keyword arguments at all.

src/dyca.jl
```diff
@@ -0,0 +1,79 @@
...
+        gradient[1,:] = (matrix[2,:] .- matrix[1,:]) ;
+        gradient[end,:] = (matrix[end,:] .- matrix[end-1,:]);
+        gradient[2:end-1,:] = (matrix[3:end,:] .- matrix[1:end-2,:]) .*0.5 ;
+    end
+    return gradient
+end
+
+"""
+    DyCA(data::Array,eig_thresold::float64)
+Computes the Dynamical Component analysis matrix used for Dimensionality reduction. Here, we solve the main eigenvalue equation:
```
Datseris, line 36: Dimensionality reduction => dimensionality reduction . It is a bit weird to capitalize the first word.

src/dyca.jl
```diff
@@ -0,0 +1,79 @@
...
+        gradient[2:end-1,:] = (matrix[3:end,:] .- matrix[1:end-2,:]) .*0.5 ;
+    end
+    return gradient
+end
+
+"""
+    DyCA(data::Array,eig_thresold::float64)
+Computes the Dynamical Component analysis matrix used for Dimensionality reduction. Here, we solve the main eigenvalue equation: 
+$$C_1 C_0^{-1} C_1^\top u = \lambda C_2 u$$
+where $C_0$ is the correlation matrix of the signal with itself, $C_1$ the correlation matrix of the signal with its derivative, and $C_2$ the correlation matrix of the derivative of the data with itself. The eigenvectors $u$ to eigenvalues approximately $1$ and their $C_1^{-1} C_2 u$ counterpart form the space where to project onto.
```
Datseris, line 38: Unfortunately you cannot use `$` for mathematics in Documenter.jl... You have to use double backticks and the math environment, see here: https://github.com/JuliaDynamics/ChaosTools.jl/blob/master/src/chaosdetection/lyapunovs.jl#L171-L184

src/dyca.jl
Anantha-Rao12: Although I haven't seen any DyCA implementation using axis=2, I thought that the need might arise when someone wants to perform DyCA along axis 2. The [NumPy documentation](https://numpy.org/doc/stable/reference/generated/numpy.gradient.html) uses an axis argument so I thought of the same.  I can think of two solutions to this : 
1. Provide only the axis=1 option and remove any conditional for the axis variable. The user will then need to simply transpose their matrix to perform DyCA along axis2. This will rightly be mentioned in the Documentation
2. No changes to be made here but will be mentioned in the Documentation. 

What do you think @Datseris ? Based on your input, I will do the needful.

src/dyca.jl
Anantha-Rao12: Sure. Will be done.

src/dyca.jl
Anantha-Rao12: I made a commit that clears this issue.

src/dyca.jl
Anantha-Rao12: Apologies. These are not keyword arguments.

src/dyca.jl
Anantha-Rao12: Thanks for letting me know! I've looked into this and performed a commit. Let me know if the format of the citation is fine and any further changes.

Anantha-Rao12 (commented):
Thanks for the message. I've looked into all the comments and added the corresponding changes. Please let me know if I missed out anything. 

> 
> Thank you for the contribution! I've left some documentation-related comments for now. Once this has tests as well, I will also check the algorithm in more detail (although this method is so trivially simple there isn't much to check, really)

Looking forward to future contributions. 
Ananth

src/dyca.jl
Datseris: >  I thought that the need might arise when someone wants to perform DyCA along axis 2

They can simply transpose the matrix before passing it to DyCA.

Side comment: please rename `DyCA` to `dyca`. Would be useful for you to also have a look at Julia's [Naming Conventions](https://docs.julialang.org/en/v1/manual/style-guide/)

Datseris:
And what about the tests? Can you simply port the test they do for their synthetic data and/or EEG data?

Anantha-Rao12:
Sure. I'll do that and share the notebook on gist.

Datseris:
> Sure. I'll do that and share the notebook on gist.

No, we need to have tests here, that are constantly checked. Please have a look at the `test` folder and model your tests accordingly.

Anantha-Rao12:
> Please have a look at the `test` folder and model your tests accordingly.

Right! I completely forgot about the unit test, although I mentioned that in the checklist for the PR. I was wondering what would be a good test for `dyca().` For the EEG data, they just compared the results of DyCA with PCA, and I couldn't think of any suitable unit test for that. Will be glad if you could advise on this. 

Thank you.

Datseris:
From reading the paper I see

image

The roessler system `Systems.roessler()` has 2 linear and 1 nonliner equation, thus if you apply this method to it, you should be able to see this.

Datseris:
Were you able to get the EEG data they used? Are they publicly available? If yes, do you get similar looking trajectories as they do?

Anantha-Rao12:
>The roessler system `Systems.roessler()` has 2 linear and 1 nonliner equation, thus if you apply this method to it, you should be able to see this.

Great! I'll have a look at that and commit the test.  

> Were you able to get the EEG data they used? Are they publicly available?

Yes. It is available in the .zip file here: https://github.com/JuliaDynamics/ChaosTools.jl/files/3065748/dyca.zip  along with a notebook that implements `dyca()` in python.

> do you get similar looking trajectories as they do?

Suprisingly no. The python implementation gives only two eigenvectors for their threshold (Original had three). You can find the notebooks for julia and python here: https://gist.github.com/Anantha-Rao12/35b479b4798fa85abac2ebe0f1dbf06d

Datseris:
> Suprisingly no

Whaaaat! Perhaps you should contact the authors about this, stating that you couldn't reproduce their work (be sure to double check your code before doing that). I always do this in DynamicalSystems.jl. You would be surprised _how often nonlinear dynamics papers are not reproducible_...

Datseris:
@Anantha-Rao12 what's the status here? We should wrap up this one soon.

Datseris:
Have you also confirmed that your Julia gradient function gives the same number's as Numpys we applied to some given data e.g. read from a textfile?

Anantha-Rao12:
Hi George. I've been trying to troubleshoot what's wrong with the Julia function here (not successful yet). The output for the EEG dataset seems to be different for me. I am getting two principal eigenvectors whereas the original implementation had 3 (which is correct and verified!). I have a hunch that this could be a problem with comparing complex values?

> Have you also confirmed that your Julia gradient function gives the same number's as numpys

Yes, both `np.gradient` and my `matrix_gradient` return the same matrices (I have tried different standard matrics) and both yield the same output.  However `matrix_gradient` works only along axis=1 (0 in python) since this is what is essentially needed for `dyca()`. 

> The roessler system Systems.roessler() has 2 linear and 1 nonliner equation, thus if you apply this method to it, you should be able to see this.

The IEEE paper did not mention anything about the eigenvalue threshold they use for the rossler system. Anyway I tried several values for the Rossler trajectory where some worked and some did not. I believe there is some issue with my code but I don't seem to find where. 

I have uploaded a notebook that summarises everything. Please have a look here: https://gist.github.com/Anantha-Rao12/7833573e44289b7dd26f1fed78270761

Thanks.

Datseris:
Have you contacted the authors yet to ask for the explicit threshoold they used for Roessler? This is something you need to familiarize yourself in doing when implementing papers of nonlinear dynamics.

Anantha-Rao12:
Dear @Datseris. The issue is solved. I'm getting the same trajectories for the seizures dataset as used by the authors.

Datseris:
> Then the data was embedded in a 25-dimensional space with additional multiplicative Gaussian noise. We used an exemplary signal to noise ratio of 15dB.

They mean that they performed delay coordinates embedding, and added to the signal gaussian noise with std 15% of the signal's std. Are you familiar with delay embedding? In principle you simple have to call `embed(x, 15, 1)` on the timeseries `x`. See here for more: https://juliadynamics.github.io/DynamicalSystems.jl/dev/embedding/reconstruction/ 

> The issue is solved. The method is complete and works

Okay, so what was the problem?

> Another method to test I thought was to obtain a trajectory of the Rossler system from Systems.Rossler and check the number of dyca components for different values of eigenvalue-thresolds.

Yes, you should do this as well. There 2 out of the 3 components should be linear.

Anantha-Rao12:
Hi @Datseris. I think the PR is complete now. Looking forward to your comments. 

> Okay, so what was the problem?

A simple bug. I was looking for an eigenvalue equation, not the generalized eigenvalue equation. 

> Are you familiar with delay embedding?

No actually, I was not familiar with delay embeddings. I read up the documentation (helped a lot!) and the Wikipedia page. Here are some points on the unit-test: 
1. Since we need to test the eigenvalues of `dyca()`, I added that to the return list of `dyca()` in `/src/dyca.jl`
```diff
- return  proj_mat, data * proj_mat
+ return eigenvalues, proj_mat, data * proj_mat
```

2. The test looks at eigenvalues (that are real) of the embedded system, counts the number of eigenvalues between (`eigenthresold`,1.0), and passes if the count is 2 (as discussed previously). I noticed that the count can fluctuate between [1,3] for some `eigenthresold` that is lower than 0.96 or are higher than 0.999 while it always passes (=2) for other thresholds. This maybe due to floating point errors :confused: 
This can be overcome for arbitrary thresholds [0.90,1.00], if we can do: 
```diff
- @test length(psum) == 2
+ @test length(psum) in (1,2,3)
```
Thanks. 
Ananth

test/dyca_tests.jl
```diff
@@ -0,0 +1,22 @@
...
+    cols = columns(ts);
+    for thresold in eigenthresold
+        for i in 1:3
+            Embedded_system = Matrix(embed(cols[i], 25, 1));
+            eigenvals,_ = dyca(Embedded_system,thresold)
+            # eigenvalues satisfy `dyca` condtiion and have imaginary part = 0.0
+            psum = a[vec(thresold .< broadcast(abs,a) .< 1.0) .& vec(imag(a) .== 0)]
+            # due to floating point errors more than 2 eigenvalues might satisy the above condition
+            # might be better to use `length(psum) in [1,2,3]`
+            @test length(psum) == 2
```
Datseris, line 18: Yes, you can use `in 1:3` instead here.

test/dyca_tests.jl
```diff
@@ -0,0 +1,22 @@
+using ChaosTools
+using Test
```
Datseris, line 2: Where are the tests for the non-embedded Roessler system? Where you take the `ts`, make it into a `Matrix`, and plug it into `dyca`? Does it give you 2 linear eigenvals?

src/dyca.jl
```diff
@@ -0,0 +1,86 @@
...
+    gradient = copy(matrix);
+    gradient[1,:] = (matrix[2,:] .- matrix[1,:]) ;
+    gradient[end,:] = (matrix[end,:] .- matrix[end-1,:]);
+    gradient[2:end-1,:] = (matrix[3:end,:] .- matrix[1:end-2,:]) .*0.5 ;
+    return gradient
+end
+
+"""
+    dyca(data::Array,eig_thresold::float64)
+Computes the Dynamical Component analysis matrix [^Uhl2018] used for dimensionality reduction. Here, we solve the generalised eigenvalue equation:
```
Datseris, line 35: Computes -> compute.  Immediately after the first sentence, state what are the input arguments, and what are the return values.

I think you should have a look at other documentation strings of `ChaosTools` and modify this one accordingly.

src/dyca.jl
```diff
@@ -0,0 +1,86 @@
...
+Computes the Dynamical Component analysis matrix [^Uhl2018] used for dimensionality reduction. Here, we solve the generalised eigenvalue equation: 
+```math
+C_1 C_0^{-1} C_1^{\\top} \\bar{u} = \\lambda C_2 \\bar{u}
+
+```
+where ``C_0`` is the correlation matrix of the signal with itself, ``C_1`` the correlation matrix of the signal with its derivative, and ``C_2`` the correlation matrix of the derivative of the data with itself. The eigenvectors ``\\bar{u}`` to eigenvalues approximately 1 and their ``C_1^{-1} C_2 u`` counterpart form the space where to project onto. 
+
+## Arguments
+- `data::Array{float64}` : The input matrix with size= (2,2)
+- `eig_thresold::float64` : the eigenvalue thresold for DyCA
```
Datseris, line 44: Does not need a dedicated section, should be part of the opening statements of the docstring.

src/dyca.jl
```diff
@@ -0,0 +1,86 @@
...
+
+## Returns
+- Eigenvalues of the Generalised eigenvalue problem
+- Projection matrix
+- Data embeded by the projection matrix
```
Datseris, line 49: same as above.

src/dyca.jl
```diff
@@ -0,0 +1,86 @@
...
+end
+
+"""
+    dyca(data::Array,eig_thresold::float64)
+Computes the Dynamical Component analysis matrix [^Uhl2018] used for dimensionality reduction. Here, we solve the generalised eigenvalue equation: 
+```math
+C_1 C_0^{-1} C_1^{\\top} \\bar{u} = \\lambda C_2 \\bar{u}
+
+```
+where ``C_0`` is the correlation matrix of the signal with itself, ``C_1`` the correlation matrix of the signal with its derivative, and ``C_2`` the correlation matrix of the derivative of the data with itself. The eigenvectors ``\\bar{u}`` to eigenvalues approximately 1 and their ``C_1^{-1} C_2 u`` counterpart form the space where to project onto.
```
Datseris, line 40: This stuff should be in a dedicated `## Description` section that gives appropriate info for the algorithm and its use.

src/dyca.jl
```diff
@@ -0,0 +1,86 @@
...
+function matrix_gradient(matrix::Matrix{Float64})
+    gradient = copy(matrix);
+    gradient[1,:] = (matrix[2,:] .- matrix[1,:]) ;
+    gradient[end,:] = (matrix[end,:] .- matrix[end-1,:]);
+    gradient[2:end-1,:] = (matrix[3:end,:] .- matrix[1:end-2,:]) .*0.5 ;
+    return gradient
+end
+
+"""
+    dyca(data::Array,eig_thresold::float64)
```
Datseris, line 34:
```suggestion
    dyca(data::Array,eig_thresold::float64) -> eigenvalues, proj_mat, projected_data
```

Datseris (changes requested):
Seems great to me, I only have comments for improving the docstring.

test/dyca_tests.jl
```diff
@@ -0,0 +1,23 @@
...
+    eigenthresold = [0.98,0.985,0.99,0.991];
+    ds = Systems.roessler(rand(3); a = 0.15, b = 0.2, c = 10);
+    ts = trajectory(ds, 1000.0, dt = 0.05);
+    cols = columns(ts);
+    for thresold in eigenthresold
+        for i in 1:3
+            Embedded_system = Matrix(embed(cols[i], 25, 1));
+            eigenvals,_ = dyca(Embedded_system,thresold)
+            # eigenvalues satisfy `dyca` condtiion and have imaginary part = 0.0
+            psum = a[vec(thresold .< broadcast(abs,a) .< 1.0) .& vec(imag(a) .== 0)]
```
Anantha-Rao12, line 15: ### Todo 
There is a bug here. It should be:
```diff
- psum = a[vec(thresold .< broadcast(abs,a) .< 1.0) .& vec(imag(a) .== 0)]
+ psum = eigenvals[vec(thresold .< broadcast(abs,eigenvals) .<= 1.0) .& vec(imag(eigenvals) .== 0)]

File: test/dyca_tests.jl (hunk 7 above)
Comment (Anantha-Rao12) [reply test/dyca_tests.jl:2]: Yeah. I'm adding that.

File: test/dyca_tests.jl (hunk 13 above)
Comment (Datseris) [reply test/dyca_tests.jl:15]: You should start having more lines of code and simpler expressions. All these things are very hard to read.

Comment (Datseris) [reply test/dyca_tests.jl:15]: Why write `broadcast(abs,eigenvals) ` instead of `abs.(eigenvals)`...?

File: src/dyca.jl (hunk 8 above)
Comment (Anantha-Rao12) [reply src/dyca.jl:35]: Cool. Working on that now.

File: test/dyca_tests.jl (hunk 6 above)
Comment (Anantha-Rao12) [reply test/dyca_tests.jl]: I used `in 0:4` and increased the range of threshold : `collect(0.90:0.01:0.995)`. The limiting cases of 0,4 occur at the extreme threshold values.

File: test/dyca_tests.jl (hunk 13 above)
Comment (Anantha-Rao12) [reply test/dyca_tests.jl]: I dont know why I chose the former. Anyway its `abs.(eigenvals)` now.

**Comment by Datseris** [comment]
Ops, one last thing we forgot before merging: please increment minor version in the Project.toml file, and add an entry in the CHANGELOG.md!

**Comment by Anantha-Rao12** [comment]
Umm. Are you sure that I should do it? (haven't done it before)
Anyway, I am thinking of the following:
# Project.toml : 
```diff
- version = "1.25.2"
+ version = "1.25.3"
```

# CHANGELOG.md
```diff
+ # 1.26
+ new function `dyca()` to perform Dynamical Component Analysis
```

Datseris (approved):

Datseris:
Have a look at https://semver.org , this is how we change versions in DynamicalSystems.jl 

This is good to go, I'll merge once tests pass.

Anantha-Rao12:
Trying to understand what this `LoadError` means.  Any idea why `UndefVarError: dyca not defined` occurred? 
Also, (just realized), is it better to group all test sets in dyca_tests.jl under another testset, or is the current version fine?

Datseris:
without an `export` statement the names are not available when `using`.

Datseris:
@Anantha-Rao12 I'd like you to add more info into the `Decription` part of the docstring. While you state what numerical procedure is done, you don't actually say anything about the method, or why it is useful, or what it actually does.

Can you write a couple of sentences in your own words?

Datseris:
Also, there is no reason to end every line with `;` in Julia.

src/dyca.jl
```diff
@@ -0,0 +1,102 @@
...
+
+Return the eigenvalues, projection matrix, and reduced-dimension data
+(which are just `data*proj_mat`).
+
+<<<<<<< HEAD
+## Returns
+- Eigenvalues of the Generalised eigenvalue problem
+- The Projection matrix
+- Data of the reduced dimension obtained by the projection matrix
+
```
Datseris, line 48: What are all these `<<<<<<<<` head stuff?

Anantha-Rao12: I actually have no idea. I thought it came when I pulled your docstring edits. I thought that it's a part of documenter.jl and left it as it is. If it's unrelated, then you can remove it.

Datseris: You have to learn how to properly merge using git so that when there is a conflict you can take care of it. This happened bbecause we both pushed commits at the same branch. Please have a look at one of GitHub's videos or have a look at YouTube for "git merge conflicts".

src/dyca.jl
```diff
@@ -0,0 +1,82 @@
...
+Return the eigenvalues, projection matrix, and reduced-dimension data
+(which are just `data*proj_mat`).
+
+## Description
+Dynamical Component Analysis (DyCA) is a dimensionallity redcution technique leading to a
+classification of the underlying dynamics of high-dimensional, multivariate signals and for
+a certain type of dynamics - to a signal subspace representing the dynamics of the data.
+Unlike methods like Principal Component analysis (PCA) that rely on some sort of stochastic
+model assumption, DyCA relies on a special deterministic model assumption and is a suitable
+tool for the reduction of chaotic time-series.
```
Datseris, line 43: @Anantha-Rao12 two comments: First, please do not use automatic wrapping on your text editor. Source code lines should never exceed 92 characters in width.

Second: the docstring now has more info, but I think we can improve it further yet. It sounds vague at the moment and does not convey the information it has to. Some examples:

* "leading to a classification of the underlying dynamics of high-dimensional" what is the classification? This is never stated. If you classify things, then you have to state the categories, otherwise what is the classification done for.
* "For a certain type of dynamics", which type? 
* "DyCA relies on a special deterministic model assumption" what assumption?

After reading the docstring I know that DyCa is a dimensionality reduction techinque but I don't know why I should use it. For example why is there no discussion of the linear / nonlinear subspaces?

src/dyca.jl
Anantha-Rao12: Sure. Will do that. Thanks for letting me know!

src/dyca.jl
Anantha-Rao12: Hmm. I understand. Working on it now! Will probably commit the changes after a few hours. Thanks for your detailed comments!

src/dyca.jl
```diff
@@ -0,0 +1,88 @@
...
+Compute the Dynamical Component analysis (DyCA) of the given `data` [^Uhl2018]
+used for dimensionality reduction.
+
+Return the eigenvalues, projection matrix, and reduced-dimension data
+(which are just `data*proj_mat`).
+
+## Description
+Dynamical Component Analysis (DyCA) is a method to detect projection vectors to reduce
+the dimensionality of multi-variate, high-dimensional deterministic datasets. Unlike
+methods like PCA or ICA that make a stochasticity assumption, DyCA relys on a determinacy
```
Datseris, line 40:
```suggestion
methods like PCA or ICA that make a stochasticity assumption, DyCA relies on a determinacy
```

Anantha-Rao12:
Thank you @Datseris. It was a great experience to work on this issue. Learnt so many things! Looking forward to many more contributions! 

PS Any idea why the GALI tests keep failing?

Datseris:
me too!

nope. they do not fail locally for me. blasted CI.

Issue #21, opened 2018-01-27
Use PkgBenchmark *everywhere*


Datseris:
https://github.com/JuliaCI/PkgBenchmark.jl

Well, there isn't anything more to say. This will improve consistency and allow to do performance comparisons between PRs easily.

Datseris:
Info for `Hacktoberfest`: Any PR that adds benchmakrs for parts of this library is accepted. (See `benchmark` folder).

Datseris:
This doesn't make sense for this repo, because each algorithm is completely independent.

PR #166, opened 2021-04-13
fixed mass algorithm for the calculation of the fractal dimension

apbraun:
I've written a simple fixed mass algorithm, that returns reasonable results for the Henon and Lorenz System. I've yet to crunch the numbers, but by feeling it is also reasonably fast. It was proposed by Grassberger in 1988:
Screenshot from 2021-04-13 14-44-35
As far as my understanding of the matter goes (which is not very far to be honest), the difference between upper and lower case for the ψ seems to be a typo and the equality in the first equation should be a proportionality (~). 
The implemented function returns the mean over the radii and the right hand side of the equation to compute a linear region. I will work a bit further on tests and documentation but currently 
```julia
henon = Systems.henon(0.1rand(2))
X = trajectory(henon, 10000, Ttr = 100)
rs, es = fixed_mass_algorithm(X, 30)
linear_region(rs, es)
```
produces reasonable estimates for the fractal dimension of the Henon set.


apbraun:
fixedmassdemo
These are my results so far.

apbraun:
I'll insert some changes on the boxed correlation sum in here soon, but I was wondering whether `M` should also be renamed to `P`, since that's the name used in the paper.

src/dimensions/correlationdim.jl
```diff
@@ -189,16 +189,16 @@ end
 ################################################################################
 # Correlationsum, but we distributed data to boxes beforehand
 ################################################################################
-function boxed_correlationdim(data; M = size(data, 2), q = 2)
+function boxed_correlationdim(data; M = size(data, 2), q = 2, w = 0)
     r0 = estimate_r0_buenoorovio(data, M)
     ε0 = minimum_pairwise_distance(data)[1]
     εs = 10 .^ range(log10(ε0), log10(r0), length = 16)
-    boxed_correlationdim(data, εs, r0; M = M, q = q)
+    boxed_correlationdim(data, εs, r0; M = M, q = q, w = w)
```
Datseris, line 196:
```suggestion
    boxed_correlationdim(data, εs, r0; M, q, w)
```

Julia 1.5+ has this nice feature that if the variable name and keyword name coincide, you don't have to type both.

src/dimensions/fixedmass.jl
```diff
@@ -0,0 +1,50 @@
...
+A fixed mass algorithm for the calculation of the fractal dimension according
+to [^Grassberger 1988] with `max_j` the maximum number of neighbours that
+should be considered for the calculation, `M` defines the number of points
+considered for the calculation, default is the whole data set.
+
+Implements
+```math
+D_q \\overline{\\log r^{(j)}} \\sim Ψ(j) - \\log N
+```
+where `` Ψ(j) = \\frac{\\text{d} \\log Γ(j)}{\\text{d} j}
```
Datseris, line 14: I think you need to write `\Psi` instead of `Ψ` in there, because it is inside latex formula.

src/dimensions/fixedmass.jl
```diff
@@ -0,0 +1,50 @@
...
+    N = length(data)
+    # Transform the data into a tree.
+    tree = KDTree(data, metric)
+    # Calculate all nearest neighbours up to max_j.
+    _, distances =
+        knn(
+            tree,
+            # If M is given, only M random points of the set are considered.
+            M == N ?
+                [point for point in data] :
```
Datseris, line 32: this command is literally equivalent with just `data`.

src/dimensions/fixedmass.jl
```diff
@@ -0,0 +1,50 @@
...
+                [data[index] for index in randperm(N)[1:M]],
```
Datseris, line 33:
```suggestion
                data[view(randperm(N), 1:M)],
```

this is a critical performance improvement you should be aware about (view)

src/dimensions/fixedmass.jl
```diff
@@ -0,0 +1,50 @@
...
+considered for the calculation, default is the whole data set.
+
+Implements
+```math
+D_q \\overline{\\log r^{(j)}} \\sim Ψ(j) - \\log N
+```
+where `` Ψ(j) = \\frac{\\text{d} \\log Γ(j)}{\\text{d} j}
+``, `rs` = ``\\overline{\\log r^{(j)}}`` and `ϵs` = ``Ψ(j) - \\log N``.
+
+``D_q`` can be computed by using `linear_region(rs, ϵs)`.
```
Datseris, line 17: Can you please change the `εs` to something else? use `ys` for example. Throughout the library we use it to mean box sizes or radii, but here it has more like the role of entropy or log(C).

src/dimensions/fixedmass.jl
apbraun: I also thought so, but I thought to have tested it and it did not work. I'll have a look at it once more

Datseris: you might need `data.data`, which is the actual vector of static vectors that the `Dataset` wraps.

apbraun: `data[:]` works as well, but `data.data` is more elegant.

apbraun:
What do you think about the name? Maybe `fixed_mass_correlation` or `fixed_mass_fractal_dimension`?

src/dimensions/fixedmass.jl
Datseris: `data[:]` makes an unecessary copy of the array.

Datseris:
I'd say `correlationsum_fixedmass`, as in essence this is the application of the correlation sum algorithm to fixed sum?

apbraun:
Seems reasonable. I've got an idea how to incorporate the Theiler window. Should I do it?

Datseris:
This can be done automatically with Neighborhoods.jl and `Theiler(w)`. https://julianeighbors.github.io/Neighborhood.jl/dev/#Theiler-window 

Aren't you using this library to find the neighbors?

apbraun:
No, I used the NearestNeighbour.jl library, but I'll change it.

Datseris:
How? NearestNeighbors.jl is not a dependency of CHaosTOols anymore, I have moved all code to use Neighborhood.jl.

apbraun:
I only used it in my own REPL without activating the environment and imported it manually. But now I fixed it. I still require a dependency on `SpecialFunctions.jl` for the digamma function.

Datseris:
you haven;t pushed anything here yet

src/dimensions/fixedmass.jl
```diff
@@ -11,40 +13,34 @@ Implements
 ```math
 D_q \\overline{\\log r^{(j)}} \\sim Ψ(j) - \\log N
 ```
-where `` Ψ(j) = \\frac{\\text{d} \\log Γ(j)}{\\text{d} j}
-``, `rs` = ``\\overline{\\log r^{(j)}}`` and `ϵs` = ``Ψ(j) - \\log N``.
+where `` \\Psi(j) = \\frac{\\text{d} \\log Γ(j)}{\\text{d} j}
+``, `rs` = ``\\overline{\\log r^{(j)}}`` and `ys` = ``\\Psi(j) - \\log N``.
 
-``D_q`` can be computed by using `linear_region(rs, ϵs)`.
+``D_q`` can be computed by using `linear_region(rs, ys)`.
```
Datseris, line 19: Here we are using `D_q` but there is no `q`. Does this correspond to the same dimension as the correlation sum for `q=2`? Then we can use some symbol ``\\Delta` and call it the correlation dimension.

Datseris:
Can you please also add some basic tests for this?

src/dimensions/fixedmass.jl
apbraun: It corresponds to the Correlation dimension for `q=1`, hence `D_q` is clearly false here.

apbraun:
Yes, that's the reason it took so long. There is some issue when GALI, that has the tests failing in some circumstances but not always.

apbraun:
Also I'm unsure whether the update to `Project.toml` is done correctly. I used `activate ChaosTools.jl` in the folder and then `]add SpecialFunctions`.

apbraun:
From my side this would be ready for merging. Unfortunately I'm not aware of any normalization in the used equations that allow for an easy check in tests but I test for the Lorenz and the Henon system varying all keywords.

apbraun:
Another question. I just saw issue #21 and was wondering whether it would be useful to write a benchmark for all fractal dimension estimation algorithms present in this package.

Datseris:
Well sure having benchmarks would be great, but unfortunately #21 is quite a bit of work. If you're interested you can have a look at how we do it for e.g. DynamicalBilliards.jl: https://github.com/JuliaDynamics/DynamicalBilliards.jl/blob/master/benchmark/benchmarks.jl  but of course for that package it required much more lines of code... If you want to do this, feel free to do it as your own open source contribution independently of the fractal dim project. Because it will take time and is not really related to the project.

GALI tests are a nightmare but forget them for now. they always pass locally for me and have been unchanged for years now.

---

tests are fine

---

okay but then if the algorithm is valid only for a given q (here q=1) then no reason to have subscript `q` into the docstring right? Also, we shouldn't use the letter `D` as that is used for the dataset dimensionality. I'll do these changes and merge and tag new version then.

apbraun:
I've gone ahead and changed it myself.

src/dimensions/fixedmass.jl
apbraun: Oh, good to know thought it was just `view(data, :)`.

PR #167, opened 2021-04-18
Added new basins of attraction branch

awage:
Hello! 

I worked on the code so that the systems like the magnetic pendulum could also work. Although these basins seem less reliable. 

Some issues not solve yet: 
* I have some basic tests but I have to think about a method to make them more reliable. 
* There are still some problems to solve with the concrete typing stuff. 
  
The architecture is the following: 
* There are high level functions adapted to the kind of problems (ODE, maps, ...) 
* There is a low level function called `draw_basin` that iterates the initial conditions that have not been classified already. 
* Then there is a function called `procedure!` that tracks the attractors and construct the basin. It is a bit hard to follow but the general idea is to be sure that some boxes have been visited many times by the orbit. This is where the work is done. 

Best

### DIFF
--- a/src/basins/compute_ba.jl
+++ b/src/basins/compute_ba.jl
@@ -1,0 +1,368 @@
+export draw_basin, basin_poincare_map, basin_stroboscopic_map, basin_discrete_map
+export basin_general_ds
+
+
+mutable struct basin_info{I,F,V}
+    basin :: I
+    xg :: F
+    yg :: F
+    iter_f! :: Function
+    reinit_f! :: Function
+    get_u :: Function
+    current_color :: Int64
+    next_avail_color :: Int64
+    consecutive_match :: Int64
+    consecutive_other_basins :: Int64
+    prevConsecutives :: Int64
+    prev_attr :: Int64
+    prev_bas :: Int64
+    prev_step :: Int64
+    step :: Int64
+    attractors :: V
+end
+
+
+"""
+    basin_poincare_map(xg, yg, integ; kwargs...)
+Compute an estimate of the basin of attraction on a two-dimensional plane using a Poincaré map.
+
+[H. E. Nusse and J. A. Yorke, Dynamics: numerical explorations, Springer, New York, 1997]
+
+## Arguments
+* `xg`, `yg` : 1-dim range vector that defines the grid of the initial conditions to test.
+* `pmap` : A Poincaré map as defined in [ChaosTools.jl](https://github.com/JuliaDynamics/ChaosTools.jl)
+
+## Keyword Arguments
+* `Ncheck` : A parameter that sets the number of consecutives hits of an attractor before deciding the basin
+of the initial condition.
+
+## Example:
+
+```jl
+using DynamicalSystems, Basins
+ds = Systems.rikitake(μ = 0.47, α = 1.0)
+xg=range(-6.,6.,length=200); yg=range(-6.,6.,length=200)
+pmap = poincaremap(ds, (3, 0.), Tmax=1e6; idxs = 1:2, rootkw = (xrtol = 1e-8, atol = 1e-8), reltol=1e-9)
+bsn = basin_poincare_map(xg, yg, pmap)
+```
+
+"""
+function basin_poincare_map(xg, yg, pmap, idxs = 1:2; Ncheck = 2)
+    i = typeof(idxs) <: Int ? i : SVector{length(idxs), Int}(idxs...)
+    # Carefully set the initial conditions on the defined plane and
+    reinit_f! = (integ,y) -> _initf(integ, y, i)
+    get_u = (pmap) -> pmap.integ.u[i]
+    basin = draw_basin!(xg, yg, pmap, step!, reinit_f!, get_u, Ncheck)
+
+end
+
+
+function _initf(pmap, y, idxs)
+    u = zeros(1,length(pmap.integ.u))
+    u[idxs] = y
+    # all other coordinates are zero
+    reinit!(pmap, u)
+end
+
+
+
+"""
+    basin_stroboscopic_map(xg, yg, integ; T=1., idxs=1:2)
+Compute an estimate of the basin of attraction on a two-dimensional plane using a stroboscopic map.
+
+## Arguments
+* `xg`, `yg` : 1-dim range vector that defines the grid of the initial conditions to test.
+* `integ` : integrator handle of the dynamical system.
+
+## Keyword Arguments
+* `T` : Period of the stroboscopic map
+* `idxs = 1:D` : Optionally you can choose which variables to save. Defaults to the entire state.
+"""
+function basin_stroboscopic_map(xg, yg, integ; T=1., idxs=1:2, Ncheck = 2)
+
+    i = typeof(idxs) <: Int ? i : SVector{length(idxs), Int}(idxs...)
+    iter_f! = (integ) -> step!(integ, T, true)
+    reinit_f! =  (integ,y) -> _init(integ, y, i)
+    get_u = (integ) -> integ.u[i]
+    return draw_basin!(xg, yg, integ, iter_f!, reinit_f!,get_u, Ncheck)
+
+end
+
+function basin_general_ds(xg, yg, integ; dt=1., idxs=1:2, Ncheck = 10)
+
+    return basin_stroboscopic_map(xg, yg, integ; T=dt, idxs=idxs, Ncheck = Ncheck)
+
+end
+
+
+function _init(integ, y, idxs)
+    u = zeros(length(integ.u))
+    u[idxs] = y
+    # all other coordinates are zero
+    reinit!(integ, u)
+end
+
+"""
+    basin_discrete_map(xg, yg, integ; idxs=1:2)
+Compute an estimate of the basin of attraction on a two-dimensional plane using a discrete map.
+
+## Arguments
+* `xg`, `yg` : 1-dim range vector that defines the grid of the initial conditions to test.
+* `integ` : integrator of the discrete system
+
+## Keyword Arguments
+* `idxs = 1:D` : Optionally you can choose which variables to save. Defaults to the entire state.
+"""
+function basin_discrete_map(xg, yg, integ; idxs=1:2)
+
+    i = typeof(idxs) <: Int ? i : SVector{length(idxs), Int}(idxs...)
+
+    iter_f! = (integ) -> step!(integ)
+    reinit_f! =  (integ,y) -> _init(integ, y, i)
+    get_u = (integ) -> integ.u[i]
+
+    return draw_basin!(xg, yg, integ, iter_f!, reinit_f!, get_u, 2)
+end
+
+
+
+
+
+
+## Procedure described in  H. E. Nusse and J. A. Yorke, Dynamics: numerical explorations, Springer, New York, 2012
+# The idea is to color the grid with the current color. When an attractor box is hit (even color), the initial condition is colored
+# with the color of its basin (odd color). If the trajectory hits another basin 10 times in row the IC is colored with the same
+# color as this basin.
+function procedure!(bsn_nfo::basin_info, n, m, u, Ncheck)
+    max_check = 60
+    next_c = bsn_nfo.basin[n,m]
+    bsn_nfo.step += 1
+
+
+    if iseven(next_c) && bsn_nfo.consecutive_match < max_check
+        # check wether or not we hit an attractor (even color). Make sure we hit Ncheck consecutive times.
+        if bsn_nfo.prev_attr == next_c
+            bsn_nfo.prevConsecutives +=1
+        else
+            # reset prevConsecutives
+            bsn_nfo.prev_attr = next_c
+            bsn_nfo.prevConsecutives =1
+            return 0;
+        end
+
+        if bsn_nfo.prevConsecutives >= Ncheck
+            # Wait if we hit the attractor a Ncheck times in a row just to check if it is not a nearby trajectory
+            #println("found IC")
+            c3 = next_c+1
+            ind = findall(bsn_nfo.basin .== bsn_nfo.current_color+1)
+            if Ncheck == 2
+                # For maps we can color the previous steps as well. Every point of the trajectory lead
+                # to the attractor
+                [ bsn_nfo.basin[k[1],k[2]] = c3  for k in ind]
+            else
+                # For higher dimensions we erase the past iterations.
+                [ bsn_nfo.basin[k[1],k[2]] = 1  for k in ind] # erase visited boxes
+            end
+            reset_bsn_nfo!(bsn_nfo)
+            return c3
+         end
+    end
+
+    if next_c == 1 && bsn_nfo.consecutive_match < max_check
+        # uncolored box, color it with current odd color
+        bsn_nfo.basin[n,m] = bsn_nfo.current_color + 1
+        bsn_nfo.consecutive_match = 0
+        return 0
+    elseif next_c == 1 && bsn_nfo.consecutive_match >= max_check
+        # Maybe chaotic attractor, perodic or long recursion.
+        # Color this box as part of an attractor
+        bsn_nfo.basin[n,m] = bsn_nfo.current_color
+        bsn_nfo.consecutive_match = max_check
+        #println("1 y > max_check")
+        return 0
+    elseif next_c == bsn_nfo.current_color + 1
+        # hit a previously visited box with the current color, possible attractor?
+        if bsn_nfo.consecutive_match < max_check
+            bsn_nfo.consecutive_match += 1
+            return 0
+        else
+            #println("got attractor")
+            #ind = findall(bsn_nfo.basin .== bsn_nfo.current_color+1)
+            #[ bsn_nfo.basin[k[1],k[2]] = 1 for k in ind]
+            bsn_nfo.basin[n,m] = bsn_nfo.current_color
+            push!(bsn_nfo.attractors, [bsn_nfo.current_color/2, u]) # store attractor
+            # We continue iterating until we hit again the same attractor. In which case we stop.
+            return 0;
+        end
+    elseif isodd(next_c) && 0 < next_c < bsn_nfo.current_color &&  bsn_nfo.consecutive_match < max_check && Ncheck == 2
+        # hit a colored basin point of the wrong basin, happens all the time, we check if it happens
+        #10 times in a row or if it happens N times along the trajectory whether to decide if it is another basin.
+        bsn_nfo.consecutive_other_basins += 1
+
+        if bsn_nfo.prev_bas == next_c &&  bsn_nfo.prev_step == bsn_nfo.step-1
+            bsn_nfo.prevConsecutives +=1
+            bsn_nfo.prev_step += 1
+        else
+            bsn_nfo.prev_bas = next_c
+            bsn_nfo.prev_step = bsn_nfo.step
+            bsn_nfo.prevConsecutives =1
+        end
+
+        if bsn_nfo.consecutive_other_basins > 60 || bsn_nfo.prevConsecutives > 10
+            ind = findall(bsn_nfo.basin .== bsn_nfo.current_color+1)
+            [bsn_nfo.basin[k[1],k[2]] = next_c for k in ind]
+            reset_bsn_nfo!(bsn_nfo)
+            return next_c
+        end
+        return 0
+    elseif iseven(next_c) &&   (max_check <= bsn_nfo.consecutive_match < 2*max_check)
+        # We make sure we hit the attractor 60 consecutive times
+        bsn_nfo.consecutive_match+=1
+        return 0
+    elseif iseven(next_c) && bsn_nfo.consecutive_match >= max_check*2
+        # We have checked the presence of an attractor: tidy up everything and get a new box.
+        ind = findall(bsn_nfo.basin .== bsn_nfo.current_color+1)
+        [ bsn_nfo.basin[k[1],k[2]] = 1 for k in ind]
+        bsn_nfo.basin[n,m] = bsn_nfo.current_color
+        bsn_nfo.current_color = bsn_nfo.next_avail_color
+        bsn_nfo.next_avail_color += 2
+        #println("even and > max check ", next_c)
+        reset_bsn_nfo!(bsn_nfo)
+        return next_c+1;
+    else
+        return 0
+    end
+end
+
+
+"""
+    draw_basin(xg, yg, integ, iter_f!::Function, reinit_f!::Function)
+Compute an estimate of the basin of attraction on a two-dimensional plane. This is a low level function,
+for higher level functions see: `basin_poincare_map`, `basin_discrete_map`, `basin_stroboscopic_map`
+
+## Arguments
+* `xg`, `yg` : 1-dim range vector that defines the grid of the initial conditions to test.
+* `integ` : integrator handle of the dynamical system.
+* `iter_f!` : function that iterates the map or the system, see step! from DifferentialEquations.jl and
+examples for a Poincaré map of a continuous system.
+* `reinit_f!` : function that sets the initial condition to test on a two dimensional projection of the phase space.
+"""
+function draw_basin!(xg, yg, integ, iter_f!::Function, reinit_f!::Function, get_u::Function, Ncheck)
+
+    complete = 0;
+
+    bsn_nfo = basin_info(ones(Int16, length(xg), length(yg)), xg, yg, iter_f!, reinit_f!, get_u, 2,4,0,0,0,1,1,0,0,[])
+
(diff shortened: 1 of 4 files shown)


Datseris:
Is there any functional difference between the "poincare map" and "discrete map"? In essence they are both discrete dynamical systems, right? perhaps we should have only a single function and the user should not care weather they have a discrete system or a PoincareMap. You can still use multiple dispatch if you want two difference functions since we have the type `PoincareMap`.

awage:
You are absolutely right. I have simplified the interface and remove some useless code. It is neater now. 

I have to work on some issues with typing. For example I do not know if these function should be typed more concretely or if the compiler takes care of it:

```
function _init_map(pmap::PoincareMap, y, idxs)
function _init_ds(integ, y, idxs)
function procedure!(bsn_nfo::basin_info, n::Int, m::Int, u, Ncheck::Int)

```

Datseris:
Great thanks. Please allow me some time; at the moment I'm drowning in work and this PR will need a bit of time to review properly. I hope to do this within the next two weeks but if not do not worry, it is on my to-do list!

awage:
Sure. There is no rush. In the meanwhile I have seen some glitches I have to polish.

Datseris (commented):
Is the resource you used for this somewhere online? I'm trying to understand the central struct here and it is not easy. I'm getting a bit confused with the way the color coding works.

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,370 @@
...
+end
+
+
+"""
+    basin_map(xg, yg, integ; kwargs...)
+Compute an estimate of the basin of attraction on a two-dimensional plane using a map of the plane onto itself.
+The dynamical system should be a discrete two dimensional system such as:
+    * Discrete 2D map.
+    * 2D poincaré map.
+    * A 2D stroboscopic map.
```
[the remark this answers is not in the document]
awage: This method has been developed initially for 2D maps. And it works very well for these systems. You can use the same method for higher order system but it is less reliable has you can miss an attractor if you don't have the right projection on the plane. Also the result can depend on the time step for some systems. 

The algorithm is a finite state machine. You start coloring with the first available color and as soon as you have found a new attractor you pick the next one. I kept the original names of the program but this can be improved. 


I drop you a file with the basics of the algorithm. The original program is available and still compiles fine. However it is a old school C program that costs hundreds of hours to decypher. By the way there are cool methods for periodic orbits, but that's another PR.  

http://yorke.umd.edu/dynamics/
[salida.pdf](https://github.com/JuliaDynamics/ChaosTools.jl/files/6363687/salida.pdf)

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,370 @@
...
+    else
+        return 0
+    end
+end
+
+
+"""
+    draw_basin!(xg, yg, integ, iter_f!::Function, reinit_f!::Function)
+Compute an estimate of the basin of attraction on a two-dimensional plane. This is a low level function,
+for higher level functions see: `basin_poincare_map`, `basin_discrete_map`, `basin_stroboscopic_map`
```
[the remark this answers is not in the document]
awage: yes they have.

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,374 @@
+export draw_basin!, basin_map, basin_general_ds
+
+
+mutable struct BasinInfo{I,F,V}
+    basin :: I
```
Datseris, line 5: Please remove the spaces between the letters and the `::` for stylistic similarity with the rest of the library

Datseris (commented):
@awage sorry for being so slow here, I have a lot of work on outside projects. Something that would help me a lot would be if you provided some kind of documentation in the sense of examples, of how one would go about using this functionality. 

For example, think of the documentation pages of the [DynamicalSystems.jl docs](https://juliadynamics.github.io/DynamicalSystems.jl/latest/chaos/fractaldim/) . Can you write a simple similar markdown-based page, and push it here into the `basin` folder? We can then use this and move it to the DynamicalSystems.jl repo as documentation. 

There you can set up a simple dynamical system, walk us through the thought process of getting basins of attraction, and plotting them. (Notice: I believe the interface is really straightforward here, but we need to make such a documentation page anyways to actually add this feature to the docs, and it would help me a lot if you did a first starting draft. Use one of your many well-defined tests as a starting point).

Another question: the numbers in the `I` matrix of the `BasinInfo` do not have any particular meaning, besides counting different basins and attractors, right? I run the example you have with the stroboscopic map of the duffing oscillator, and `unique(bsn.basin)` provides:
```
julia> unique(ans)
4-element Vector{Int16}:
 3
 5
 2
 4
```

Why is there no `1` number? Can you please explain how this numbering occurs?

Datseris:
Furthermore, can you please add a concise, yet illustrative, description of the actual algorithm used for the detection in the docstring of the main `basin` function? If you have a look at functions like `lyapunovspectrum` or `gali`, they provide some information on how the algorithm works. You already have this information in the source code as comments:
```

## Procedure described in  H. E. Nusse and J. A. Yorke, Dynamics: numerical explorations, Springer, New York, 2012
# The idea is to color the grid with the current color. When an attractor box is hit (even color), the initial condition is colored
# with the color of its basin (odd color). If the trajectory hits another basin 10 times in row the IC is colored with the same
# color as this basin.
```

but it should be in the docstring instead.

Datseris:
And final request: The tuning parameters of the algorithm should be tunable keyword arguments from the user. For example, you say that if a trajectory hits same basin 10 times in a row, you get the same color. this 10 should be a named variable and a keyword argument.

Datseris:
Another question: the `BasinInfo.basin` contains both the attractors (odd numbers) and their basins (even numbers). 

1. This is not explained in the docstring, and I do not think it is intuitive to a user without explicit explanation. 
2. Does this even make sense? For e.g. the duffing system, the attractors are points. They do not have any area value. So coloring a pixel as the attractor may not be optimal. On the other hand, for non-fixed point attractors, e.g. a limit cycle, this might be a better idea nevertheless. Do you have any example btw that converges to a limit cycle that we could show? Or, we could show convergence to chaotic attractors. Let's grab a simple 3D chaotic system that has 2 coexisting attractors and see what our algorithm outputs?

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,374 @@
...
+pmap = poincaremap(ds, (3, 0.), Tmax=1e6; idxs = 1:2, rootkw = (xrtol = 1e-8, atol = 1e-8), reltol=1e-9)
+bsn = basin_map(xg, yg, pmap)
+```
+
+"""
+function basin_map(xg, yg, pmap::PoincareMap; Ncheck = 2)
+    reinit_f! = (pmap,y) -> _init_map(pmap, y, pmap.i)
+    get_u = (pmap) -> pmap.integ.u[pmap.i]
+    basin = draw_basin!(xg, yg, pmap, step!, reinit_f!, get_u, Ncheck)
+end
```
Datseris, line 67: Is there any safety handling happening for input `PoincareMap` that exceeds its `Tmax`? This map is a bit special in the sense that `step!(pmap)` will return `nothing` if the map is evolved for more than its maximum time.

Datseris:
image

This coloring scheme explanation **must** be included in the documentation string, perhaps in a bit more concise form.

awage:
Hi! I made some improvements on the doc string and also pushed a markdown file with examples to illustrate the functionalities. As you said the results are not intuitive at first sight but the structure of the matrix makes sense when explained. There is a problem when Poincaré map are not well defined. I suggest to throw a warning or an error when this happens. 

There is a thing with diverging trajectories. For some systems it is considered a "attractor", these are the escape basins. It is possible to include a function called `escapes` that would track these outputs. As an example: 

[hunt1999.pdf](https://github.com/JuliaDynamics/ChaosTools.jl/files/6417959/hunt1999.pdf)

The map described goes to +/- Inf for every initial conditions. Also some Hamiltonian dynamical systems have escapes in different direction for every initial conditions: 

[BarrioBlesaSerrano-2008-FractalstructuresintheHnon-HeilesHamiltonian.pdf](https://github.com/JuliaDynamics/ChaosTools.jl/files/6418008/BarrioBlesaSerrano-2008-FractalstructuresintheHnon-HeilesHamiltonian.pdf)
 
This can be included as an additional feature or can be a separated function. It would be 10 lines of codes on its own, but then the basin can be analyzed with other tools (fractal dimension and so on). 

Take your time there is no hurry.

src/orbitdiagrams/poincare.jl
```diff
@@ -155,8 +154,11 @@ function poincaremap!(integ, f, planecrossing, Tmax, idxs, rootkw)
         step!(integ)
         side = planecrossing(integ.u)
     end
- 	integ.t > Tmax && return nothing
 
+	if integ.t > Tmax
+		@warn "The Poincaré map is ill defined"
```
Datseris, line 159:
```suggestion
		@warn "The Poincaré map is ill defined or time evolution exceeded `Tmax`."
```

Datseris:
@awage you are doing a wonderful work here, thanks a lot! I improved the docs a bit. I will improve further. For now I have some suggestions:

1. I think the functions should be renamed into `basins_map2D` and `basins_general`. Notice also the plural.
2. I've changed the return signature of the basin functions. They now return the basins matrix and the attractors vector. I don't think the users care about the `BasinInfo` struct so we can skip it.
3. I think you should allow the users to give you a function for `basins_general` that provides values for the other variables of the higher dimensional system, instead of always using the values 0. You could allow for a function `f(x, y) -> remaining_variables` with `x, y` the current grid point.
4. I've run the duffing example. Its output `attractors` is: 
```
julia> attractors
150-element Vector{Any}:
 [1.0, -0.7767200433871121, -0.015674280248876037]
 [1.0, -0.7798948287686913, -0.032079859801086484]
```
Why is the element type `Any`, and why are the points three-dimensional? I guess the 3rd coordinate is time. but then, why is it negative...?
5. Have you checked your code for performance? I think we might have a case of unoptimized code here. This `Any` I saw above is quite scary: you really shouldn't have data structurs with this element type. The `procedure!` function is really what we care about: it does the crux of the work. Have you done any basic profiling on this function to see if anything in performance is off? E.g. some line taking unecessarily longer than it should? ProfileViews.jl can provide a simple plot that can help you with that.


> There is a thing with diverging trajectories. For some systems it is considered a "attractor", these are the escape basins

My understanding so far is that your code already takes care of diverging trajectories, attributing them to attractor `-1`. So what's the problem? 

> This can be included as an additional feature or can be a separated function. It would be 10 lines of codes on its own, but then the basin can be analyzed with other tools (fractal dimension and so on).

Yes this is a totally fine idea. but I haven't understood yet why it is necessary, since `basins` already tracks diverging trajectories...?

Datseris:
@awage can you please tell me your name and/or affiliation as well? I'd like to put you in the shoutouts document

awage:
Hi, I have made some changes: 


>     1. I think the functions should be renamed into `basins_map2D` and `basins_general`. Notice also the plural.
Done. I also changed the docs a little bit but it will require proofreading when the code is stable. 

>     2. I've changed the return signature of the basin functions. They now return the basins matrix and the attractors vector. I don't think the users care about the `BasinInfo` struct so we can skip it.
This is OK. If needed there is the low-level function `draw_basin!` that can be useful for strange cases. 
 
>     3. I think you should allow the users to give you a function for `basins_general` that provides values for the other variables of the higher dimensional system, instead of always using the values 0. You could allow for a function `f(x, y) -> remaining_variables` with `x, y` the current grid point.

It was on my mind at some point and I forgot to include this feature. I made a change allowing a constant vector for the other coordinates. It is simpler. However, maybe you have something on your mind like defining an arbitrary plane? In this case a function is necessary. 

>     4. I've run the duffing example. Its output `attractors` is ... 
I have changed this part, it was not consistent the basin structure. Now `attractors` is a dictionary with the keys corresponding to the attractor number. The value is a `Dataset` structure with a list of points corresponding to the attractor. There might be very close points in this array. Also the attractor may not be complete, this is the case for chaotic and quasiperiodic attractors. This part can be improved. Another possibility is to return the coordinates of the pixels on the map,but it is less accurate. 

> 
>     1. Have you checked your code for performance? I think we might have a case of unoptimized code here. 

I solved the issue with Any[] with a typed dictionary. I have also profiled the code: 
* For ODEs, the code spends about 90-95% of the time in solving the ODE. 
* For maps, iteration of the map is less than 1% of the total time. There is a line that can be improved without much effort: 
```jl        
 # pick the first empty box
 get_empt_box = findall(bsn_nfo.basin .== 1)
```

It looks for the first box with 1 inside. This code is very inefficient. I will have a though about it. 

In the `procedure!` function the `findall` calls take almost all the computation time. It can be improved but I don't think it is worth it since it would require more memory space and vector allocations. 

> My understanding so far is that your code already takes care of diverging trajectories, attributing them to attractor `-1`. So what's the problem?

For some dynamical systems you have diverging trajectory to +Inf or -Inf, and the direction matters. For now the code only says that it has escaped from the grid. But it can be an independent PR since it is a different approach.

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,418 @@
+export draw_basin!, basins_map2D, basins_general
+
+
+mutable struct BasinInfo{I,F,D}
+    basin::I
```
Datseris, line 5: To my understanding `basin` is always `Matrix{Int16}` so we don't need a type parameterization, right?

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,418 @@
...
+    current_color::Int64
+    next_avail_color::Int64
+    consecutive_match::Int64
+    consecutive_other_basins::Int64
+    prevConsecutives::Int64
+    prev_attr::Int64
+    prev_bas::Int64
+    prev_step::Int64
+    step::Int64
+    attractors::D
```
Datseris, line 20: I would be more specific here. Instead of Parameterizing `attractors` as `D` I would parameterize as `Dict{Int16, Dataset{2, T}`. `T` is the numeric element type.

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,418 @@
...
+        # hit a previously visited box with the current color, possible attractor?
+        if bsn_nfo.consecutive_match < max_check
+            bsn_nfo.consecutive_match += 1
+            return 0
+        else
+            bsn_nfo.basin[n,m] = bsn_nfo.current_color
+            if haskey(bsn_nfo.attractors , bsn_nfo.current_color)
+                push!(bsn_nfo.attractors[bsn_nfo.current_color],  u) # store attractor
+            else
+                bsn_nfo.attractors[bsn_nfo.current_color] = Dataset([u[1] u[2]]) # init dic
```
Datseris, line 237: Here `[u[1] u[2]]` is not necessary as it makes an intermediate matrix. Instead do `Dataset([SVector(u[1], u[2])])`

awage:
> @awage can you please tell me your name and/or affiliation as well? I'd like to put you in the shoutouts document

Here you go: 

Alexandre Wagemakers  
Universidad Rey Juan Carlos
Nonlinear Dynamics, Chaos and Complex Systems Group

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,418 @@
...
+         end
+
+         ni = get_empt_box[1][1]; mi = get_empt_box[1][2]
+         x0 = xg[ni]; y0 = yg[mi]
+
+         # Tentatively assign a color: odd is for basins, even for attractors.
+         # First color is one
+         bsn_nfo.basin[ni,mi] = bsn_nfo.current_color + 1
+
+         u0=[x0, y0]
```
Datseris, line 323: I've checked the `procedure!` function. In there `u0` is never modified. Therefore, why do we make it a `Vector`? We should instead make it a static vector as `u0 = SVector(x0, y0)`.

Furthermore, could you please replace the symbol `u` with `u0` in the function `procedure!`? It misslead me to think it was a mutable current state, instead of an immutable initial condition.

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,418 @@
...
+
+    reset_bsn_nfo!(bsn_nfo)
+
+    while complete == 0
+         # pick the first empty box
+         get_empt_box = findall(bsn_nfo.basin .== 1)
+         if length(get_empt_box) == 0
+             complete = 1
+             break
+         end
```
Datseris, line 314: Your approach here is highly inefficient. In essence you want the index of the first element of `basin` that has the value `1`. That is much, much better done with the code:
```julia
j = findfirst(isequal(1), bsn_info.basin)
if isnothing(j)
    complete = 1
    break
end

ni = j[1]; mi = j[2]
```
This avoids a huge amount of overhead: first, it does not need to create a boolean matrix via the operation `bsn_nfo.basin .== 1`. Second, it does not have to check all the elements of `basin`. Only in the case where no element is 1 then all elements are checked. Third, it does not have to identify all vectors that satisfy `basin == 1`. 

In fact, you can improve this even further. If you store somewhere the index of the previously initialized index of `basin == 1`, you can start your search for a non-1 cell based on the previous index. For this, you scan the `basin` linearly, as if it is a vector, instead of as if it is a matrix, using `vec(basin)` (do not worry, this does not allocate).

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,418 @@
...
+    return n,m
+end
+
+
+function check_outside_the_screen!(bsn_nfo::BasinInfo, new_u, old_u, inlimbo)
+
+    if norm(new_u-old_u) < 1e-5
+        #println("Got stuck somewhere, Maybe an attractor outside the screen: ", new_u)
+        ind = findall(bsn_nfo.basin .== bsn_nfo.current_color+1)
+        [ bsn_nfo.basin[k[1],k[2]] = 1  for k in ind]
```
Datseris, line 393: ??? What's going on here? Why do you create a vector with the `[ ... ]` syntax if you will never use this vector? Instead change this line to:

```
for k in ind; bsn_nfo.basin[k[1],k[2]] = 1; end
```

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,418 @@
...
+        reset_bsn_nfo!(bsn_nfo)
+        # this CI goes to a attractor outside the screen, set to -1 (even color)
+        return -1  # get next box
+    elseif inlimbo > 60*20
+        #println("trajectory diverges: ", new_u)
+        ind = findall(bsn_nfo.basin .== bsn_nfo.current_color+1)
+        [ bsn_nfo.basin[k[1],k[2]] = 1  for k in ind]
```
Datseris, line 400: same as above.

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,418 @@
...
+* `integ` : integrator handle of the dynamical system.
+* `iter_f!` : function that iterates the map or the system, see step! from DifferentialEquations.jl and
+examples for a Poincaré map of a continuous system.
+* `reinit_f!` : function that sets the initial condition to test on a two dimensional projection of the phase space.
+"""
+function draw_basin!(xg, yg, integ, iter_f!::Function, reinit_f!::Function, get_u::Function, Ncheck)
+
+    complete = 0;
+
+    bsn_nfo = BasinInfo(ones(Int16, length(xg), length(yg)), xg, yg, iter_f!, reinit_f!, get_u, 2,4,0,0,0,1,1,0,0,Dict{Int16,Dataset{2,Float64}}())
```
Datseris, line 304: Here instead of `Dataset{2, Float64}` as a parameter I'd use `Dataset{2, eltype(integ.u)}` so that the element type of the variables is preserved

src/basins/compute_ba.jl
```diff
@@ -0,0 +1,418 @@
...
+            bsn_nfo.prev_step += 1
+        else
+            bsn_nfo.prev_bas = next_c
+            bsn_nfo.prev_step = bsn_nfo.step
+            bsn_nfo.prevConsecutives =1
+        end
+
+        if bsn_nfo.consecutive_other_basins > 60 || bsn_nfo.prevConsecutives > 10
+            ind = findall(bsn_nfo.basin .== bsn_nfo.current_color+1)
+            for k in ind; bsn_nfo.basin[k] = next_c; end
```
Datseris, line 258: Okay this is certainly necessary to create the `ind` array. In Julia, loops are fast, you don't have to make arrays. I propose to change this to:
```julia
for k in eachindex(bsn_nfo.basin)
if bsn_nfo.basin[k] == bsn_nfo.current_color+1
bsn_nfo.basin[k] = next_c
end
end
```

Datseris (commented):
@awage great, looks better and better. I've left some really useful comments regarding performance. Especially your `findall` problems. I think not a single `findall` is necessary. I told you how to get read of 2/3 performance critical `findall` instances. I recommend that you incorporate my change, and do a profile, for every replacement of `findall`. This will give you an impression of how much impact this has!

Datseris:
> Alexandre Wagemakers

Oh, _wow_ an assistant professor is contributing to DynamicalSystems.jl, we are quite honored! 

Off-topic, but looking at your homepage you might like reading this: https://juliamusic.github.io/JuliaMusic_documentation.jl/dev/blog/drumpatterns/

src/basins/compute_ba.jl
Datseris: Oh yeah, this needs to be fixed critically, because it hits the warning:
```
┌ Warning: You are attempting to make a Dataset of a matrix with more columns than rows.
└ @ DelayEmbeddings C:\Users\datse\.julia\dev\DelayEmbeddings\src\dataset.jl:227
```
and warnings typically have massive hits on performance

awage:
> Oh, _wow_ an assistant professor is contributing to DynamicalSystems.jl, we are quite honored!
Pleasure is mine!  This is fun to work on this project. 

> Off-topic, but looking at your homepage you might like reading this: https://juliamusic.github.io/JuliaMusic_documentation.jl/dev/blog/drumpatterns/
Nice! I wish I had the time to study again patterns and stuff like that. Good work there!

src/basins/compute_ba.jl
awage: I am not sure to understand clearly: u is the current state in `procedure!`. It is used to store the attractor points if detected.

I will change u0 to a SVector

src/basins/compute_ba.jl
awage: I have found a solution with CartesianIndex. It is not very elegant but there is a huge improvement in speed.

src/basins/compute_ba.jl
awage: I missed this one.

src/basins/compute_ba.jl
awage: yep.

src/basins/compute_ba.jl
awage: I have tried this solution and the results are not conclusive. It doesn't bring a clear improvement. For now I will leave the original code unless I come across something better.

Datseris: Okay, but also notice that you do not need `findall`. You can do:
```
ind = bsn_nfo.basin .== bsn_nfo.current_color+1
bsn_nfo.basin[ind] .= next_c
```

src/basins/compute_ba.jl
Datseris: It does, but you cannot measure it because it is insignificant when compared to the rest of the code.

Datseris:
@awage I have a final question. There is no need for `basins` to be a Matrix, right? I am thinking of situations where a user will want to provide a _vector of coordinates_ `xy = [(0.1, 0.1), (0.2, 0.2), ...]` instead of making an _orthogonal grid_ of initial conditions. Reasons why I think this would be useful is to have a higher precision at some parts of the plane.

To my understanding, the algorithm does not really care whether basins is a matrix or a vector (with a value at each coordinate tuple), as it iterates over its indices in a linear manner, correct? Then, after we merge this PR we should open a Feature Request issue that asks for `basins_...` to allow `xy` as coordinate vector.

awage:
> @awage I have a final question. There is no need for `basins` to be a Matrix, right? 

It does not need to be a matrix but you have to define a grid of some sort. The function `get_box` returns the closest initial condition assuming a rectangular grid. This function can be adapted to all sort of grid. 

I have thought about this kind of multi-resolution basins. It can be an adaptative mesh that can be refined near the boundary. It would require some sort of iterative process: 
* First compute the basin on a coarse grid. 
* Find the boundary and refine the mesh on this region
* Compute the basin on this new grid. 

 
> To my understanding, the algorithm does not really care whether basins is a matrix or a vector (with a value at each coordinate tuple), as it iterates over its indices in a linear manner, correct? Then, after we merge this PR we should open a Feature Request issue that asks for `basins_...` to allow `xy` as coordinate vector.

Totally possible, it can be a linear vector with a set of points. But the critical issue is to define a grid that covers the region of interest.

PR #168, opened 2021-04-19
Fixed to boxed correlationsum

Datseris:
@apbraun can you help here. You say:

The best choice given by Theiler is $P = 0.5\log_2N$ and should be used when $D$ exceeds $0.75\log_2N$.

Is the last part correct? And if D is not greater than this, do we return size(data, 2)?


Datseris:
The function `find_neighborboxes_q` does not use its argument `q`.

Datseris:
@apbraun I've noticed that the boxed_correlation sum for q != 2 becomes really slow... I guess this function has never been benchmarked, profiled and/or optimized, right?

apbraun:
It's complicated. Theiler claims that $P = 0.5\log_2(N)$ is the fastest while Bueno-Orovio and García-Pérez claim that they always use 2 for dimensions bigger than two. I've run some low effort measurements and found that Theilers choice wasn't the fastest while a fixed choice of 2 ran into partition problems for high dimensional data.

apbraun:
The correlationsum for q != 2 is much harder to optimise and known to be less efficient. Since the number of all points is raised to a power unequal to 1, we need to find every single point instead of only the next points that are within radius of that point. This applies to the search for boxes as well as the correlationsum. Thereby we need to find more boxes and calculate more points in the inner correlationsum. I thought it would still scale better than the ordinary q-order correlationsum.

Datseris:
> heiler claims that $P = 0.5\log_2(N)$ is the fastest while Bueno-Orovio and García-Pérez claim that they always use 2 for dimensions bigger than two. I've run some low effort measurements and found that Theilers choice wasn't the fastest while a fixed choice of 2 ran into partition problems for high dimensional data.

Yes , okay but is Theiler's P faster than using `size(data, 2)`?

apbraun:
Yes it's faster than `size(data,2)`

Datseris:
>  when $D$ exceeds $0.75\log_2N$.

This will literally never happen. Even for 1000 datapoints, 
```
julia> 0.75log2(1000)
7.474338213496566
```

apbraun:
True. On the other hand the for 1.000.000 points the value is 12.45, since the logarithm is growing so slowly.

Datseris:
I think you didn't get my point. Theiler's suggestion will never happen in practice, because $D$ will never _exceed_ this amount. are you sure it is when D exceeds this amount, and not when D is smaller than this amount?

Issue #121, opened 2020-08-19
Fractal basin boundary dimension estimation: final state uncertainty


Datseris:
This paper: https://www.sciencedirect.com/science/article/pii/0375960183909453?via%3Dihub describes a method to estimate fractal dimensions of fractal basins boundary of attraction, which is useful because it is very hard to get the precise form of the boundary and then use stuff like `genentropy` (because you typically have all states that converge to an attractor, not just their boundary).

Implementation sketch:

The way I imagine it is with either by initializing several integrators in parallel (what EnsembleProblem does) or by doing reinit! for every new initial condition (what I typically do in DynamicalSystems.jl, because you typically parallelize at a higher level, e.g. for different parameters).

The user can provide a terminate function that checks the state of the integrator and terminates integration on request (e.g. terminate(integ) = norm(integ.u[3:4]) < 1e-3 a typical condition that terminates when velocity becomes too small). Once terminated, you check the final state of the integrator with another user provided function that returns `true` if the integrator has reached the fixed point you care about, or `false` for something else (and thus the point is "uncertain").

Regarding optimization, the function can take a series of sizes ε as input, because for ε2 > ε1 you don’t have to recompute what initial conditions starting in ball ε1 do; you have already done that in the ε1 step.

Lachlan-Bn:
Another optimization may be how I set out my example in https://discourse.julialang.org/t/ensemble-simulations-with-paired-trajectories/45194. I.e, that instead of integrating N trajectories for each value of ε and taking the ratio of "uncertain" points to N, you instead integrate until you find a certain number of uncertain points (a colleague suggests ~10) and take 10/N* as the estimate for `f`, where N* is the number of trajectories you needed. 

In problems where the basin geometry is very complicated (the magnetic pendulum may be a good example), this would ensure the integration doesn't spend too long on "large" values of ε where the perturbed point is more likely to be in a different basin. It would also, I think, have the bonus of better controlling the error in the estimator.

Datseris:
The magnetic pendulum is implemented https://github.com/JuliaDynamics/DynamicalSystemsBase.jl/blob/master/src/continuous_famous_systems.jl#L736  , here is a figure I made with it:

![magneticpendulum_full](https://user-images.githubusercontent.com/19669089/90632411-a8634700-e224-11ea-9e79-36f62b20387d.png)

[...]

PR #172, opened 2021-05-17
Uncertainty exponent

awage:
Hi! This is the part that was in the opened issue #121 in the first place. This is the estimation of the uncertainty dimension with the original algorithm in:  
* C. Grebogi, S. W. McDonald, E. Ott and J. A. Yorke, Final state sensitivity: An obstruction to predictability, Physics Letters A, 99, 9, 1983

The idea is to look along one of the dimension, the boxes are lines along the x or y axis.  This is done in order to avoid problems with different resolutions along each dimension. The theory says it should converge for any ball. 

The basin is sampled with a box of a given size and the ratio `f_ε` of boxes that contain at least two different "colors" is evaluated. The basin is sampled with some bootstrapping since it is a random event with replacement. This is a binomial process and the variance of the estimator of the mean can be evaluated. Once this variance is low enough (the mean has converged), we increase the size of the box and we go on with the next value of the ratio. This part is not in the original algorithm but some sort of stopping criterion must be introduced in order to have good estimates. 

At the end the slope of `f_ε` as a function of `ε` is evaluated in log-log scale with the function `linear_region`. 

I have written two tests with known value of this exponent. The value from the original paper cited above and the newton fractal with three basins. For this latest example, the value of the box counting dimension published is 1.42. The algorithm returns around 0.6. Which is not bad. 

I have considered using the function `generalized_dim` to evaluate the box counting dimension of the boundary but it is quite difficult to obtain a reliable value (d = 2 - α  with d the box counting dim and α is the uncertainty exponent) . It is not easy to get exactly the points of the boundary. If you are interested I can include the code also. 

I think this PR is more straightforward although sure there are lot of things to improve.


src/basins/uncertainty_exp.jl
```diff
@@ -0,0 +1,97 @@
...
+    for (k,eps) in enumerate(r_ε)
+        Nb=0; Nu=0; μ=0; σ²=0; M₂=0;
+        completed = 0;
+        # Find uncertain boxes
+        while completed == 0
+            kx = rand(1:nx)
+            ky = rand(ceil(Int64,eps+1):floor(Int64,ny-eps))
+
+            indy = range(ky-eps,ky+eps,step=1)
+            c = [basins[kx,ky] for ky in indy]
```
Datseris, line 51:
```suggestion
            c = basins[kx, indy]
```

test/basins/uncertainty_tests.jl
```diff
@@ -0,0 +1,71 @@
...
+        dz[2]=-J₀*cos(θ)
+        return
+    end
+
+    # dummy function to keep the initializator happy
+    function grebogi_map_J(J,z0, p, n)
+       return
+    end
+    ds = DiscreteDynamicalSystem(grebogi_map,[1., -1.], [] , grebogi_map_J)
+    integ  = integrator(ds)
```
Datseris, line 25: You can make a PR to DynamicalSystemsBase.jl to add this to the available systems!

src/basins/uncertainty_exp.jl
```diff
@@ -0,0 +1,97 @@
+export uncertainty_exponent
+
+"""
+    uncertainty_exponent(xg, yg, basins, integ; precision=1e-4, max_size=0) -> α,ε,f_ε
+This function estimates the uncertainty exponent of the basins.
```
Datseris, line 5: It is not clear here what `basins` are. Is it the output of the function `basins_general`?

src/basins/uncertainty_exp.jl
```diff
@@ -0,0 +1,97 @@
...
+The output `f_ε` is the ratio of the ball of radius `ε` that contains at least two initial conditions that lead to different attractors. `α` is the slope of the curve `f_ε` against `ε`. If the precision or the resolution is not high enough, the function gives a warning and a visual inspection of this curve is required in order to estimate the slope correctly.
+
+Notice that the uncertainty exponent and the box counting dimension of the boundary are related. We have `d = 2 - α` where `d` is the box couting or capacity dimension.
+
+[^Grebogi1983]: C. Grebogi, S. W. McDonald, E. Ott and J. A. Yorke, Final state sensitivity: An obstruction to predictability, Physics Letters A, 99, 9, 1983
+
+## Arguments
+* `basin` : the matrix containing the information of the basin.
+* `xg`, `yg` : 1-dim range vector that defines the grid of the initial conditions.
```
Datseris, line 16: For stylistic similarity to the rest of the library, there should not be an "Arguments" section. Non-keyword arguments should be explained in the opening sentences of the documentation string.

src/basins/uncertainty_exp.jl
```diff
@@ -0,0 +1,97 @@
+export uncertainty_exponent
+
+"""
+    uncertainty_exponent(xg, yg, basins, integ; precision=1e-4, max_size=0) -> α,ε,f_ε
+This function estimates the uncertainty exponent of the basins.
+
+The ouput `α` is the estimation of the uncertainty exponent of the basins of attraction. This exponent is related to the final state sensitivity of the trajectories in the phase space. An exponent close to `1` means basins with smooth boundaries whereas an exponent close to `0` represent complety fractalized basins called a riddled basins.
+The output `f_ε` is the ratio of the ball of radius `ε` that contains at least two initial conditions that lead to different attractors. `α` is the slope of the curve `f_ε` against `ε`. If the precision or the resolution is not high enough, the function gives a warning and a visual inspection of this curve is required in order to estimate the slope correctly.
+
+Notice that the uncertainty exponent and the box counting dimension of the boundary are related. We have `d = 2 - α` where `d` is the box couting or capacity dimension.
```
Datseris, line 10: Please put this under a section `## Description` for stylistic similarity to the rest of the library.

src/basins/uncertainty_exp.jl
```diff
@@ -0,0 +1,97 @@
...
+
+    # remove zeros in case there are:
+    ind = f_ε .> 0.
+    f_ε =  f_ε[ind]
+    ε = ε[ind]
+    # get exponent
+    # a,D =  linreg(vec(log10.(ε)), vec(log10.(f_ε)))
+    D = linear_region(vec(log10.(ε)), vec(log10.(f_ε)))
+    return D[2], vec(log10.(ε)), vec(log10.(f_ε))
+end
```
Datseris, line 86: In DynamicalSystems.jl v2.0, which will be released soon, I will stop automating so much this "automatic fit to linear slope" from all functions. I think it is not a good idea to do this without "consent", and this is motivated by further research (currently unpublished) on estimating fractal dimensions.

To combat this, I propose to simply change the order of arguments to return `ε, f_ε, α`.

Datseris:
Thanks a lot @awage !!! I'll start a PR at DynamicalSystems.jl that adds this to the docs, so we can see how it goes! I think the only comments I left are minor and concern style mostly.

awage:
Before merging anything there are some issue I wanted to comment. Allow me
some time until tomorrow!

src/basins/uncertainty_exp.jl
awage: In this case it should be the result of the computation of `basins_map2D` or `basins_general`. 

In forgot to remove the attractors from the computation. I will fix that

src/basins/uncertainty_exp.jl
awage: Ok, I will stick to the guidelines.

src/basins/uncertainty_exp.jl
awage: It is a point that I wanted to talk about. As you can see the line 83 uses the linear regression routine without any estimation of the best linear region. I was not sure about what was the best solution but now I ran some statistics and the linear regression with all the datas is more consistent than estimating the best linear region. 

The first graph is the estimation of \alpha with the `linear_region` function.  The `x` axis is the `precision`. The second graph uses the function `linreg`. You can see that the results are way better when you relax the `precision` variable (alpha should be around 0.6)   
imagen
imagen

Another interesting point about regression is that you can estimate the error of the slope also. It can be an extra value.

Datseris: How did you calculate the error bars @awage ?

src/basins/uncertainty_exp.jl
```diff
@@ -0,0 +1,105 @@
...
+
+## Keyword arguments
+* `precision` is the variance of the estimator of the uncertainty function. Values between 1e-7 and 1e-5 brings reasonable results.
+* `max_size` is the maximum size in pixels of the ball to test.
+
+## Description
+
+A phase space with a fractal boundary may cause a uncertainty on the final state of the dynamical system for a given initial condition. A measure of this final state sensitivity is the uncertainty exponent. The algorithm probe the basin of attraction with balls of size `ε` at random. If there are a least two initial conditions that lead to different attractors, a ball is tagged `uncertain`. `f_ε` is the fraction of `uncertain ball` to the total number of tries in the basin. In analogy to the capacity dimension, there is a scaling law between `f_ε` and `ε`. The number that characterizes this scaling is called the uncertainty exponent `α`.
+
+Notice that the uncertainty exponent and the box counting dimension of the boundary are related. We have `d = 2 - α` where `d` is the box couting or capacity dimension.
```
Datseris, line 23: oh yeah, sorry forgot this delay: you need to limit source code lines to max 92 characters... this is true even for docstrings. In general my advice is to not wrap lines in your editor if you have source files. This is because break line can have meaningful impact to code execution, so best avoid "apparent" breaklines via wrapping.

src/basins/uncertainty_exp.jl
```diff
@@ -0,0 +1,105 @@
...
+function uncertainty_exponent(xg, yg, basins::Matrix; precision=1e-4, max_size=0)
+
+    nx=length(xg)
+    ny=length(yg)
+    y_grid_res=yg[2]-yg[1]
+
+    # remove attractors from the basins:
+    bsn = deepcopy(basins)
+    ind = iseven.(bsn)
+    bsn[ind] .= bsn[ind] .+ 1
```
Datseris, line 34: You see, I was actually thinking about that and what `basins` returns. Given that `basins_general` returns both the `basins`, as well as the `attractors`, is it really useful to have some "attractor" cells in the `basins` Matrix? Perhaps it makes more sense to just "normalize" basins after the function has finished computation and just return a matrix that has integers `1, 2, 3,...` all of them counting attractors?

src/basins/uncertainty_exp.jl
awage: The error bars is the variance of the different slopes (something went wrong with the first graph)

src/basins/uncertainty_exp.jl
awage: It depends it you want it as a feature, but for most computations on basins it is worthless (uncertainty exponent, basin entropy etc...). 

If the user doesn't care about the internals of the algorithm (most of them won't), we can provide a matrix with basins numbered from 1 to N.

Datseris: I agree with that. Would you mind modifying `basins_2Dmap` and `basins_general` in this PR to update this change?

awage: I think it would be less confusing for the user now.

src/basins/compute_ba.jl
```diff
@@ -293,6 +282,16 @@ function procedure!(bsn_nfo::BasinInfo, n::Int, m::Int, u, Ncheck::Int;  max_che
 end
 
 
+function store_attractor!(bsn_nfo::BasinInfo, u)
+    # We divide by to order the attractors from 1 to Na
+    if haskey(bsn_nfo.attractors , bsn_nfo.current_color/2)
+        push!(bsn_nfo.attractors[bsn_nfo.current_color/2],  u) # store attractor
```
Datseris, line 288: Wouldn't this `/2` create problems? `4/2` does not give `2` but `2.0` and I think you cannot index with floats...

You need to do `4÷2` instead, (use \div<TAB> to get ÷ )

awage: Julia didn't protest! But it is safer the other way.

Datseris: Aaaaah it is because `attractors` is a dictionary. So there is an implicit automatic conversion to the keytype. I would still use div for source code clarity: we know the keys are of type Int.

Datseris:
@awage should the docstring of the basins function be modified with our changes? E.g. it says
```
1. The trajectory hits 60 times in a row an unnumbered cell: it is considered an attractor
   and is labelled with a even number.
```
Also, shouldn't this `60` number be a keyword?

Datseris:
Similarly:
```
The trajectory hits a known basin 10 times in a row: the initial condition belongs to
   that basin and is numbered accordingly.
```
shouldn't the 10 be a parameter?

awage:
Thanks for the guidelines. I will try to follow the style and avoid possible confusing notations.

src/basins/uncertainty_exp.jl
awage: > @awage should the docstring of the basins function be modified with our changes? E.g. it says
> 
> ```
> 1. The trajectory hits 60 times in a row an unnumbered cell: it is considered an attractor
>    and is labelled with a even number.
> ```
> 
> Also, shouldn't this `60` number be a keyword?

This number is a keyword in the function `procedure` but it is not propagated to the user functions. It is not very important since in the algorithm the attractors are double checked. First you have to hit 60 times in a row and then 60 times again. But I don't think it is worth mentioning.

awage:
> Similarly:
> 
> ```
> The trajectory hits a known basin 10 times in a row: the initial condition belongs to
>    that basin and is numbered accordingly.
> ```
> 
> shouldn't the 10 be a keyword argument?

It could be also introduced as a keyword. It works for maps only but for some reason it is not very reliable for Poincaré maps so it is not considered in this case. 

10 times should be OK in most cases. But you never know if you have a trajectory that sticks very close from another basin boundary during 10 iterations. In normal cases it should be pushed away from the boundary quite quickly.

Datseris:
So then if could you make these two parameters keywords and do any necessary changes to the docstrings, that's all that I see from my end! Then I'll merge!!! I'm writing a new doc page as we speak!

awage:
@Datseris : Before committing the last changes I found an interesting optimization look at these figures: 

```
@btime bsn=Basins.basins_map2D(xg, yg, integ)
  2.194 s (642800 allocations: 814.23 MiB)
Basin of attraction structure
 size :       (300, 300)
 Number of attractors found: 3
```

```
julia> @btime bsn=Basins.basins_map2D(xg, yg, integ)
  98.254 ms (586606 allocations: 20.74 MiB)
Basin of attraction structure
 size :       (300, 300)
 Number of attractors found: 3
```

I think it is worth including this change, what do you think? It should be 10 to 15 lines of code.

Datseris:
Uuuh, of course?! Why wouldn't we add a 20-fold performance increase.?

awage:
Performance on discrete maps are quite insane: 

```
function newton_map(dz,z, p, n)
    f(x) = x^p[1]-1
    df(x)= p[1]*x^(p[1]-1)
    z1 = z[1] + im*z[2]
    dz1 = f(z1)/df(z1)
    z1 = z1 - dz1
    dz[1]=real(z1)
    dz[2]=imag(z1)
    return
end

function newton_map_J(J,z0, p, n)
   return
end

ds = DiscreteDynamicalSystem(newton_map,[0.1, 0.2], [3] , newton_map_J)
integ  = integrator(ds)

xg=range(-1.,1.,length=1000)
yg=range(-1.,1.,length=1000)

@btime bsn,att=basins_map2D(xg, yg, integ)
  1.135 s (9256447 allocations: 346.29 MiB)
```

PR #179, opened 2021-05-27
matching attractors functionality

### LINKED ISSUE #176: Attractor "matching"
Recent PRs introduced some really great functionality regarding estimating basins of attraction: `basins_general`.

I've started applying these methods, but I quickly encountered something I feared: the attractor IDs are not "matched" when computing basins for different parameters. For example, I am interested in the basins of the magnetic pendulum when changing one of the magnet's strength. I get:

image

Exactly because the enumeration of the attractors is somewhat "random", I get different numbers for the "same" attractors. Now, I know that "same" isn't really valid here, as different attractors exist at different parameters, but you get my point.

Here is one proposal I have: create a function `match_attractors!(b1, a1, b2, a2)` that tries to match the basin and attractors at parameter "1" with those at "2". We could have different ways of how to do it. Ideas:

1. maximum overlap of basins = same attractors
2. State space distance between attractors `a1` and `a2`

I'll start a PR on this soon. @awage any other ideas?

Datseris:
@awage well, this was much, much harder than I thought. I really spent several hours thinking about all cases and testing things out. Now it works. It works for any scenario including those that there is a significant change in the number of attractors.

Feel free to give a review!

Closes #176

### DIFF
--- a/src/basins/compute_ba.jl
+++ b/src/basins/compute_ba.jl
@@ -28,7 +28,7 @@ end
 
 """
     basins_map2D(xg, yg, integ; kwargs...) → basins, attractors
-Compute an estimate of the basins of attraction of a "discrete two dimensional"
+Compute an estimate of the basins of attraction of a "discrete two dimensional system"
 of the plane onto itself according to the method of Nusse & Yorke[^Yorke1997].
 The dynamical system can be:
 * An actual 2D `DiscreteDynamicalSystem`.
@@ -73,19 +73,19 @@ returns a matrix coding the basins of attraction and a dictionary with all attra
 `attractors` has the following organization:
 * The keys of the dictionary correspond to the number of the attractor.
 * The value associated to a key is a [`Dataset`](@ref) with the *guessed* location of the
-  attractor on the plane.
+  attractor on the state space.
 
 The method starts by picking the first available initial condition on the plane not yet
 numbered. The dynamical system is then iterated until one of the following conditions
 happens:
 1. The trajectory hits a known attractor already numbered: the initial condition is
-   numbered with corresponding odd number.
+   numbered with the corresponding number.
 1. The trajectory diverges or hits an attractor outside the defined grid: the initial
    condition is set to -1
 1. The trajectory hits a known basin `mc_bas` times in a row: the initial condition belongs to
    that basin and is numbered accordingly.
 1. The trajectory hits `mc_unmb` times in a row an unnumbered cell: it is considered an attractor
-   and is labelled with a even number.
+   and is labelled with a new number.
 
 Regarding performace, this method is at worst as fast as tracking the attractors.
 In most cases there is a signicative improvement in speed.
@@ -124,22 +124,22 @@ Compute an estimate of the basins of attraction of a higher-dimensional dynamica
 on a projection of the system dynamics on a two-dimensional plane.
 
 Like [`basins_map2D`](@ref), `xg, yg` are ranges defining the grid of initial conditions
-on the plane. Refer to [`basins_map2D`](@ref) for more details.
+on the plane. Refer to [`basins_map2D`](@ref) for more details regarding the algorithm.
 
-This function can be used to make attractor basins of higher dimension. E.g. to make
-a 3D basin you can make many 2D basins and concatenate them. For example:
+This function can be used to make attractor basins of higher dimension via the `complete_state`
+keyword. E.g. to make 3D basins you can make many 2D basins and concatenate them.
+For example:
 ```julia
 zg = 0:0.01:1 # the range defining the z part of the grid
-bs = Matrix{Int8}[]
-for z ∈ zg
-    b = basins_general(xg, yg, ds; complete = [z, 0.0])
-    push!(bs, b)
-end
+bs = [
+    basins_general(xg, yg, ds; complete_state = [z, 0.0])
+    for z ∈ zg
+]
 basins_3D = cat(bs...; dims = 3)
 ```
 
 ## Keyword Arguments
-* `dt = 1`: Approxiamte time step of the integrator. It is recommended to use values ≥ 1.
+* `dt = 1`: Approximate time step of the integrator. It is recommended to use values ≥ 1.
 * `idxs = 1:2`: This vector selects the two variables of the system that will define the
   "plane" the dynamics will be projected into.
 * `complete_state = zeros(D-2)`: This argument allows setting the _remaining_ variables
@@ -150,7 +150,8 @@ basins_3D = cat(bs...; dims = 3)
 * `diffeq...`: Keyword arguments propagated to [`integrator`](@ref).
 """
 function basins_general(xg, yg, ds::DynamicalSystem;
-        dt=1, idxs = SVector(1, 2), mc_att = 10, mc_bas = 10, mc_unmb = 60, complete_state=zeros(dimension(ds)-2), diffeq...
+        dt=1, idxs = SVector(1, 2), mc_att = 10, mc_bas = 10, mc_unmb = 60,
+        complete_state=zeros(dimension(ds)-2), diffeq...
     )
     integ = integrator(ds; diffeq...)
     idxs = SVector(idxs...)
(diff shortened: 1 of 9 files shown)


awage:
Nice! I will have a look.

src/basins/compute_ba.jl
```diff
@@ -421,3 +418,92 @@ function reset_bsn_nfo!(bsn_nfo::BasinInfo)
...
+end
+
+function _match_from_overlaps(b₋, a₋, ids₋, b₊, a₊, ids₊)
+    # Compute normalized overlaps of each basin with each other basin
+    overlaps = zeros(length(ids₊), length(ids₋))
+    for (i, ι) in enumerate(ids₊)
+        Bi = findall(isequal(ι), b₊)
+        for (j, ξ) in enumerate(ids₋)
+            Bj = findall(isequal(ξ), b₋)
+            overlaps[i, j] = length(Bi ∩ Bj)/length(Bj)
```
awage, line 492: Why do you need to normalize? I try to figure out a case where it can be problematic

Datseris: I have to normalize because of the case where the Bj becomes really small. I've tested this with the magnetic pendulum and if I didn't normalize I was getting incorrect results.

awage: Ahh! I see, this is when a basin is drastically reduced in size. You have a small that is "contained" in a bigger one. 

If you look at the following video, you will see that colors switches from time to time. This feature can fix this problem on most cases. 

https://www.youtube.com/watch?v=JyhaHyCvgw4

awage: Nothing more to suggest. I will try to do some pretty gifs with this feature.

src/basins/compute_ba.jl
```diff
@@ -421,3 +418,92 @@ function reset_bsn_nfo!(bsn_nfo::BasinInfo)
     bsn_nfo.prev_step = 0
     bsn_nfo.step = 0
 end
+
+
+"""
+    match_attractors!(b₋, a₋, b₊, a₊, [, method = :overlap])
+Attempt to match the attractors in basins/attractors `b₊, a₊` with those at `b₋, a₋`.
+`b` is an array whose values encode the attractor ID, while `a` is a dictionary mapping
```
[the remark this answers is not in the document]
Datseris: Not true, it can be any dimensional array. While our basins outputs 2D one can make several 2D slices by changing the initial conditions over a 3rd dimension and concatenate all these slices. then they would have a 3D array.

I wanted to put this in the documentation of `basins_general` but I forgot. Will do so now.

(the algorithms I've written regarding tipping points / etc. are dimension agnostic, will work with 3D or 4D basins)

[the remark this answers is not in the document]
Datseris: @awage  I've just pushed the documentation. Can you have a look at the docstring of `basins_general`?

Datseris:
@awage I have a difficulty / question here. Why is it that the outputted guessed attractors have only the location on the plane? This makes it hard to generalize to 3D, as the information is missing. Specifically:
```julia
d, α, ω = 0.3, 0.2, 0.5
ds = Systems.magnetic_pendulum(; d, α, ω)
xg = yg = range(-3, 3, length = 100)
b₋, a₋ = basins_general(xg, yg, ds; dt=1., idxs=1:2)
a₋[1]
```

```
2-dimensional Dataset{Float64} with 3 points
 0.97807    2.22512e-5
 0.978041  -5.83129e-6
 0.978048  -4.00438e-6
```

why isn't this dataset 4D instead?

src/basins/compute_ba.jl
```diff
@@ -123,10 +124,19 @@ Compute an estimate of the basins of attraction of a higher-dimensional dynamica
...
 
 Like [`basins_map2D`](@ref), `xg, yg` are ranges defining the grid of initial conditions
-on the plane.
-
-Refer to [`basins_map2D`](@ref) for detailed information on the
-computation and the structure of `basin` and `attractors`.
+on the plane. Refer to [`basins_map2D`](@ref) for more details.
+
+This function can be used to make attractor basins of higher dimension. E.g. to make
+a 3D basin you can make many 2D basins and concatenate them. For example:
```
awage, line 130:
```suggestion
3D basins you can make many 2D basins and concatenate them. For example:
```

Not sure when to use singular.

awage:
The other coordinate is 0 in this case, the integrator only keeps the coordinate on the plane z=0. I never though about this but you can complete the vectors using `complete_state` in `general_basins`. 

> @awage I have a difficulty / question here. Why is it that the outputted guessed attractors, for `basins_general`, have only the location on the plane? This makes it hard to generalize to 3D, as the information is missing. Specifically:
> 
> ```julia
> d, α, ω = 0.3, 0.2, 0.5
> ds = Systems.magnetic_pendulum(; d, α, ω)
> xg = yg = range(-3, 3, length = 100)
> b₋, a₋ = basins_general(xg, yg, ds; dt=1., idxs=1:2)
> a₋[1]
> ```
> 
> ```
> 2-dimensional Dataset{Float64} with 3 points
>  0.97807    2.22512e-5
>  0.978041  -5.83129e-6
>  0.978048  -4.00438e-6
> ```
> 
> why isn't this dataset 4D instead?

awage:
You are right. I'm stuck in 2D.  The algorithm looks only at the 2D projection on the plane. If an orbit is recurrent 60 times in a row then it is tagged as an attractor. 

If there is bad luck you may have two attractors that are close on the plane but are far apart in 3D. It is a drawback of this method.

There is a simple fix in `get_color_point!`

Change 
```
   done = procedure!(bsn_nfo, n, new_u, mc_att, mc_bas, mc_unmb)
```
for 
```
   done = procedure!(bsn_nfo, n, integ.u, mc_att, mc_bas, mc_unmb)
```

Also the function `store_attractor!` should be changed to store the whole vector.

Datseris:
@awage due to your excellent contributions you have push rights to the ChaosTools.jl repo. Do you mind checking out my branch and making these improvements to store the full vector? I'm not yet completely comfortable with the algorithm.

> If there is bad luck you may have two attractors that are close on the plane but are far apart in 3D. It is a drawback of this method.

Okay, this is an important point that I'll mention in the docstring.

awage:
Many thanks! I will fix the issue soon.

Datseris:
This is good to go from my end. @awage a final question: `basins_map2D` *should* work with a 2D continuous dynamical system as well, right? Or am I missing something?

awage:
Yes it works. As an example the Duffing oscillator without forcing is an 2D ODE:  

```
ds=Systems.duffing(rand(2); ω=0.1, f=0, d=0.1, β=-1)
integ  = integrator(ds)
xg=yg=range(-2,2,length=200)
bas,att = basins_map2D(xg, yg, integ; mc_att=3)
´´´

In this case it is better to use `mc_att = 3` since it is is more precise on the boundary. ´basins_general´ also works fine:

```
bas,att = basins_general(xg, yg, ds)
```

![imagen](https://user-images.githubusercontent.com/3494384/120031781-8a1f4880-bff9-11eb-9c3f-0aa91e0d1328.png)

**Comment by Datseris** [comment]
I've noticed that the keyword `mc_att` is not explained in the algorithm description:
```
1. The trajectory hits a known attractor already numbered: the initial condition is
   numbered with the corresponding number.
1. The trajectory diverges or hits an attractor outside the defined grid: the initial
   condition is set to -1
1. The trajectory hits a known basin `mc_bas` times in a row: the initial condition belongs to
   that basin and is numbered accordingly.
1. The trajectory hits `mc_unmb` times in a row an unnumbered cell: it is considered an attractor
   and is labelled with a new number.
```
Is it correct that the keyword tunes the first part of the algorithm?

**Comment by Datseris** [comment]
I'm changing the name to `basins_2D`, as apparently having a "map" is not a requirement.

File: src/basins/compute_ba.jl [hunk 4]
```diff
@@ -72,24 +73,24 @@ returns a matrix coding the basins of attraction and a dictionary with all attra
 `attractors` has the following organization:
 * The keys of the dictionary correspond to the number of the attractor.
 * The value associated to a key is a [`Dataset`](@ref) with the *guessed* location of the
-  attractor on the plane.
+  attractor on the state space.
 
 The method starts by picking the first available initial condition on the plane not yet
 numbered. The dynamical system is then iterated until one of the following conditions
 happens:
 1. The trajectory hits a known attractor already numbered: the initial condition is
```
awage, line 81:
```suggestion
1. The trajectory hits a known attractor already numbered `mc_att` consecutive times: the initial condition is
```

awage:
> Is it correct that the keyword tunes the first part of the algorithm?

Yes it does, it should be 2 or 3. Maybe for some systems higher values are necessary but I didn't see any differences above for values above 3. 

For me it is good to go.

Issue #183, opened 2021-06-10
basins_2D is unreasonably slow


Datseris:
MWE:
```julia
using DynamicalSystems

# Henon global bifurcation
he = Systems.henon(;a = 1.4, b = 0.3)

Z = 300
xg = range(-1.5, 1.5; length = Z)
yg = range(-0.5, 0.5; length = Z)
@time basins, attractors = basins_2D(xg, yg, integrator(he); mc_bas = 3)
```

```
  4.711121 seconds (58.35 M allocations: 1.551 GiB, 5.56% gc time)
```

This large amount of allocations, 58 million (!!!) is definitely an error. I see no reason why we should be allocating so much memory all the time. My guess is that we are creating a lot of vectors in the algorithm that shouldn't exist.

cc @awage

awage:
I think I know what is going on. It should be easy to fix.

awage:
Some improvements:
```
@time basins, attractors = basins_2D(xg, yg, integrator(he))
0.770388 seconds (9.51 M allocations: 403.362 MiB, 8.06% gc time)
```

Datseris:
Can you add the fix to the open #184 ?

awage:
We can close #183, fixed in #184

Issue #182, opened 2021-05-31
Basins for 2D Poincare map finds incorrect attractors


Datseris:
MWE:
```julia
using DynamicalSystems
ds = Systems.rikitake(μ = 0.47, α = 1.0)
plane = (3, 0.0)
pmap = poincaremap(ds, (3, 0.), Tmax=1e6;
    idxs = 1:2, rootkw = (xrtol = 1e-12, atol = 1e-12), reltol=1e-9
)
xg = yg = range(-6.,6.,length=200)
basin, attractors = basins_2D(xg, yg, pmap)

attractors[1]
```
```
julia> attractors[1]
3-dimensional Dataset{Float64} with 9 points
 -2.13894   1.1333   0.190275
  2.30516  -1.41217  0.214518
  2.33192  -1.38681  0.129178
 -2.19281   1.06658  0.0302465
  2.35349  -1.36082  0.0550896
 -2.12238   1.14843  0.235889
  2.28525  -1.42716  0.274113
 -2.17312   1.09433  0.0907607
 -2.17888   1.08661  0.0732694
```

As you can see, the z-coordinate is way off. It should instead be exactly 0.

I think this happens because of the modification we did for `basins_general` where we record the full state. Howver in the poincare map we have to be a bit careful. Stepping the integrator and then getting `integ.u` is **incorrect**. The poincare map logic does an interpolation and then returns the interpolated state, but here we are missing this interpolation. We need to get the state the integrator returns upon calling `u = step!(pmap)`.

cc @awage as well.

awage:
I think I understand. This is why I had some weird results with `rikitake`. Maybe it is possible to modify `get_state` to return the interpolated state instead of the integrator state.

awage:
There is a possible fix: you can add the interpolated state to the struct `PoincareMap` and have it returned with `get_state`

This is not consistent with the definition of `get_state` that return the state of the integrator but it is a minimal change to have a correct projected state on the map for basins.

Datseris:
We can agree that at the moment PoincareMap shouldn't be supported. Will remove it from the documentation until we fix these issues.

awage:
I leave a possible solution in PR #190.

PR #184, opened 2021-06-10
Extension of basin general  to n-dim

awage:
I found a way to extend to n-dimension. It seems to work with the magnetic pendulum quite fine. I am amazed how good Julia handle n-dim arrays and CartesianIndex is a smashing. It was not much to change after all. 

I tried in 4D and it worked but computational load explodes quickly. 

```
ds = Systems.magnetic_pendulum(γ=1, d=0.3, α=0.2, ω=0.5, N=3)
integ = integrator(ds, u0=[0,0,0,0], reltol=1e-9)
xg=range(-2,2,length=100)
yg=range(-2,2,length=100)
zg=range(-1,1,length=10)
bsn , att = basins_general([xg, yg, zg], ds; dt=1., idxs=1:3, complete_state=[0])

```

I have still some tests to do and also I have to check the `complete_state` functionality. It may be broken.


Datseris:
We should test this PR with some system that has attractors spanning all three dimensions. For example: https://www.sciencedirect.com/science/article/pii/S0375960114002850?via%3Dihub 

Furthermore we should try to address #183 as much as possible in this PR. I will be doing some performance improvements over the weekend.

src/basins/compute_ba.jl
```diff
@@ -415,15 +415,11 @@ function get_box(u, bsn_nfo::BasinInfo)
 end
 
 function check_outside_the_screen!(bsn_nfo::BasinInfo, new_u, old_u, inlimbo)
-    if norm(new_u-old_u) < 1e-5
+    bb_max = maximum(abs.(bsn_nfo.bbox[3])); # this is the largest size of the phase space
+    if norm(new_u-old_u) < 1e-5 || inlimbo > 60*20 ||  norm(new_u) > 10*bb_max
```
Datseris, line 419: What's `60*20`?

awage: It was in the original code. It is supposed to be a long transient before we consider the orbit lost outside the grid. 

The last condition is also arbitrary. I considered that if the orbit leaves a square ten times bigger than the grid then it diverges. Maybe it is too much.

src/basins/compute_ba.jl
```diff
@@ -356,11 +360,11 @@ function draw_basin!(xg, yg, integ, iter_f!::Function, reinit_f!::Function, get_
              break
          end
 
-         x0 , y0  = xg[ind[1]] , yg[ind[2]]
          # Tentatively assign a color: odd is for basins, even for attractors.
          # First color is one
          bsn_nfo.basin[ind] = bsn_nfo.current_color + 1
-         u0 = SVector(x0, y0)
+         #u0 = SVector(x0, y0)
+         u0 = SVector([grid[k][ind[k]] for k in 1:NDU]...)
```
awage, line 367: This line is trouble. In profiling it is slow.

src/basins/compute_ba.jl
```diff
@@ -399,31 +403,23 @@ function get_color_point!(bsn_nfo::BasinInfo, integ, u0, mc_att, mc_bas, mc_unmb
...
-    xstep = (xg[2]-xg[1])
-    ystep = (yg[2]-yg[1])
-    xu, yu = u
-    n = m = 0
     # check boundary
-    if xu >= xg[1] && xu <= xg[end] && yu >= yg[1] && yu <= yg[end]
-        n = Int(round((xu-xg[1])/xstep)) + 1
-        m = Int(round((yu-yg[1])/ystep)) + 1 # +1 for 1 based indexing
+    bb_max = bsn_nfo.bbox[3]; bb_min = bsn_nfo.bbox[2]; steps = bsn_nfo.bbox[1];
+    if prod(bb_min .<= u .<= bb_max)
```
awage, line 408: This is also very slow, also the line 410 and 411.

Datseris: Wow, what's going on here? Do you want to check that all elements in `u` are between the `bb_min` and `bb_max`? If so, this is the most ineficient way to do it probably. You should do instead `if all(bb_min <= x <= bb_max for x in u)`

src/basins/compute_ba.jl
Datseris: Oh no this is big trouble. This is type unstable. You should be constructing `SVector` from vectors. What we have to do instead is to have a static inference on the dimensionality of the basin. The reason we don't have this now is because you made the decision that the input is `[xg, yg, zg]` instead of the _tuple_ `(xg, yg, zg)`. Tuples have known size at compile time, which we can propagate later.

I can fix this tomorrow when I will work on this.

Datseris:
@awage keep those performance annotations coming. Tomorrow I will work on the performance of this PR and will incorporate fixes for all comments.

src/basins/compute_ba.jl
```diff
@@ -1,9 +1,8 @@
 export draw_basin!, basins_2D, basins_general, match_attractors!
 
 mutable struct BasinInfo{F,N,T,Q}
-    basin::Matrix{Int16}
-    xg::F
-    yg::F
+    basin::AbstractArray
```
Datseris, line 4: Really not a good idea to leave the abstract type here. Operations on `basin` are done all the time, it's type should be inferrable.

src/basins/compute_ba.jl
```diff
@@ -93,7 +92,7 @@ In most cases there is a signicative improvement in speed.
 function basins_2D(xg, yg, pmap::PoincareMap; mc_att = 3, mc_bas = 10, mc_unmb = 60)
     reinit_f! = (pmap,y) -> _init_map(pmap, y, pmap.i)
     get_u = (pmap) -> pmap.integ.u[pmap.i]
-    bsn_nfo = draw_basin!(xg, yg, pmap, step!, reinit_f!, get_u, mc_att, mc_bas, mc_unmb)
+    bsn_nfo = draw_basin!([xg, yg], pmap, step!, reinit_f!, get_u, mc_att, mc_bas, mc_unmb)
```
Datseris, line 95: This should be a tuple `(xg, yg)`, not a vector. Vectors do not have known size at compile time, and this will be important later on in the algorithm execution.

src/basins/compute_ba.jl
```diff
@@ -324,17 +316,29 @@ Compute an estimate of the basin of attraction on a two-dimensional plane. This
...
 examples for a Poincaré map of a continuous system.
 * `reinit_f!` : function that sets the initial condition to test on a two dimensional projection of the phase space.
 """
-function draw_basin!(xg, yg, integ, iter_f!::Function, reinit_f!::Function, get_u::Function, mc_att, mc_bas, mc_unmb)
+function draw_basin!(grid, integ, iter_f!::Function, reinit_f!::Function, get_u::Function, mc_att, mc_bas, mc_unmb)
     NDS = length(get_state(integ))
     complete = false
-    bsn_nfo = BasinInfo(ones(Int16, length(xg), length(yg)), xg, yg,
-                iter_f!, reinit_f!, get_u, 2,4,0,0,0,1,1,0,0,Dict{Int16,Dataset{NDS,eltype(get_state(integ))}}(), Vector{CartesianIndex}())
+    nstep=map(x->x[2]-x[1],grid)
```
Datseris, line 328: if `grid` is guatanteed to contain ranges, this is simpler written as `nstep = step.(grid)`

(p.s.: please don't forget the spaces between `=`, they make the code much more readable)

src/basins/compute_ba.jl
Datseris: All of these should be keyword arguments. Not only it allows users to tune, but it makes the code much, much more readable by replacing literal constant with a intention-revealing name.

Datseris:
@awage I'm working on this right now. One thing that we should definitely improve before merging is the code readability. One thing we should definitely improve before merging is the `procedure!` function:

1. It is very large, 100 lines of code. It should be split up into more, smaller, atomic functions, each being up to 20 lines of code.
2. It's name does not convey it's purpose, so I will rename it to something that includes the "basins"

Datseris:
While working on the source code I have found _extremely confusing_ the following: what the states `u` represent. Typically `u` as a symbol is used to represent the full system state. In the source, it is not at all for me clear when `u` means the full state, or the projection of the state on the sub-space denoted by `grid`. @awage do you think you can go through and clarify this? I don't think I can do it. 

Let's do the following:

* `u` is used exclusively to denote the full system state, as is done in the rest of the library.
* `y` can be used to denote the _projected_ state, which is of smaller dimensionality, and same dimensionality as the `grid`.

awage:
Ok, I will work on these issues next week. Since the algorithm is a finite state machine maybe there is way to define states like `:search_attractor` `:atractor_found` and so on. I will try to find a way.

Datseris:
I've addressed most of the concerns, did a lot of performance improvements, and increased code clarity a lot. What remains is:

* Use symbol `y` for projected state and `u` for full state
* Replace all literal constants like `60` or `20*60` with named keyword arguments.
* Address the `TODO` markings I've left in the source code.

Have a look and let me know what can be improved. Even though I've tried to make the source be as identical as originally regarding output, this was really hard to control, as I've done some really large changes. As a result, now the e.g. Henon map test fails. It says that `count(basin .== 1)` is 4261 instead of 4700 and that `count(basin .== -1)` is 5739 instead of 5730. I couldn't pinpoint exactly which change made this disparity.

awage:
> We should test this PR with some system that has attractors spanning all three dimensions. For example: https://www.sciencedirect.com/science/article/pii/S0375960114002850?via%3Dihub

This system is a pain in the neck with its non-attracting torus, the algorithm only detect pieces of the torus and the attractor. There is more gentle system with attractors in 3D: 

http://inaesp.org/PublicJG/freire_et_al_CHAOS2008.pdf   (https://doi.org/10.1063/1.2953589)

`basins_general` works just fine:

test

Datseris:
Awesome, you can put this system in DynamicalSystemsBase.jl if you'd like!

awage:
I have complety changed the algorithm. I think it is much less cryptic right now. It will be easier to maintain and to extend. In the meantime I solved some problems in the original algorithm since they were cutting corners. 

The idea now  is that you are in particular state and depending on the cell you hit you have a transition. Then depending on an internal counter of the state you may have actions or transition (I left a schematic drawing). There are still some minor issues to solve. 

It took some serious brain time to put it that way! 


thumbnail_IMG_6453

Datseris:
Beautiful. Just beautiful. I'll do another round of review when I have the time. What I'm thinking is that now, with the approach we have, how much of it is really based on the work of Nussa/Yorke? Probably better to make a nice figure of the state machine and use this as docs instead of saying "we implement the method of Nussa/Yorke"?

awage:
The basic idea is the same: 
* you have a grid and you keep track of the events in an array
* you make sure to keep the transient times to the attractors as short as possible.

The internals of the algorithm are now very different. I will document the algorithm as clearly as possible later this week. 

There is something I need to solve. The :bas_hit state is only valid for 2D maps. It assumes that when you hit the same basin several times in a row you will end to the attractor of that basin. This is not always reliable in dimension above 2. But for 2D it is really quick so it is worth keeping it.

awage:
@Datseris : I have a question about the functionality of `basins_general`. You left some comments that suggest that the grid should have the dimension of the phase space of the dynamical system. It is not necessary, you can always use a grid in a lower dimension projected space. For example: 

```
ds = Systems.magnetic_pendulum(γ=1, d=0.3, α=0.2, ω=0.5, N=3)
integ = integrator(ds, u0=[0,0,0,0], reltol=1e-9)
xg=range(-2,2,length=100)
yg=range(-2,2,length=100)
zg=range(-1,1,length=10)
bsn,att = basins_general((xg, yg, zg), ds; dt=1., idxs=1:3, complete_state=[0])
```
This should work. Maybe you want to cleave the functionalities: on the one hand use `basins_2D` for projection and `basins_general` for n-dim cases. This is a design choice, since we are not limited by the algorithm.

Datseris:
> You left some comments that suggest that the grid should have the dimension of the phase space of the dynamical system

Oh sorry, this was never my intention. Maybe I misspoke in the comment. Sure, the grid can have smaller dimensionality.

My idea is that `basins_general` becomes _the_ basins function that has all important documentation, while `basins_2D` will become just a convenience that also allows for poincare maps and stroboscopic maps.

I also suggest that we rename `basins_general` to `basins_of_attraction` or `attractor_basins`. Just `basins` would be nice, but I'd prefer to leave this name unbinded so that users can use it for the actual basins matrix.

awage:
I think we can get everything working with `basins_general`. It seems that maps are working and Poincaré maps also with dispatch on `basins_general`. The interface would be a unique function that in most case doesn't need much keywords arguments.

The only systems left are stroboscopic maps. But it can be solved if we allow to step exactly at T. 

What is your catch?

Datseris:
You are right. Everything can be done with a single function. For stroboscopic maps, a user would provide `diffeq = (alg = Tsit5(), dt = T, adaptive = false)` to ensure the stepping is one period precicely.

src/basins/basins_highlevel.jl
```diff
@@ -1,21 +1,25 @@
...
-* An actual 2D `DiscreteDynamicalSystem` or `ContinuousDynamicalSystem`.
-* 2D poincaré map of a 3D `ContinuousDynamicalSystem`.
-* A 2D stroboscopic map, i.e. a periodically forced 2D `ContinuousDynamicalSystem`.
 
-For a higher-dimensional dynamical systems, use [`basins_general`](@ref).
+"""
+    basins_of_attraction(grid::Tuple, ds::DynamicalSystem; kwargs...) -> basins, attractors
+Compute an estimate of the basins of attraction of a dynamical system `ds` on
+a partitioning of the state space given by `grid`. The method has been
+inspired by the 2D grid approach devellopped by Nusse & Yorke [^Yorke1997].
```
Datseris, line 8: devellopped -> developed

Datseris:
todo for me: we can make `complete_state` a functor to avoid re-creating states all the time.

awage:
From my end it is OK and closes #183 and #182 . But I detected a problem with the tag propagation of diffeq. I tried to add a callback on an equation but the program doesn't respond. I don't know if it should be addressed in this PR:

```
using DifferentialEquations
 using DynamicalSystems
function forced_pendulum(u, p, t)
    @inbounds begin
    d = p[1]; F = p[2]; omega = p[3]
    du1 = u[2]
    du2 = -d*u[2] - sin(u[1])+ F*cos(omega*t)
    return SVector{2}(du1, du2)
    end
end

# We have to define a callback to wrap the phase in [-π,π]
function affect!(integrator)
    if integrator.u[1] < 0
        integrator.u[1] += 2*π
    else
        integrator.u[1] -= 2*π
    end
end
condition(u,t,integrator) = (integrator.u[1] < -π  || integrator.u[1] > π)
cb = DiscreteCallback(condition,affect!)

F = 1.66 ; ω = 1.; d=0.2;
ds=ContinuousDynamicalSystem(forced_pendulum, rand(2),[d, F, ω])
xg = range(-pi,pi,length=100); yg = range(-2.,4.,length=100)
bsn,att = basins_of_attraction((xg, yg), ds; T=2*pi/ω, diffeq = ( save_everystep=false, reltol=1e-9, callback =cb))
```

Datseris:
Thanks a lot for this great work. I'll have a look over the weekend, fix as much as I can, and merge and tag. Remaining issues (e.g. if I can't fix the callback) will be opened as new issues.

Datseris:
@awage I am doing a performance improvement now. One thing I've noticed is that during the re-initialization process we were constantly creating new vectors. So I created a helper type that stores two dummy vectors and does in-place creation of a new state. What I'm having difficulty is extending this to the Poincare Map. Why is it that we're doing completely different handling for it?

Datseris:
ok I got what's going on, I'll push a fix soon

awage:
> @awage I am doing a performance improvement now. One thing I've noticed is that during the re-initialization process we were constantly creating new vectors. So I created a helper type that stores two dummy vectors and does in-place creation of a new state. What I'm having difficulty is extending this to the Poincare Map. Why is it that we're doing completely different handling for it?

Pmap has its own integrator, it is a bit special.

Datseris:
Alright, I'm finally done. I was a bit discontent with how much more complexity the Poincare map was adding to the code. I decided to eliminate the complexity, in favor of treating the "state" of the Poincare map as the full system state. I will add a bit more info about this at the docs. I have to open a PR at DynamicalSystems.jl for the new documentation.

Datseris:
Hm, something may be off. I'm finding 4 attractors for the duffing oscillator:
image

four colors from 
```
using DynamicalSystems, PyPlot
ω=1.0; f = 0.2
ds = Systems.duffing([0.1, 0.25]; ω, f, d = 0.15, β = -1)
basins, attractors = basins_of_attraction((xg, yg), ds; T=2π/ω)
pcolormesh(xg, yg, basins')
```

Datseris:
I've also noticed that the `uncertainty_exponent` is still limited to a 2D grid. Is there any fundamental limitation that keeps it there, or we can also improve this algorithm to be arbitrary dimensional?

Datseris:
I have the documentation PR ready. https://github.com/JuliaDynamics/DynamicalSystems.jl/pull/143  

So after we resolve the the multiple basins for duffing I merge and tag!

awage:
> Hm, something may be off. I'm finding 4 attractors for the duffing oscillator:

Something awkward happens with `SimpleATsit5()` again. It does not seems related to `DynamicalSystems`, MWE: 

```
using  DynamicalSystems, Plots, SimpleDiffEq, OrdinaryDiffEq
ds = Systems.duffing([0.1, 0.25]; ω = 1., f = 0.2, d = 0.15, β = -1)
T = 2π/1.
integ = integrator(ds, alg=SimpleATsit5())
integ2 = integrator(ds, alg=Tsit5())
integ.u = integ2.u = [-0.2, 0.1]
v=Vector{SVector{2,Float64}}()
v2=Vector{SVector{2,Float64}}()
for k = 1:700
    step!(integ, T, true)
    step!(integ2, T, true)
    push!(v,integ.u)
    push!(v2,integ2.u)
end
v = Dataset(v); v2 = Dataset(v2)
plot(v[:,2],seriestype=:scatter)
plot!(v2[:,2],seriestype=:scatter)
```
 
If you look at the picture the problem is not that they go to different attractor (it could happen since it is chaotic and we have different solvers). The problem is the very long transient before getting to the steady state. It is maybe the stepping problem that 
pops up again. 

imagen

Datseris:
I'll simply modify the documentation PR to instruct to users to use a higher accuracy solver than the default one.

Datseris:
This is good to go and merge from my end, if you don't have any final remarks.

awage:
> I've also noticed that the `uncertainty_exponent` is still limited to a 2D grid. Is there any fundamental limitation that keeps it there, or we can also improve this algorithm to be arbitrary dimensional?

Well, this is something to dig. In the original algorithm they take "balls" in one dimension. The reason is that you need "balls" of radius `ε` to measure correctly the uncertainty exponent `α`. When the resolution is different along axes you may have arrays of different size along the dimensions. It causes trouble.  

On the other hand, I did not find this solution satisfactory since you may have a bias due to the direction of the axis relative to the boundary. I will try to find where they say that it is independent of the chosen ball.

awage:
Thank you very much @Datseris!  for me everything fine.

PR #186, opened 2021-06-21
Bugfix `basins_of_attraction`

Datseris:


Datseris:
@awage have we actually tested this function with three-dimensional basins? How did you produce the lorenz84 animation?

At the moment I'm running
```julia

xg = yg = range(-1.0, 2.0; length=300)
zg = range(-1.5, 1.5; length=30)
bsn, att = basins_of_attraction((xg, yg, zg), ds; mx_chk_att=4)
```
and it takes a _lot_ of time. I'm hoping it will terminate some time and there isn't any error :D How long did it take for you? 

In any case, we should come up with a rudimentary test that will not take a lot of time to run and we can run it on CI that tests higher dimensional functionality.

Datseris:
Hm, I've reduced to `xg = yg = range(-1.0, 2.0; length=30)`. But it still takes way too long. If we think about scaling, a 30x30x30 grid and a 300x300 grid shouldn't take that much longer... 

Perhaps there is something wrong?

Datseris:
(@awage remember you can push into this branch directly because you are an org member now. You just need to set as upstream the `origin`)

Datseris:
Okay, we guaranteed have a problem. Not even this terminates:
```julia
using DynamicalSystems
ds = Systems.lorenz84()
xg = yg = range(-1.0, 2.0; length=5)
zg = range(-1.5, 1.5; length=5)
@time bsn, att = basins_of_attraction((xg, yg, zg), ds; mx_chk_att=4)
unique(bsn)
```

So I guess we need a new keyword flag regarding max integration time. If the trajectory has been integrated longer than `Tfail` and none of the identification conditions have been satisfied, we need to terminate iteration and throw a warning message.

Datseris:
With my addition of `Tfail` this terminates:
```julia
using DynamicalSystems
ds = Systems.lorenz84()
xg = yg = range(-1.0, 2.0; length=30)
zg = range(-1.5, 1.5; length=30)
@time bsn, att = basins_of_attraction((xg, yg, zg), ds; mx_chk_att=4)
unique(bsn)
```
and gives 
```
julia> unique(bsn)
5-element Vector{Int16}:
      1
 -16384
      2
      3
      0
```
which is bad. I'm not sure why `0` is there, no cell should have `0` value, right...?

Datseris:
Oh wow, @awage you won't believe it. I've noticed a _massive_ performance problem that I've just fixed in the last commit. Before/After:
```julia

 32.068627 seconds (193.74 M allocations: 3.825 GiB, 1.75% gc time)

 18.460672 seconds (6.30 M allocations: 158.412 MiB, 0.17% gc time, 0.34% compilation time)
```
The problem is that in `basin_cell_index` when we are doing `ind = @. round(Int, (y_grid_state - bsn_nfo.grid_minima)/bsn_nfo.grid_steps) + 1` the output of this is a `Vector`. This cannot have statically deducable dimension so dynamic dispatch occurs when we do `CartesianIndex(ind...)`. I've fixed that.

Datseris:
Indeed, my latest commit brings it down to:
```julia
 18.152256 seconds (2.07 M allocations: 33.561 MiB)
```
While it doesn't save a lot of time, it does save a pretty large amount of allocations, from 6 million to 2 million.

---

This reveals _yet another problem_. Where are all these allocations coming from? They are definitely not from stepping the integrator:
```
julia> integ = integrator(ds)
t: 0.0
u: 3-element SVector{3, Float64} with indices SOneTo(3):
 0.1
 0.1
 0.1

julia> using BenchmarkTools;

julia> @btime step!($integ, 1.0)
  3.222 μs (0 allocations: 0 bytes)
```
So, why are there so many allocations? I would imagine that after we have created the `BasinInfo` struct, all subsequent operations shouldn't do a lot of allocations, as it is now a manner of stepping the integrator and checking the state machine...

awage:
> which is bad. I'm not sure why `0` is there, no cell should have `0` value, right...? (I've made it so that cells that exceed `Tfail` are given the value `-16384` for now)

It seems that it is a classic int16 overflow. Thats's weird. I will try to understand what is going on.

awage:
> This revealed another performance problem. The `ind` calculated above _should_ be a static vector. It isn't probably because `y_grid_state` is not a static vector. However, that is a problem on our side. It should be a static vector because accessing a static vector with another static vector returns a static vector. I'll try to fix that.

Wow! That is well above my understanding of how Julia works. This is fine tuning.

awage:
> @awage have we actually tested this function with three-dimensional basins? How did you produce the lorenz84 animation?

I don't remember but you can have a coffee and a shower before having the results. It is a lot of points to compute. 

I have tried with the magnetic pendulum in 4D and also gives correct results.

awage:
Something is broken, I have to check where it happens. On another branch I get: 
´´´
xg=yg=range(-1.,2.,length=200)
zg=range(-1.5,1.5,length=30)
@time bsn,att = basins_of_attraction((xg, yg, zg), ds; Δt=1.,  idxs=1:3, complete_state=[])
231.798212 seconds (102.91 M allocations: 2.970 GiB, 0.21% gc time, 1.01% compilation time)
```
Its pretty fast and the basins match up.

**Comment by Datseris** [comment]
> On another branch I get

What's this other branch?

**Comment by awage** [comment]
I checked out ndim, head is at 50325bd7

**Comment by awage** [comment]
> ds = Systems.lorenz84()
> xg = yg = range(-1.0, 2.0; length=30)
> zg = range(-1.5, 1.5; length=30)
> @time bsn, att = basins_of_attraction((xg, yg, zg), ds; mx_chk_att=4)

I have: 
```
unique(bsn)
3-element Vector{Int16}:
 1
 2
 3
```

It is missing an attractor but it is not surprinsing since it is a coarse grid.

Datseris:
Okay that's good. So clearly I messed up somewhere with all the changes I did. Damn, well let's try to find out. Now, the awesome part is that if it is already so fast in your branch, with the optimizations I did here it should skyrocket (my optimizations in this PR shouldn't really affect the algorithm. The mistake happened somewhere in the merged #184 already)

awage:
#184 seems to be all right up to the last commit 611e88734e

Datseris:
I'm confused now. I opened this PR exactly because just using the merged #184 had problems for me and the computation was taking forever :S Are you telling me that with #184 , which is also the current stable release of ChaosTools, you have no problems, provided that you give the argument `complete_state = []` ?

awage:
I will checkout master just to see. I also removed `mx_chk_att=4` which is needed only to be extra sure that you have the right attractor. Default value is allright for most cases.

Datseris:
I just checked `master`. I confirmed what you said. What the hell happened in between then... :S :S :S

Datseris:
Okay, tomorrow I will start this Pull Request from scratch, and do a benchmark test at every step.

Datseris:
Nah, actually I'll do it tonight after dinner.

awage:
No rush at all! Tomorrow I can help you with this.

PR #187, opened 2021-06-21
Increase performance of basins, and fix minor bugs

Datseris:
Superseedes #186


Datseris:
Alright, holy crap this was rewarding. Test code snippet:
```julia
using DynamicalSystems
ds = Systems.lorenz84()
xg=yg=range(-1.,2.,length=100)
zg=range(-1.5,1.5,length=30)
# `complete_state` keyword is necessary for `master` to work.
@time bsn,att = basins_of_attraction((xg, yg, zg), ds; complete_state = [])
@show unique(bsn)
```

This PR:

```
 24.209292 seconds (1.20 M allocations: 40.092 MiB)
unique(bsn) = Int16[1, 2, 3, 4]
```

Current master:
```
 45.606271 seconds (302.77 M allocations: 6.119 GiB, 1.68% gc time)
unique(bsn) = Int16[1, 2, 3, 4]
```

Now _that's_ what I'm talking about. Also, fun fact: turns out that merging all the `if` clauses in `_identify_basin_of_cell!` into a single `if-elseif-...-else` clause makes the function **twice as slow**. Took me a while to identify. Not sure why, probably some crazy compiler optimizations happening under the hood.

Datseris:
Will merge and tag new release once tests pass.

Datseris:
One thing we haven't done is test this with in-place systems.

Datseris:
@awage something for you to know for the future: Performance of `basins_of_attraction` drops by a factor of 2.2 (from 20 seconds to 45) if you give the keyword `complete_state = []`. This is because `[]` is actually `Vector{Any}`. Given that we use `complete_state` literally all the time, this type instability brings massive performance drop.

Datseris:
Good thing I've tested out of place systems. Because it doesn't work for them. Example:
```julia
    ds = Systems.henon_iip(zeros(2); a = 1.4, b = 0.3)
    xg = yg = range(-2.,2.,length=100)
    basin, attractors = basins_of_attraction((xg,yg), ds)
    pcolormesh(xg,yg, basin')
```
which gives:

image

using `ds = Systems.henon()` instead, which is an out-of-place, static vector based system, gives correct result:

image

I've checked and the problem is _not_ the line:
```julia
    get_projected_state = (integ) -> view(get_state(integ), idxs)
```
I thought that because it returns a view, we might be doing wrong over write.

What I suspect happens is that we modify somewhere in the code, in place, the full system state / integrator state. @awage can you think of where we might be doing something like that? Or in general where in-place systems might be failing?

awage:
I am debugging this and it seems that the iip doing something strange with the variables. I modified the `henon_iip` code to trace the orbit:  

```
function hiip(dx, x, p, n)
    @show x,dx
    dx[1] = 1.0 - p[1]*x[1]^2 + x[2]
    @show x,dx
    dx[2] = p[2]*x[1]
    @show x,dx
    return
end
```

For one specific initial condition I have: 
```
(integ.u, y0) = ([-2.0, -2.0], [-2.0, -2.0])
(x, dx) = ([-2.0, -2.0], [-2.0, -2.0])
(x, dx) = ([-6.6, -2.0], [-6.6, -2.0])
(x, dx) = ([-6.6, -1.9799999999999998], [-6.6, -1.9799999999999998])
```

If you look carefully you have `x = dx` at each step. This is wrong. I will try to see where it comes from.

awage:
Maybe it is here:

```
function step!(integ::MDI{true})
    # vector swap
    integ.dummy, integ.u = integ.u, integ.dummy
    integ.f(integ.u, integ.dummy, integ.p, integ.t)
    integ.t += 1
    return
end
```

It looks like there is a static binding created.

awage:
This do the fix, but maybe there is a better way in `discrete.jl` 

```
function reinit!(integ::MDI, u = integ.u, Q0 = nothing; t0 = integ.t0)
    integ.u = copy(u)
    integ.dummy = copy(u)
    integ.t = t0
    if Q0 != nothing
        set_deviations!(integ, Q0)
    end
    return
end
```

awage:
@Datseris: I was about to open a new branch in DynamicalSystemBase. But I am not comfortable messing with the repository. Do you prefer a PR from my github account?

Datseris:
@awage : from my end, there is nothing to fear! As long as you do not commit to the master branch, but instead open a new branch and start a PR from there, all is good. But of course, if you are not comfortable with it, please do go ahead and work from your own account!

thanks a lot for spotting this out! Will also have time to work on it again tonight!

awage:
Just for fun (takes only 4min to compute a 200x200x100 grid, it's crazy!):
 
test

awage:
More psychedelic stuff (this one has also atractors in ·3D):

```
xg = yg = zg = range(-6.,6.,length=200)
@time bsn, att = basins_of_attraction((xg, yg, zg), ds)
anim = @animate for k = 1:200 ; plot(xg, yg, bsn[:,:,k]', seriestype=:heatmap, colorbar=false); end;
gif(anim, "test3.gif", fps=10)
```


test3

Datseris:
Awesome! Super happy! I can now starting working on merging the PRs and getting everything together.

Datseris:
Ah, unfortunately, we still have some work on our hands. The in-place continuous systems do not seem to work:
```julia
    ds = Systems.lorenz_iip()
    xg = yg = range(-5.,5.; length=5)
    zg = range(0,10; length=5)
    @time basin, attractors = basins_of_attraction((xg,yg,zg), ds)
```
I've been running this for 5 minutes now and it still hasn't terminated, even though I got a coarse grid. 

BUT, that being said, I will still merge this and open a new issue to track this. I want to get this really good functionality tagged out.

awage:
A progress bar? You made my day!!!

Datseris:
Hope you like cool animations


https://user-images.githubusercontent.com/19669089/123008029-7679be80-d3ba-11eb-988f-77604285182e.mp4

This functionality is just superb... wow.

awage:
It is interesting because it is a case that wouldn't work with a 2D projection. The two fixed points are symmetrical and collide on the z=0 plane. I can watch this for hours. 

I don't think there are any software around that can do this.

Datseris:
Yeap. Well by now 90% of DynamicalSystems.jl does not exist anywhere else.

Datseris:
I want to redo these computationsi n the chaotic regime and see how well the algorithm holds.

PR #196, opened 2021-07-06
Uncertain exponent for n-dim basins

awage:
This is a drastic turn in the computation of the uncertainty dimension. Following a theorem in [1], there is a relation between box-counting dimension (capacity dimension) and uncertainty exponent. I programmed a naive version of the box-counting dimension that performs in n-dim (tested up to 3D) and then returns the uncertainty exponent. 

There is the common problems of the estimation of the box-counting dim. However users should be more familiar with this measure and might get around the common pitfalls. 

There is the possibility to use `generalized_dim` but first the boundary has to be transformed into a `dataset`, i.e., we have to find the points in the boundary (which is not obvious at all). 

[1] McDonald, S. W., Grebogi, C., Ott, E., & Yorke, J. A. (1985). Fractal basin boundaries. Physica D: Nonlinear Phenomena, 17(2), 125-153.

EDIT: Now I find totally awkward that the function returns the box size `ε` the number of balls covering the boundary `N_ε` and then convert it to the uncertainty exponent `α`. This function should be called `basin_boundary_dimension` or something like that. The uncertainty exponent should be mentioned but it is a sort of side result.


Datseris:
> This is a drastic turn in the computation of the uncertainty dimension

Ha, not sure if this means a drastic _decrease_ in computation time, or _increase_, as the latter would be rather bad...

I'm going on vacation in two days for a week, so if I don't have a proper review until then do not worry.

awage:
Now it is much faster, 20 fold or so. However I have to think a little bit about the alternatives. It is tricky to estimate correctly the box-counting dimension. 

No deadlines for this PR!

awage:
I found a possible way to use `generalized_dim` with basins. With a simple filter you can detect the boundary. The assumption is that a point is the boundary if it has at least one neighbor in a different basin. This assumption is also used for the naive box-counting algorithm. Once we have a dataset with the points of the boundary we let `generalized_dim` do its magic: 

```
function get_boundary_filt(basin)
    # Kernel
    w = centered([1 1 1; 1 -8 1 ; 1 1 1])
    #w = centered([0 1 0; 1 -4 1 ; 0 1 0])

    # replicate for boundary conditions
    res = imfilter(basin,w,"replicate")
    res = abs.(res) .> 0.
    return res
end

function box_counting_dim(xg, yg, basin; kwargs...)
    # get points coordinates in the boudary
    bnd = get_boundary_filt(basin)
    I1 = findall(bnd .== 1);
    v = Dataset{2,eltype(xg)}()
    for  ind in I1; push!(v, [xg[ind[1]] , yg[ind[2]]]); end;
    return generalized_dim(v; q=0, kwargs...)
end
```

There is two issues: it introduces a dependency on `ImageFiltering.jl` and I don't know if it works in more than 2D.

awage:
I have added the basin entropy and the fractality test in this commit. The test works fine. 

There is an issue with `Neighborhood.jl` with the KDTree search I have this timing: 

```
@btime test_res, Sbb = basins_fractal_test(basin; ε=5)
  4.020 s (116223 allocations: 20.62 MiB)
```
With the BallTree structure from `NearestNeighbors.jl`I obtain: 
```
@btime test_res, Sbb = Basins.fractal_test(basin; ε=5)
  50.041 ms (105158 allocations: 17.60 MiB)
(:fractal, 0.5906221663828939)
```

Maybe I can open a PR in  `NearestNeighbors.jl`

Datseris:
It may be a problem with Neighborhood.jl, so I would suggest to open an issue there first. I just got back from vacation and will work on this during the weekend as well!

src/basins/basins_utilities.jl
```diff
@@ -86,4 +86,150 @@ function _match_from_distance(b₋, a₋, ids₋, b₊, a₊, ids₊)
...
+
+## Keyword arguments
+* `ε`: define the size of the box that will be used to sample the basin
+"""
+function basin_entropy(basins; ε = 20)
+    n_dim = size(basins)
+    vals = unique(basins)
+    S=Int16(length(vals))
+    pn=zeros(Float64,1,S)
+    Sb=0; Nb=0; N=0
```
Datseris, line 124: We need a consistent style. spaces should be either always or never around `=`. As the rest of the library uses them always, then every `=` needs space arounds it. 

(consistent style isn't nit-picking, it really increases code readability a lot)

awage:
Sorry for this, it is some old code recycling and I didn’t checked the
style and indentation. I will address this.

src/basins/uncertainty_exp.jl
```diff
@@ -1,99 +1,99 @@
...
-Estimate the uncertainty exponent[^Grebogi1983] of the basins of attraction. This exponent
-is related to the final state sensitivity of the trajectories in the phase space.
-An exponent close to `1` means basins with smooth boundaries whereas an exponent close
-to `0` represent complety fractalized basins, also called riddled basins.
-
-`xg`, `yg` are 1-dim ranges that defines the grid of the initial conditions.
-`basins` is the matrix containing the information of the basin.
-This functionality is currently implemented only for 2D basins.
+    basins_fractal_dimension(basins; kwargs...) -> ε, N_ε ,d
+Estimate the Haussdorff dimension `d` of the basins of attraction using the box-counting
```
Datseris, line 5: instead of Hasdorf dimension, I think it is probably better if we say "fractal dimension". Furthermore, if you typeset it as `[Fractal Dimension](@ref)`, it will link to the documentation page for the fractal dimension https://juliadynamics.github.io/DynamicalSystems.jl/dev/chaos/fractaldim/

src/basins/uncertainty_exp.jl
```diff
@@ -1,99 +1,99 @@
...
-    uncertainty_exponent(xg, yg, basins::Matrix; kwargs...) -> ε, f_ε ,α
-Estimate the uncertainty exponent[^Grebogi1983] of the basins of attraction. This exponent
-is related to the final state sensitivity of the trajectories in the phase space.
-An exponent close to `1` means basins with smooth boundaries whereas an exponent close
-to `0` represent complety fractalized basins, also called riddled basins.
-
-`xg`, `yg` are 1-dim ranges that defines the grid of the initial conditions.
-`basins` is the matrix containing the information of the basin.
-This functionality is currently implemented only for 2D basins.
+    basins_fractal_dimension(basins; kwargs...) -> ε, N_ε ,d
```
Datseris, line 4: To be clear: here we are not really calculating the fractal dimension of the basins, but of their boundaries instead, right? If yes, we should be clear about this in the docstring.

src/basins/uncertainty_exp.jl
```diff
@@ -1,99 +1,99 @@
...
+Notice that the uncertainty exponent and the box counting dimension of the boundary are
+related. We have `Δ₀ = 2 - α` where `Δ₀` is the box counting dimension,
+see [Fractal Dimension](@ref). The algorithm first estimates the box counting dimension of the
+boundary and returns the uncertainty exponent.
+
+[^Grebogi1983]: C. Grebogi, S. W. McDonald, E. Ott and J. A. Yorke, Final state sensitivity: An obstruction to predictability, Physics Letters A, 99, 9, 1983
+"""
+function uncertainty_exponent(basins::AbstractArray; range_ε = 2:maximum(size(basins))÷20)
+    ε, N_ε, d = basins_fractal_dimension(basins; range_ε)
+    return ε, N_ε, length(size(basins)) - d
```
Datseris, line 98: `length(size(basins)) - d`: shouldn't we reference the function `basins_fractal_dimension` with an `@ref` as well when you type `Δ₀ = 2 - α`? In the end, that's what we use!

src/basins/basins_utilities.jl
```diff
@@ -86,4 +86,144 @@ function _match_from_distance(b₋, a₋, ids₋, b₊, a₊, ids₊)
...
+
+
+
+
+"""
+    basin_entropy(basins; ε = 20) -> Sb, Sbb
+
+This algorithm computes the basin entropy `Sb` of a basins of attraction on a regular grid.
+It is a measure of the uncertainty on the initial conditions of the basins. The basin
+entropy is maximum at the value `log(n_att)` being `n_att` the number of attractor. In
```
Datseris, line 99:
```suggestion
entropy is maximum at the value `log(n_att)` being `n_att` the number of attractors. In
```

src/basins/basins_utilities.jl
```diff
@@ -86,4 +86,144 @@ function _match_from_distance(b₋, a₋, ids₋, b₊, a₊, ids₊)
...
+this case the boundary is intermingled: for a given initial condition we can find
+another initial condition that lead to another basin arbitriraly close. It provides also
+a simple criterion for fractality: if the boundary basin entropy `Sbb` is above `log(2)`
+then we have a fractal boundary. It doesn't mean that basins with values below cannot
+have a fractal boundary.
+
+The basin entropy is an average of the Gibbs entropy for a many boxes of side ε
+(along all dimensions). The function return the basin entropy `Sb` and the boundary
+basin entropy `Sbb` which is the entropy averaged only on the boxes on the boundary
+(the boxes with only one attractor are discarded).
```
Datseris, line 109: I think the docstring should be a bit more clear of what ε is. It is the "size" of a larger box, that contains that many actual "boxes" or "grid points", right? Because this sentence confused me a bit: I thought "wait, each box of `basins` can only have a single attractor, by definition, given how we compute it".

src/basins/basins_utilities.jl
```diff
@@ -86,4 +86,144 @@ function _match_from_distance(b₋, a₋, ids₋, b₊, a₊, ids₊)
...
     closeness
-end
\ No newline at end of file
+end
+
+
+
+
+"""
+    basin_entropy(basins; ε = 20) -> Sb, Sbb
```
Datseris, line 95: I'd suggest that `ε` becomes a positional argument and directly explained as soon as possible in the docstring.

src/basins/basins_utilities.jl
```diff
@@ -86,4 +86,144 @@ function _match_from_distance(b₋, a₋, ids₋, b₊, a₊, ids₊)
...
+end
+
+
+
+
+"""
+    basins_fractal_test(basins; ε = 20, Ntotal = 1000) -> test_res, Sbb
+
+This is an automated test to decide if the boundary of the basins has fractal structures.
+The bottom line is to look at the basins with a magnifier of size `ε` a litlle bit everywhere.
```
Datseris, line 153: "a litlle bit everywhere". First, litlle -> little. Second: I'm having trouble what "a little bit everywhere" means. Can you please rephrase it in different words?

src/basins/basins_utilities.jl
```diff
@@ -86,4 +86,144 @@ function _match_from_distance(b₋, a₋, ids₋, b₊, a₊, ids₊)
...
+If what we see in the magnifier looks like a smooth boundary we decide that the boundary if smooth.
```
Datseris, line 154:
```suggestion
If what we see in the magnifier looks like a smooth boundary we decide that the boundary is smooth.
```

Datseris (commented):
@awage I am very sorry for being so slow on this. This is a very high quality contribution. I've left some comments. I want to have a look at the performance of KDTree vs BallTree you've pointed out. Can you please provide me with the code you used to do this comparison, and I'll have a closer look?

src/basins/uncertainty_exp.jl
awage: > @awage I am very sorry for being so slow on this. This is a very high quality contribution. I've left some comments. I want to have a look at the performance of KDTree vs BallTree you've pointed out. Can you please provide me with the code you used to do this comparison, and I'll have a closer look?

Thanks for the review, on monday I will have some time to spend on this. As for the KDTree performance, I used KDTree in `NearestNeighbors` to have a fair comparison. I suspect the problem is linked to a `StaticVector` issue of some sort: 

```
using Neighborhood, BenchmarkTools

basins = rand(500,500)
v_pts = zeros(Float64, 2, 500*500)
I = CartesianIndices(basins)
# We fill v_pts with the indices as coordinates of each points (no need to use real coordinates)
for (k,coord) in enumerate(I)
     v_pts[:, k] = [Tuple(coord)...]
end

tree = searchstructure(KDTree, v_pts, Euclidean())
@btime for k in 1:1000
    # generate a random coordinate (not necesarilly on a grid point)
    p = [rand()*500, rand()*500]
    (idxs,r) = inrange(tree, p, 5)
end


using NearestNeighbors

btree = KDTree(v_pts)
@btime for k in 1:1000
    # generate a random coordinate (not necesarilly on a grid point)
    p = [rand()*500, rand()*500]
    idxs = NearestNeighbors.inrange(btree, p, 5, true)
end
```

Datseris: Okay, the answer is simple: you have to do `idxs = isearch(tree, p, WithinRange(5.0))` instead of using `search`. The latter computes all distances in an inefficient manner that I still have to resolve. NearestNeighbor's `inrange` also does not compute the distances.

awage:
I think I have addressed all comments. If something is missing let me know.

src/basins/uncertainty_exp.jl
```diff
@@ -1,99 +1,101 @@
...
-Estimate the uncertainty exponent[^Grebogi1983] of the basins of attraction. This exponent
-is related to the final state sensitivity of the trajectories in the phase space.
-An exponent close to `1` means basins with smooth boundaries whereas an exponent close
-to `0` represent complety fractalized basins, also called riddled basins.
-
-`xg`, `yg` are 1-dim ranges that defines the grid of the initial conditions.
-`basins` is the matrix containing the information of the basin.
-This functionality is currently implemented only for 2D basins.
+    basins_fractal_dimension(basins; kwargs...) -> V_ε, N_ε ,d
+Estimates the [Fractal Dimension](@ref) `d` of the boundary between basins of attraction using
```
Datseris, line 5:
```suggestion
Estimate the [Fractal Dimension](@ref) `d` of the boundary between basins of attraction using
```

src/basins/uncertainty_exp.jl
```diff
@@ -1,99 +1,101 @@
 export uncertainty_exponent
```
Datseris, line 1:
```suggestion
export uncertainty_exponent, basins_fractal_dimension
```

Datseris:
Awesome, new features are registered as version 1.32!

PR #200, opened 2021-07-25
Basins zoom

awage:
I introduced the "zoom" capability. When the attractors are known we can pass them to `basins_of_attraction`. 

It works but the user should be advised that it will require some tweaking of the parameter since you don't have a grid to work on. For example, the radius of search is critical. Also you need to bump `mx_chk_lost` to 1000 at least since the trajectory will spend most of the time outside the grid. 

I did not check if there is an impact on the performances. I think it is negligible.

### DIFF
--- a/src/basins/basins_lowlevel.jl
+++ b/src/basins/basins_lowlevel.jl
@@ -1,6 +1,6 @@
 import ProgressMeter
 
-mutable struct BasinInfo{B, IF, RF, UF, D, T, Q}
+mutable struct BasinInfo{B, IF, RF, UF, D, T, Q, K}
     basin::Array{Int16, B}
     grid_steps::SVector{B, Float64}
     grid_maxima::SVector{B, Float64}
@@ -16,6 +16,7 @@ mutable struct BasinInfo{B, IF, RF, UF, D, T, Q}
     prev_clr::Int
     attractors::Dict{Int16, Dataset{D, T}}
     visited::Q
+    search_trees::Vector{K}
 end
 
 
@@ -27,11 +28,17 @@ at a new full state, given the state on the grid.
 """
 function draw_basin!(
         grid::Tuple, integ, iter_f!::Function, complete_and_reinit!, get_projected_state::Function;
-        show_progress = true, kwargs...,
+        show_progress = true, attractors = nothing, kwargs...,
     )
     B = length(grid)
     D = length(get_state(integ)) # dimension of the full state space
     complete = false
+    trees = Vector{KDTree{}}()
+    if !isnothing(attractors)
+        for k in keys(attractors)
+            push!(trees, searchstructure(KDTree, attractors[k], Euclidean()))
+        end
+    end
     grid_steps = step.(grid)
     grid_maxima = maximum.(grid)
     grid_minima = minimum.(grid)
@@ -46,7 +53,8 @@ function draw_basin!(
         :att_search,
         2,4,0,1,1,
         Dict{Int16,Dataset{D,eltype(get_state(integ))}}(),
-        Vector{CartesianIndex{B}}()
+        Vector{CartesianIndex{B}}(),
+        trees
     )
     reset_basin_counters!(bsn_nfo)
     I = CartesianIndices(bsn_nfo.basin)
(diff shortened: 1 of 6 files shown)


awage:
By the way I fixed a mistake in `basins_test_fractal` nothing very important.

Datseris:
Thanks @awage! So this introduces two new keywords to `basins_of_attraction`, so we need to add these in our documentation string!

src/basins/basins_lowlevel.jl
```diff
@@ -27,11 +28,17 @@ at a new full state, given the state on the grid.
...
     )
     B = length(grid)
     D = length(get_state(integ)) # dimension of the full state space
     complete = false
+    trees = Vector{KDTree{}}()
+    if !isnothing(attractors)
+        for k in keys(attractors)
+            push!(trees, searchstructure(KDTree, attractors[k], Euclidean()))
+        end
+    end
```
Datseris, line 41: This is not good for performance. `KDTree{}` is **not** a concrete type. So, `Vector{KDTree{}}` isn't one either. Computations involving the elements of `trees` will be inefficient. I would say that instead of declaring `search_trees::Vector{K}` in `BasinInfo`, we should declare directly `search_trees::K`. So the type-parameterization is `K`. This way, if the given `attractors` are nothing, then `searc_trees` also becomes `nothing`. Else, it gets its value directly, and also without pusing. You can do

`trees = [searchstructure(KDTree, att, Euclidean()) for att in values(attractors)]`

src/basins/basins_lowlevel.jl
```diff
@@ -129,8 +137,18 @@ and the trajectories staying outside the grid are coded with -1.
...
         mx_chk_att = 2, mx_chk_hit_bas = 10, mx_chk_fnd_att = 100, mx_chk_lost = 100,
-        horizon_limit = 1e6
+        horizon_limit = 1e6, ε = 1e-5
     )
+    # search attractors directly
+    if !isempty(bsn_nfo.search_trees)
+        for (k,t) in enumerate(bsn_nfo.search_trees)
+            idxs = isearch(t, u_full_state, WithinRange(ε))
+            if !isempty(idxs)
+                return  2*k + 1
```
Datseris, line 147: Question: here I think we lose the attractor ID information, right? Because when you add attractors to the `searc_tree`, you don't keep their ID, as dictionaries are not sorted objects. So the user gives atteractors with a specific ID, but they then get a somewhat random ID.

awage: I didn't see that one. Now it is reordered in such a way that the numbers matches. 

I think it will be a little bit faster if we allow a transient, but it implies substantial changes in the code. I am not completely satisfied because I feel that there are some loose ends. For example: what happens if you change the step size? Maybe you never find the attractor. It should be stated in the docstring that it is not as reliable as the method on the grid. 

I will add the keyword arguments in the docstring.

Datseris:
I think we are still not matching attractor IDs properly. I beleive we have to completely avoid using a `Vector` for the trees, and use a dictionary which maps attractor IDs to Trees. Because, for example, have a look at this test case:

```
using PyPlot, DynamicalSystems

ds = Systems.magnetic_pendulum(γ=1, d=0.2, α=0.2, ω=0.8, N=3)
xg = range(-2,2,length=100)
yg = range(-2,2,length=100)
complete_state(y) = SVector(0,0)
basin, attractors = basins_of_attraction((xg,yg), ds;
idxs=1:2, complete_state, show_progress = false)
figure(); pcolormesh(xg,yg, basin')

# Now test the zoom capability
xg = yg = range(-2,-1.5,length=100)
basins, att = basins_of_attraction((xg,yg), ds;
idxs = 1:2, complete_state, show_progress = false,
attractors = attractors, mx_chk_lost = 1000, ε = 1e-3)

figure(); pcolormesh(xg,yg, basins')
```

image

The colors are not consistent between the two figures. So we mixup the attror ids by sorting a vector. In addition, the current method wouldn't work if a user gave attractors that only had ID 2 and 3, but not 1, which is e.g. the case if one uses the `match_attractors!` function.

Datseris:
Yeap, I can confirm: the latest change I did fixed the issue! :)

Datseris:
@awage  I've added docs and stuff. There are some final sentences remaining to be written at the docs which I'll leave to you.

awage:
Great! Tomorrow morning I finish the docs. I realized that to work correctly ε should be about the size of the grid cell where the attractors have been computed initially. 

Also I will try this feature with other systems.

awage:
I tested the feature with other systems and it works better than I expected. It is much slower though. If it is becomes a "popular" feature it can be improved if we skip a transient time before looking for attractors. 

The benchmarks for this PR and master are almost identical so there is no loss in performance. 

For me it is good to merge.

PR #207, opened 2021-08-18
Improvements of boxed correlation sum

Datseris:
@apbraun let's do all necessary improvements in this PR.

### DIFF
--- a/src/dimensions/correlationdim.jl
+++ b/src/dimensions/correlationdim.jl
@@ -3,7 +3,7 @@
 #######################################################################################
 using Distances, Roots
 export correlationsum, grassberger_dim, boxed_correlationsum,
-estimate_r0_buenoorovio, data_boxing
+estimate_r0_buenoorovio, data_boxing, autoprismdim
 
 """
     correlationsum(X, ε::Real; w = 0, norm = Euclidean(), q = 2) → C_q(ε)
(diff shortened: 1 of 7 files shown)


src/dimensions/correlationdim.jl
```diff
@@ -230,11 +231,16 @@ function boxed_correlationsum(
...
+# An algorithm to find the ideal choice of a prism size for the boxed correlation dimension.
+# See Theiler as mentioned in `boxed_correlationsum`
 function _autoprismdim(data)
     D = dimension(data)
-    L = length(data)
-    # The best choice given by Theiler is $P = 0.5\log_2N$ and should be used when $D$ exceeds $0.75\log_2N$.
-    return size(data, 2)
+    N = length(data)
+    if D > 0.75 * log2(N)
+        return ceil(0.5 * log2(N))
```
Datseris, line 240: Should we have this be inside a `max(2, ...)` ? I think we say in the paper someone used always 2?

apbraun: Yes that seems reasonable.

Datseris:
@apbraun can you please remind me: these are all the improvements we wanted to do, right? I can merge away?

apbraun:
Yes these are all improvements. We spoke about adjusting the `_autoprism` algorithm to the Theiler version and catch `r0` < `ε0` in the `boxed_correlationsum`, I also spotted another bug in `boxed_correlationsum` that I fixed, so from my part we're ready to go.

Datseris:
@apbraun please give me some help here. In our paper you said that "Bueno-Orovio and Pérez-García chose P=2 for D>2" where I assume that P stands for the Prism dimension. Here we don't do that, but we use I think an older approach by Theiler. Which one should we use in the end...?

PR #214, opened 2021-09-15
Rewritten rescale in lyapunovs.jl

SudoMishra:
I rewrote the rescale function in lyapunov.jl with get_state and set_state functions as you had mentioned.
In the current form of the rescale function,  we no longer need multiple rescale functions as that can be handled by the multiple dispatches of get_state and set_state for different scenarios.

I'm unsure about the type of declarations that would go with the new rescale function but I think as now it handles both vector and matrix scenarios, the type declaration might not be needed.

I'm new to Julia and git. This is my first PR. I'm quite interested in contributing more to this package.
Thank you.

### DIFF
--- a/src/chaosdetection/lyapunovs.jl
+++ b/src/chaosdetection/lyapunovs.jl
@@ -337,7 +337,7 @@ function rescale!(integ,a,r::SVector)
 end
 function rescale!(integ,a,r::AbstractArray)  
     r .= get_state(integ,1) .+ (get_state(integ,2).-get_state(integ,1))./a
-    set_state!(integ,r,2)
+    set_state!(integ,copy(r),2)
 end
 
 
--- a/src/chaosdetection/lyapunovs.jl
+++ b/src/chaosdetection/lyapunovs.jl
@@ -269,16 +269,19 @@ end
 function lyapunov(pinteg, T, Ttr, Δt, d0, ut, lt)
     # transient
     t0 = pinteg.t
+    # array to preallocate state for rescale
+    r = copy(get_state(pinteg,2))
+    
     while pinteg.t < t0 + Ttr
         step!(pinteg, Δt)
         d = λdist(pinteg)
-        lt ≤ d ≤ ut || rescale!(pinteg, d/d0)
+        lt ≤ d ≤ ut || rescale!(pinteg, d/d0, r)
     end
 
     t0 = pinteg.t
     d = λdist(pinteg)
     d == 0 && error("Initial distance between states is zero!!!")
-    rescale!(pinteg, d/d0)
+    rescale!(pinteg, d/d0, r)
     λ = zero(d)
     while pinteg.t < t0 + T
         d = λdist(pinteg)
@@ -291,7 +294,7 @@ function lyapunov(pinteg, T, Ttr, Δt, d0, ut, lt)
         # local lyapunov exponent is simply the relative distance of the trajectories
         a = d/d0
         λ += log(a)
-        rescale!(pinteg, a)
+        rescale!(pinteg, a, r)
     end
     # Do final rescale, in case no other happened
     d = λdist(pinteg)
@@ -328,27 +331,13 @@ function λdist(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}) where {Alg, I
 end
 
 # Rescales:
-function rescale!(integ::MinimalDiscreteIntegrator{true, Vector{S}}, a) where {S<:SVector}
-    integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
+function rescale!(integ,a,r::SVector)  
+    r = get_state(integ,1) .+ (get_state(integ,2).-get_state(integ,1))./a
+    set_state!(integ,r,2)
 end
-function rescale!(integ::MinimalDiscreteIntegrator{true, Vector{S}}, a) where {S<:Vector}
-    @. integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
-end
-function rescale!(integ::AbstractODEIntegrator{Alg, IIP, M}, a) where {Alg, IIP, M<:Matrix}
-    for i in 1:size(integ.u)[1]
-        integ.u[i, 2] = integ.u[i,1] + (integ.u[i,2] - integ.u[i,1])/a
-    end
-    u_modified!(integ, true)
-end
-function rescale!(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}, a) where {Alg, IIP, S<:Vector}
-    @. integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
-end
-function rescale!(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}, a) where {Alg, IIP, S<:SVector}
-    integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
+function rescale!(integ,a,r::AbstractArray)  
+    r .= get_state(integ,1) .+ (get_state(integ,2).-get_state(integ,1))./a
+    set_state!(integ,r,2)
 end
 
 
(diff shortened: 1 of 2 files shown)


Datseris (commented):
That's a good start. However, now each time `rescale!` is called it allocates a new vector `r`. The best is to pre-allocate `r = copy(u)`. Then, you can use broadcasting `r .= u .+ (...)...` so that you overwrite an existing `r` in-place instead of creating a new one each time.

src/chaosdetection/lyapunovs.jl
```diff
@@ -269,16 +269,18 @@ end
 function lyapunov(pinteg, T, Ttr, Δt, d0, ut, lt)
     # transient
     t0 = pinteg.t
+    # array to preallocate state for rescale
+    r = [get_state(pinteg,1)]
```
Datseris, line 273: I don't understand. Why is this a vector of vectors, instead of just `copy(get_state(pinteg, 2))`...?

src/chaosdetection/lyapunovs.jl
```diff
@@ -328,12 +330,14 @@ function λdist(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}) where {Alg, I
 end
 
 # Rescales:
-function rescale!(integ,a)
-    r = get_state(integ,1) + (get_state(integ,2)-get_state(integ,1))./a
-    set_state!(integ,r,2)
+function rescale!(integ,a,r)
+    # r =[get_state(pinteg,1)] for preallocated computation  
+    r[1] = get_state(integ,1) + (get_state(integ,2)-get_state(integ,1))./a
```
Datseris, line 335: This is still allocating. _A lot_. You need to broadcast the operations by adding `@.` before `r = ...`.

(and you will replace `r[1]` with `r`, as there is no reason to define `r` as a vector of vectors.)

Datseris:
Huh, why was this PR closed? It was going good with only a small amount of tweaking needed.

SudoMishra:
Oh sorry, I was trying to make a PR on my forked repo and I think I messed up. I closed this PR. I'll reopen it.

src/chaosdetection/lyapunovs.jl
SudoMishra: I was trying this `copy(get_state(pinteg,2))` but then I get an r as an `SVector`. With that when I try to do 
`@. r=get_state(integ,1) + (get_state(integ,2)-get_state(integ,1))/a` I get a set_index error as you cannot change the contents of an SVector.

SudoMishra:
Oh sorry, I was trying to make a PR on my forked repo and I think I messed up. I closed this PR. I'll reopen it.

Datseris:
@SudoMishra test fail so something is going wrong with the way you rescale, can you have a closer look?

src/chaosdetection/lyapunovs.jl
```diff
@@ -328,27 +331,9 @@ function λdist(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}) where {Alg, I
...
-function rescale!(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}, a) where {Alg, IIP, S<:Vector}
-    @. integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
-end
-function rescale!(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}, a) where {Alg, IIP, S<:SVector}
-    integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
+function rescale!(integ,a,r)  
+    r .= get_state(integ,1) .+ (get_state(integ,2).-get_state(integ,1))./a
+    set_state!(integ,r,2)
```
Datseris, line 336: You need two versions of this function that dispatch via multiple dispatch on `r`. If `r::SVector` then you cannot use `.=` and you just type `r = ` (leaving the rest the same). If instead `r::AbstractArray` then you use the above version.

SudoMishra: I've added the multiple dispatched rescale function. So test 5 is failing in `lyapunov_exponents.jl` and its a bit strange why that is happening. It is the case where `r::Abstract_Array`
When I do rescale with no preallocation i.e.
`r = get_state(integ,1) .+ (get_state(integ,2).-get_state(integ,1))./a`
the test is passing and I'm getting the correct lyapunov exponent.
However, when I simply change the above line to use the preallocated r, i.e.
` r .= get_state(integ,1) .+ (get_state(integ,2).-get_state(integ,1))./a`
The test fails as the lyapunov value that I get is wrong (around 18.3..).
I'm trying to figure out why this is happening.

SudoMishra: Hi @Datseris , I was trying to resolve this issue but its getting a bit strange. So, in the lyapunov function I've created a new variable `r=copy(get_state(pinteg,2))`. As I've used `copy()` for creating `r`, it should now be independent of `pinteg`. So it should only change when the `rescale!` function is called, where it used as a placeholder for preallocation. However, that is not the case, and it even changes outside the `rescale!`. Due to this, during the `set_state!`  in `rescale!` a wrong value is being passed on which results in the test failing. I'm not sure why this is happening, I've also used `deepcopy` instead of `copy` but the problem still persists. Am I missing some basic thing here? I would be grateful if you could give some advice on how to resolve this.

Datseris: Can you repeat the same with `deepcopy` on `r`?

SudoMishra: Yeah, it is happening even with `deepcopy`. I was visualising the variables using the debugger and even with `deepcopy` the `r` variable gets updated during `step!` . And apparently this is only affecting one of the test cases.

SudoMishra: I haved asked about this on julia [discourse](https://discourse.julialang.org/t/variable-gets-updated-without-even-being-referenced-advice-on-debugging-this/68612) as well. I have posted a gif showing the bug happening as well.

SudoMishra: Hi @Datseris, in `set_state!` this is happening :  `integ.u[k] = u` . This makes a reference to `r` that I've passed because of which during `step!` when `integ.u` changes `r` also changes.

Datseris: Yeah, you are right. Sorry this is my mistake. ` integ.u[k] = u ` must only be done for `u::SVector`. For `u::Vector` this must be ` integ.u[k] .= u`  (notice the dot).

This needs to be fixed by adding one more dispatch method to DynamicalSystemsBase.jl. Would you please do this via a PR there...?

Datseris: Actually, I"m still struggling to understand what exactly is your problem with respect to the algorithm. Why does it matter if `r` changes? `r` is a dummy variable. Its correct value is set exactly before giving this to `set_state`. Even if `r` and the 2nd state are referencing each other, we still don't care.  ...?

SudoMishra: I'm actually a bit confused myself. Even I think that since the correct value of `r`  is getting changed just before being sent to `set_state` it shouldn't matter and the algorithm should work. But for some reason, test 5 in `lyapunov_exponents` was failing. I tried to figure out why this was happening and the only thing wierd thing that I saw happening was that r was changing. I still think my code should be working just by `set_state!(integ,r,2)` but its failing for that specific case. I'll try to add the dispatch method to DynamicalSystemsBase.jl .

Datseris: Okay. I'll look over this locally over the weekend and will identify the error. Thanks for investigating as well.

SudoMishra: I have fixed the `set_state!` and opened a PR in `DynamicalSystemsBase.jl`. Using the corrected `set_state`, `set_state!(integ,r,2)` is working properly and its passing all tests. I can commit the proper `rescale` function once the `DynamicalSystemsBase.jl` PR is merged and  ChaosTools uses that.

SudoMishra: I have added the updated `rescale` function without `copy(r)` in `set_state`. One test case will not pass currently as `ChaosTools.jl` currently uses the wrong `set_state` function from `DynamicalSystemsBase.jl` . I have added a PR to fix that in `DynamicalSystemsBase.jl`. For `u::AbstractVector`, `set_state` should use `.=`.

Issue #211, opened 2021-09-08
basin stability of multi-stable dynamical systems using clustering


Datseris:
https://link.springer.com/article/10.1007%2Fs11071-021-06786-5

Method from Stender & Hoffman

This paper implements a method to calculate basin stability that works for high dimensional systems efficiently, by combining Monte Carlo sampling and a clustering algorithm. Perhaps it can interplay well with our existing `basins_of_attraction` and tipping points related stuff.

The method is implemented open source in matlab so it should be an easy port.

KalelR:
Hi, I'm interested in doing this! I can work on it and open a PR once I have a first version ready. Would that be ok?

Datseris:
hI @KalelR , great that you say this. I'm also interested in working on this and I have already sketched out the documentation string of the function. Tell you what: I first open a pull request sketching out the function input and output, and then you can take it from there?

KalelR:
Sure, that'd be great!

awage:
If you don't mind I would like to participate. I am very interested in the sampling aspect of the basin. It will be very interesting for systems with high dimension (for example a network of coupled oscillators).

KalelR:
Oh, that's amazing, I'd be interested in networks also. I have implemented a decent part of the code already (I had started a bit before commenting here). Once @Datseris opens the pull request, I can commit what I have so far.

Datseris:
Implemented in https://github.com/JuliaDynamics/ChaosTools.jl/pull/219

PR #216, opened 2021-09-27
bSTAB implementation

Datseris:
WIP. see https://github.com/JuliaDynamics/ChaosTools.jl/issues/211


Datseris:
@KalelR , @awage here is the sketch. Kale feel free to take this away. Since you don't have push access to ChaosTools.jl yet (but you probably will after this PR), you can open a Pull Request from your fork _into this branch_.

PR #217, opened 2021-09-28
bSTAB implementation continuation. 

KalelR:
Added code for calculating stability of basins, building on top of @Datseris first pull request for the bSTAB algorithm by Stender and Hoffmann. I tested and replicated the first two results of the paper, on the damped pendulum and the duffing oscillator. I can add these tests in the /tests directory, if needed.
The code still has to be improved, optimized, and better documented, but I thought I'd share what I have so far. 

Also, hope I am doing this PR correctly. Please let me know otherwise!

### DIFF
--- a/src/basins/basins_fractions.jl
+++ b/src/basins/basins_fractions.jl
@@ -84,11 +84,11 @@ conditions to attractors using a clustering technique.
 """
 function basin_fractions(ds::DynamicalSystem, feature_extraction::Function, ic_grid::AbstractMatrix, clust_mod="supervised"; kwargs...)
     #-- Integrate ICs to find features
-    println("Integrating ICs to find features.")
+    # println("Integrating ICs to find features.")
     feature_array = featurizer(ds, ic_grid,  feature_extraction; kwargs...)
 
     #-- Classify features
-    println("Classifying features in $(clust_mod) manner.")
+    # println("Classifying features in $(clust_mod) manner.")
     if(clust_mod == "supervised")
         y0 = kwargs[:y0]
         feature_templates = featurizer(ds, y0, feature_extraction; kwargs...) #TODO: could allow the user to directly pass the templates featuers also, instead of just their initial conditions
--- a/src/basins/basins_fractions.jl
+++ b/src/basins/basins_fractions.jl
@@ -1,6 +1,6 @@
 export basin_fractions
 using Statistics: mean
-using NearestNeighbors #for kNN
+using Neighborhood #for kNN
 using Distances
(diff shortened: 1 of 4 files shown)


src/basins/basins_fractions.jl
```diff
@@ -1,5 +1,13 @@
 export basin_fractions
+using Statistics: mean
+using NearestNeighbors #for kNN
```
Datseris, line 3: We do not use NearestNeighbors anymore, but rather the more intuitive Neighborhood.jl interface, which I've built specifically because it made DynamicalSystems.jl related stuff so much more straightforward.  https://julianeighbors.github.io/Neighborhood.jl/dev/  

When I review the remaining of the code, I'll tell you how you can transform it to that format (it is quite easy). You can also look at `lyapunov_from_data` function to see how I use it there.

Datseris:
Awesome. I'll give a review in some days from now. In the meantime, yes, please add the tests here. Tests are important because they show us also how the interface is used.

src/basins/basins_fractions.jl
KalelR: Oh, okay. It really is quite easy to transform, so I did that already.

test/basins/stability_basin_tests.jl
```diff
@@ -0,0 +1,122 @@
...
+    @test S[3] == 0.1962
+    @test S[4] == 0.028 
+    @test S[5] == 0.0248
+
+    #plot basins
+    # cmap = matplotlib.colors.ListedColormap(["red", "gray", "orange", "cyan", "blue"])
+    # fig = figure()
+    # scatter(ic_grid[1,:], ic_grid[2,:], c=class_labels, cmap=cmap)
+    # gca().set_aspect("equal")
+    # colorbar()
```
Datseris, line 120: It's great that you added this, which allows local examination! Let's leave it here even after merge. Just remember: you need to `include` the test file in the `test/runtests.jl` for it to actually be automatically tested by the CI ;)

KalelR: Yeah, the figure fits really well with the paper. I included the test the runtests also.

Datseris:
@KalelR I think it is wasteful that this PR doesn't count for you towards Hacktoberfest. I propose that you simply take the entire source code you have now and open a new PR directly to ChaosTools.jl master, so that this counts for your Hacktoberfest progress. (Trust me, this T-shirt is something you don't want to miss!!!!)

KalelR:
All right, done. Thanks for the tip!

PR #219, opened 2021-10-01
bSTAB implementation

KalelR:
Hey. Following @Datseris advice, I am opening this PR towards ChaosTools.jl master, for it to count towards my Hacktoberfest. It's the same code as in the PR towards the basin_fractions branch.


Datseris:
I will be reviewing this soon, but given the quality and extent of the contribution, I can already put the hacktoberfest-accepted tag.

KalelR:
Great to hear! Thanks  🙂

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+* templates_labels : (for supervised) labels for the templates
+* clust_params = NamedTuple() : other parameters for clustering method
+* diffeq = NamedTuple() : other parameters for the solvers of DiffEqs
+* extract_params = NamedTuplee() : other parameters for the feature_extraction function
+* min_neighbors = 10 : (for unsupervised) minimum number of neighbors (similar features) each feature needs to have in order to be considered in a cluster (fewer than this, it is labeled as an outlier, id=-1). This is somewhat hard to define, as it directly interferes with how many attractors the clustering finds. The authors use it equal to 10 always.
+
+## Available featurizers
+* lyapunov spectrum
+* statistical moments
+* whatever else in the original paper, which should be trivial to implement here
```
Datseris, line 65: Is this still true?

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+Let ``F(A)`` be the fraction the basin of attraction of an attractor ``A`` has in the
+chosen state space region ``\\mathcal{S}`` given by `sampler`. `basin_fractions` estimates ``F`` for all
+attractors in ``\\mathcal{S}`` by randomly sampling `N` initial conditions and counting which
+ones end up in which attractors. The error of this approach for each fraction is given by[^Menck2013]
+``e = \\sqrt{F(A)(1-F(A))}``.
+
+In `basin_fractions` we do not actually identify attractors, as e.g. done in [`basins_of_attraction`](@ref).
+Instead we follow the approach of Stender & Hoffmann[^Stender2021] which transforms each
+initial condition into a vector of features, and uses these to efficiently map initial
+conditions to attractors using a clustering technique.
```
Datseris, line 78: You need to improve this with much more information. It was just an initial sketch I wrote to guide the plan. For example what is supervised or not, and the arguments and keyword arguments, can be incorporated here for more details.

You can have a look at the various documentation strings of the DynamicalSystems.jl library, it will give you a good idea of the overall style.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+
+
+## Main arguments
+* ds: DynamicalSystem; so far only Continuous implemented
+* feature_extraction::Function : function that receives as input the trajectory, time vector, and optional parameters to return a matrix containing the features extracted from y, with the j-th feature in the j-th column of the matrix.
+* ic_grid::AbstractMatrix : matrix containing the grid of initial conditions; Each column contains one IC; so it is of size dimensionality x N
+* clust_mod = "unsupervised" : clustering method chosen, can be "supervised" or "unsupervised". If "supervised", user needs to input the templates features also
+
+
+## Keywork arguments
```
Datseris, line 51:
```suggestion
## Keyword Arguments
```

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+conditions to attractors using a clustering technique.
+
+
+[^Menck2013]: Menck, Heitzig, Marwan & Kurths. How basin stability complements the linear stability paradigm. [Nature Physics, 9(2), 89–92](https://doi.org/10.1038/nphys2516)
+
+[^Stender2021]: Stender & Hoffmann, [bSTAB: an open-source software for computing the basin stability of multi-stable dynamical systems](https://doi.org/10.1007/s11071-021-06786-5)
+"""
+function basin_fractions(ds::DynamicalSystem, feature_extraction::Function, ic_grid::AbstractMatrix, clust_mod="supervised"; kwargs...)
+    #-- Integrate ICs to find features
+    # println("Integrating ICs to find features.")
```
Datseris, line 87: Such comments are not helpful and make the source code larger nevertheless. I can advise you having a look at this workshop I gave about the use of comments and other things in code: https://youtu.be/iIIBFujl254

This applies to all subsequent comments that are not useful.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+    feature_array = featurizer(ds, ic_grid,  feature_extraction; kwargs...)
+
+    #-- Classify features
+    # println("Classifying features in $(clust_mod) manner.")
+    if(clust_mod == "supervised")
```
Datseris, line 92:
```suggestion
    if clust_mod == "supervised"
```

I have never-ever seen an `if` statement enclosing paranthesis. Where did you find that? Is it even a valid Julia syntax...?

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+        y0 = kwargs[:y0]
```
Datseris, line 93: What do you do if he user has not provided a keyword `y0`? Use instead `y0 = get(kwargs, :y0, default_value)`. Furthermore, replace all `y` and `y0` with `u` and `u0` throughout the code here. It is `u` that is used as the symbol for a state, not `y`.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+        feature_templates = featurizer(ds, y0, feature_extraction; kwargs...) #TODO: could allow the user to directly pass the templates featuers also, instead of just their initial conditions
+        class_labels, class_errors, template_labels_id = classify_solution(feature_array, feature_templates; kwargs...);
+    else
+        class_labels, class_errors, template_id = classify_solution(feature_array; kwargs...);
+    end
+
+    #-- Compute basin stability values
+    S = basin_fractions(class_labels)
```
Datseris, line 101: The line of code says very obviously, and very clearly, whats happening. How is this comment helpful?

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+on a state space box.
+`featurizer` is a function that takes as an input an initial condition and outputs a
+vector of "features". Various pre-defined featurizers are available, see below.
+
+
+
+## Main arguments
+* ds: DynamicalSystem; so far only Continuous implemented
+* feature_extraction::Function : function that receives as input the trajectory, time vector, and optional parameters to return a matrix containing the features extracted from y, with the j-th feature in the j-th column of the matrix.
+* ic_grid::AbstractMatrix : matrix containing the grid of initial conditions; Each column contains one IC; so it is of size dimensionality x N
```
Datseris, line 47: I am really confused here. Shouldn't the user provide an abstract sampler function that samples initial conditions in a given state space region? If the user provides exact initial conditions, the entire "random sampling of a bernoulli process" becomes meaningless...?

src/basins/basins_fractions_utilities.jl
```diff
@@ -0,0 +1,81 @@
+
+
+#------------------ GENERATING GRID OF INITIAL CONDITIONS AS IN bSTAB PAPER; MAY BE DEPRECATED AFTER IMPLEMENTING THE SAMPLER METHOD
+
+"""
+Function to generate the grid of initial conditions necessary for basins_fractions.
+So far, the only implmented method is "uniform", which is probably the most used one.
```
Datseris, line 7: In the paper, how do they justify non-uniform sampling? If it is non uniform then the error measure of a bernoulli process does not apply. A strong part of the random sampling process is that the error goes to zero with the square root of the measure.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
+export basin_fractions
```
Datseris, line 1: General comment: please restrict your code so that each line of code never exceeds 92 characters.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+within a subset of the state space. You can use [`boxregion`](@ref) to create a `sampler`
+on a state space box.
+`featurizer` is a function that takes as an input an initial condition and outputs a
+vector of "features". Various pre-defined featurizers are available, see below.
+
+
+
+## Main arguments
+* ds: DynamicalSystem; so far only Continuous implemented
+* feature_extraction::Function : function that receives as input the trajectory, time vector, and optional parameters to return a matrix containing the features extracted from y, with the j-th feature in the j-th column of the matrix.
```
Datseris, line 46: The featurizer returns a _matrix_, not a _vector_ of real values? What if the features I chose are the mean and the standard deviation of the trajectory...? How do you make these 2 real numbers into a matrix so that the j-th feature is the j-th column of the matrix?

src/basins/basins_fractions.jl
Datseris: What's your plan for default featurizers? How do you think its API should be? Can you provide an implementation plan for this?

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+#----- INTEGRATION AND FEATURE EXTRACTION
+"""
+`featurizer` receives the grid of ICs and returns their extracted features in a matrix.
+* templates_features :  matrix containing the template features used in the supervised method, j-th column containing the j-th feature template
+Returns Array{Float64, 2}, with the j-th column containing the features of the j-th trajectory
+"""
+function featurizer(ds, ic_grid::AbstractMatrix, feature_extraction::Function;  kwargs...)
+    Nactual = size(ic_grid, 2) #number of actual ICs (may be a bit more than N due to rounding)
+    feature_array = [Float64[] for i=1:Nactual]
+    for i = 1:Nactual #TODO: implement parallelization, if necessary
```
Datseris, line 115: Oh yeah it is certainly possible to parallelize this, because it is very much straightforward application of an `@threads` macro.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+The featurizer thus: 1. integrates the initial conditions; extracts its features via a (at-least-for-now) user defined feature extraction function
+Return is the return type of feature_extraction
+"""
+function featurizer(ds, u0::AbstractVector, feature_extraction; T=100, Ttr=0, Δt=1, extract_params=NamedTuple(), diffeq=NamedTuple(), kwargs...)
+        # perform the time integration
+        y = trajectory(ds, T, u0; Ttr=Ttr, Δt=Δt, diffeq) #TODO: maybe starting an integrator and using re_init! is better # reinit!(integ, ic_grid[i,:])
+        t = Ttr:Δt:T+Ttr
+        # extract descriminative features from the time signals
+        feature = feature_extraction(t, y, extract_params)
+        return feature
```
Datseris, line 138: The spacing is off here. Please be sure that each identation level is exactly 4 spaces throughout the entire file.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+end
+
+
+#----- INTEGRATION AND FEATURE EXTRACTION
+"""
+`featurizer` receives the grid of ICs and returns their extracted features in a matrix.
+* templates_features :  matrix containing the template features used in the supervised method, j-th column containing the j-th feature template
+Returns Array{Float64, 2}, with the j-th column containing the features of the j-th trajectory
+"""
+function featurizer(ds, ic_grid::AbstractMatrix, feature_extraction::Function;  kwargs...)
```
Datseris, line 112: There are two featurizer functions but I have not really understood why or what's their difference.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+and classifies these according to the templates received. This is therefore the supervised method.
+The templates are also evolved from the initial conditions y0, and are a matrix with each template in a column.
+The classification is done using a first-neighbor clustering method, identifying, for each feature,
+the closest template.
+
+label : for each feature, it is the id of the closest template == index of the template in y0
+error : distance from the feature to its closest template (in the specified metric)
+class_labels = Array{Int64,1} with the labels for each feature
+class_errors = Array{Float64,1} with the errors for each feature
+template_ids = Array{Int64, 1} with the labels for each template. Not really needed.
```
Datseris, line 155: 1. Make this a bullet list starting with `*`
2. Things that are code must be written as `code`. Do not use argument names, keyword argument names, or their values in plain text. Programming is the most precise language in the universe. If something is code, it really needs to be code. So, write `template_ids = Array{Int64, 1}` instead of just "template_ids = Array{Int64, 1} ".

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
+export basin_fractions
+using Statistics: mean
+using Neighborhood #for kNN
+using Distances
+using DataFrames
```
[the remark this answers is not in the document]
Datseris: Okay, I went through most of the code and I couldn't find any real justification for dataframes. We cannot be adding dependencies unless it is absolutely critical for the algorithm. Please either provide strong arguments for having DataFrames as a dependency, or remove it.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+The classification is done using a first-neighbor clustering method, identifying, for each feature,
+the closest template.
+
+label : for each feature, it is the id of the closest template == index of the template in y0
+error : distance from the feature to its closest template (in the specified metric)
+class_labels = Array{Int64,1} with the labels for each feature
+class_errors = Array{Float64,1} with the errors for each feature
+template_ids = Array{Int64, 1} with the labels for each template. Not really needed.
+
+dist_norm = metric for calculating the distance in kNN. Possible examples are 'seuclidean', 'cityblock', 'chebychev', 'minkowski', 'mahalanobis', 'cosine', 'correlation', 'spearman', 'hamming', 'jaccard'.
```
Datseris, line 157: Where is `dist_norm` used? I cannot see it anywhere.

Datseris (changes requested):
@KalelR while this is a great first start, there is definitely a some work remaining before we merge this. Given the extent of the PR, we will probably need to have 2-3 more rounds of review. 

I've left many specific comments, and I'll leave much more in the second round. For now, here are some things that are crucial to do:

1. Provide a high-level description of the algorithm you use and give me a high level guide of the source code. (provide this as a comment in this PR). This will help me get a better idea of what's going on without loosing myself in the details of the source, and will better allow me to guide you differently on designing a good API, and differently on e.g., performance optimizations, clear source code, better library integration, etc.
2.  The high-level description you will provide will also be used in the `## Description` section of the documentation string.
3. Please provide here figures of applying the algorithm and showing it works. You already produced the figures so just paste them here.

Once you've addressed this review I'll go for the second round.

Datseris:
Ah, also have a look at : https://docs.julialang.org/en/v1/manual/documentation/

src/basins/basins_fractions_utilities.jl
KalelR: Yup, that's true. In the paper, they say the
>  "bounded uniform distribution at random is the classical choice that guarantees basin stability values to be proportional to the true basins’ volumes for N → ∞" .

This is indeed their choice for all the examples. 
But they also acknowledge alternatives, like non-uniform sampling, can be useful, in case the user wants.  
 > "However, in certain situations different sampling strategies may be considered. For  example, if a distribution of initial conditions or admissible perturbations is known from previous studies or experiments, they can be supplied through a custom definition of ρ. In the latter case, the basin stability values will no longer be an estimate for the basins’ volumes. Instead, the basin stability values will indicate probabilities for observing a specific steady-state solution ȳ under the distribution ρ". > 

Since the algorithm works in the same way for both cases, regardless of how the initial conditions are chosen, I thought to leave both options available. They do that also in their algorithm. I can add this info in the documentation.

src/basins/basins_fractions.jl
KalelR: For this first go at the implementation, I chose to ask for the initial conditions as input, following how the authors do in their MATLAB code, because then
 (i) the user would have more control over the initial conditions, which can be useful if they want something other than the uniform random sampling (eg see the comment below this one). If they want the random uniform sampling of the distribution, they can just use the `generate_independent_uniform_distribution` function that is provided. Sorry if I'm misunderstanding something about this process, but as I understand, providing the ICs is just applying the sampler function, saving the results, and sending them to `basins_fractions`.  
 (ii) the user could more easily plot the basins of attraction (like in the nice-looking figures they have in the paper, which I have replicated, and am sending in a comment below); 
(iii) I thought it would also clean up the API, because this way we wouldn't have to ask for all the inputs necessary for generating the grid.

I'm happy to change this, however, if I'm incorrect and/or if you prefer another way :)

src/basins/basins_fractions.jl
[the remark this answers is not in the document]
KalelR: It is not necessary, I used it initially to result a DataFrame with the cluster's labels and errors, but I removed it now.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+`featurizer` is a function that takes as an input an initial condition and outputs a
+vector of "features". Various pre-defined featurizers are available, see below.
+
+
+
+## Main arguments
+* ds: DynamicalSystem; so far only Continuous implemented
+* feature_extraction::Function : function that receives as input the trajectory, time vector, and optional parameters to return a matrix containing the features extracted from y, with the j-th feature in the j-th column of the matrix.
+* ic_grid::AbstractMatrix : matrix containing the grid of initial conditions; Each column contains one IC; so it is of size dimensionality x N
+* clust_mod = "unsupervised" : clustering method chosen, can be "supervised" or "unsupervised". If "supervised", user needs to input the templates features also
```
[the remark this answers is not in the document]
KalelR: Okay, sorry. I am doing that in the updated documentation.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,242 @@
...
+
+
+## Keywork arguments
+* T = 100 : total time for evolving initial conditions (after transient)
+* Ttr = 0 : transient time to evolve initial conditions
+* Δt = 1 : Integration time step, Δt = 1/fs, fs being the sampling frequency used in the bSTAB paper
+* y0 = []: initial conditions for the templates in the "supervised case"
+* templates_labels : (for supervised) labels for the templates
```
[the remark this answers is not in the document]
KalelR: For the supervised classification method. I will explain everything better in the updated documentation, sorry about the confusion.

src/basins/basins_fractions.jl
KalelR: > Is this still true?

For this first version, which I thought would just be the implementation of the author's algorithm, the user needs to provide the features to be extracted from a given trajectory. The lyapunov spectrum thus would not work right now, but it certainly could be a nice feature.

 > What's your plan for default featurizers? How do you think its API should be? Can you provide an implementation plan for this?

I'm not sure still. In the simple examples provided in the paper, the features are quite simple, relying on simple statistics of the trajectory, like standard deviation, average, maximum or minimum. But this can depend a lot on the specific system of course, and may be much more complicated and case-dependent in more complicated cases, with more dimensions, for instance. 

To have default featurizers, we would need a reliable method to determine at least some important, characterizing, features from a given generic system, right? Maybe a brute-force attempt, just trying lots of quantifiers, and seeing which ones matter for distinguishing the trajectories, could work, but is certainly not guaranteed. I'm not sure this has a simple solution. For instance, identifying important features in a given time-series, distinguishing between chaotic attractors in general are not solved problems, I think.  

What do you think? I'd be glad to work on developing this. It would certainly be nice to have a method for this.

src/basins/basins_fractions.jl
KalelR: Ok, bad habit, sorry. Thanks for the advice!

src/basins/basins_fractions.jl
KalelR: An old habit from C, I think. It is indeed valid, though unnecessary as you say. I will remove them.

src/basins/basins_fractions.jl
KalelR: You are right, it is indeed a vector. I will correct the documentation.

src/basins/basins_fractions.jl
KalelR: One `featurizer` receives an initial condition, evolves it and and returns its extracted features in a vector. Another `featurizer` function receives all the initial conditions, and applies the first `featurizer` on all of them, returning a matrix. I have improved their documentation, explaining this. Should I name them differently?

src/basins/basins_fractions.jl
KalelR: It should be the metric used in the clustering algorithm for computing the distances. I may have changed the name in the code, and forgot the documentation. In the new version I explain this better.

src/basins/basins_fractions.jl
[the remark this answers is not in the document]
Datseris: You have nothing to be apologizing for! My comments are directly simply because it makes them clearer and saves time. Don't take anything the wrong way, this is awesome haha!

KalelR:
Hi, thanks for all the comments! I added a lot more info to the documentation, hopefully it is more understandable now. The description of the main function now reads>
> Let ``F(A)`` be the fraction of initial conditions in a region of state space
``\\mathcal{S}``, given by `ic_grid`, which are in the basin of attraction of an attractor
``A``. `basin_fractions` estimates ``F`` for attractors in ``\\mathcal{S}`` by counting
which initial conditions end up in which attractors. To do this, it evolves each trajectory
`u` for `T` times past an initial transient `Ttr`, values which have to be large enough to
guarantee the trajectory ends up in its steady state. This steady state trajectory `u` is
then transformed in a vector of features, extracted using the user-defined
`feature_extraction` function. Each feature is a number useful in characterizing the
trajectory and distinguishing it from trajectories in other attrators. For instance, a
useful feature distinguishing a stable node from a stable limit cycle is the standard
deviation of `u` (zero for the node, nonzero for the limit cycle). The vectors of features
are then used to identify to which attractor each trajectory belongs (i.e. in which basin of
attractor each initial condition is in). The algorithm presents two methods to do this. In
the supervised method, the attractors are known to the user, which provides one initial
condition for each attractor in the region. The algorithm then evolves these initial
conditions, extracts their features, and uses them as templates representing the attrators.
Each trajectory is considered to belong to the nearest template, which it finds using the
first-neighbor clustering algorithm.

> If the attractors are not as well-known, the alternative, unsupervised method, can be used.
It maps the vectors of features to an attractor by analysing how the features are clustered
in the feature space. Using the DBSCAN algorithm, it identifies these clusters of features,
and considers each cluster to represent an attractor. Features whose attractor is not
identified are labeled as `-1`. Otherwise, they are labeled starting from `1` in crescent
order.

> These labels are then returned by the algorithm, along with the fraction ``F(A)`` for each
label (attractor). The sampling error associated with this method is given by[^Stender2021]
``e = \\sqrt{F(A)(1-F(A))/N}``, with ``N`` denoting the number of initial conditions, if the
uniform random sampling is used in `ic_grid`. 

I've also changed the code in some parts to make them it more readable.  It basically involves two functions, `featurizer`, to integrate the initial conditions and extract their features, and `classify_solution`, which identifies their corresponding attractors using the clustering algorithm. There are two clustering methods, and so there are two methods for `classify_function`.

I've left at the bottom of the main file some TODO items. I believe they can be implemented after this version is perfected, but please let me know if you think otherwise!

KalelR:
I've made two sets of tests, replicating the first two examples in the paper. Both the global stability values and the basins of attraction are very similar in both cases, and also in both clustering methods. The stability values are checked in 'stability_basins_tests.jl`. 

image
image

The change in colors for the second case happens because labeling is done differently in each method: it is sorted by size in the unsupervised case, and labeled by the index of the template in the `ic_templates` for the supervised method.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,278 @@
...
+end
+
+# `sampler` is a function without any arguments that generates random initial conditions
+# within a subset of the state space. You can use [`boxregion`](@ref) to create a `sampler`
+# on a state space box. #TODO: implement sampler as more efficient alternative to ic_grid.
+"""
+function basin_fractions(ds::DynamicalSystem, feature_extraction::Function,
+    ic_grid::AbstractMatrix, ic_templates::AbstractMatrix=[]; kwargs...)
+
+Compute the global stability of basins of attraction [^Menck2013] of the given dynamical
```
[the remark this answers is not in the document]
Datseris: We need to be precise. The function calculates the fraction of the basins of attraction via random sampling. That's the end of the story and is precisely what we are doing.

To my understanding it does not make sense to say that this is the "basin stability". In my eyes the basin stability is how the fraction changes with a parameter change.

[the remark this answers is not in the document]
KalelR: > We need to be precise. The function calculates the fraction of the basins of attraction via random sampling. That's the end of the story and is precisely what we are doing.

That's true.

> To my understanding it does not make sense to say that this is the "basin stability". In my eyes the basin stability is how the fraction changes with a parameter change.

I used the term because it's what the authors also use. But we can certainly leave it out, if you prefer. Basin fractions is indeed very precise.

Datseris:
I've just pushed a change to the documentation string. I'm submitting a review soon as well. But the docstring is the central part of this function as far as the user is concerned. I made improvement based on the following mentality:

- I would always and forever avoid the term "steady state" which has no precise mathematical definition and different fields of science interpret it differently.
- What is a state space fraction of a basin is clear. Well defined. Whether this can be interpreted as "stability" is not clear nor well defined. In contrast to whether a fixed point is stable or not, the concept of a "basin of attraction being stable" is quite vague to me. And it also requires parameter variability. The point is, we really don't need to care. The function calcuates fractions of basins of attraction, that is the truth and the end of the story. If someone wants to take this further, we don't care.
- The docstring was very long. A good rule of thumb I give to contributors is the following: every word, every line in a docstring must have its weight weighted in gold. This means: you must say absolutely everything necessary, but you must do so in the most concise way possible.
- You really, _really_ shouldn't be repeating other documentation strings in new documentation strings. Don't say "to convert a Matrix to a Dataset do ...". If a user wants to know how to make a `Dataset` they should go to its documentation string.
- When you make bulletpoints in docstrings, every row must start with two spaces.

src/basins/sampler.jl
```diff
@@ -0,0 +1,51 @@
...
+
+
+"""
+Convenience sampler function for generating random points inside a region. Returns a
+     generator for the points. The region can be:
+    * a box, with edges `min_bounds` and `max_bounds`. The sampling of the points inside this
+    box is according to `method`, which can be either `"uniform"` (implemented in 
+    `expansionentropy.jl`) or `"multgauss"`.
+    * a sphare, of `spheredims` dimensions, radius `radius` and centered on `center`.
+
```
Datseris, line 37: this docstring has wrong indentation levels and some types (sphare).

Datseris (changes requested):
We are almost there. There is only a couple of things remaining.

1. So from my end, I suggest the following. If the user provides `ics::Function`, we only return the `fs` dictionary. Nothing else. If instead we get `ics::Dataset` then we return `fs, ics, class_labels`. Returning only `class_labels` is useless if the initial conditions are not returned as well.
2. Have a look at threading in Julia base, because I have an idea how we can use it to parallelize the entire thing. https://docs.julialang.org/en/v1/manual/multi-threading/ (I will do this myself probably after you address these comments)
3. There are some "TODO" comments in the file, you should address them and if you can't then summarize where you need help in a message here.
4. I've checked the tests, it looks good.
5. Remove the `seed` keyword. Once again this is something the user can do outside the function in the same amount of user-written code. Instead, I would consider using a `seed` keyword in the `sampler` function.

All in all I'm extremely happy with this PR. I would recommend you to open a PR in DynamicalSystems.jl adding this to the docs while also adding an example highlighting its usage.

Datseris:
@KalelR do you think you can give a try in using https://diffeq.sciml.ai/stable/features/ensemble/ ? It is a bit advanced but not so much. The diffeq solution object is also a vector of vectors (same thing as a dataset),

awage:
Hi, I have  a couple of comments on the docs and also on the code. May I add some comments or do you prefer to resolve all the other comments first?

Datseris:
add them on top of mine, that's better I think.

test/basins/basins_fractions_tests.jl
```diff
@@ -0,0 +1,151 @@
...
+    s = ChaosTools.sampler(min_bounds=min_limits, max_bounds=max_limits, method=sampling_method)
+    Random.seed!(1)
+    ics_foo = s
+    ics = [s() for i=1:N]
+    ics = Dataset(ics)
+
+
+    #---Running for supervised
+    #prepare templates
+    ic_templates = Dataset([0.5 0; 2.7 0]) #each IC along a row
```
awage, line 54: This `ic_templates` is the vector `attractors`. right?

Datseris: yes, will be renamed to `attractors`, or `attractors_ic` is probably even better.

test/basins/basins_fractions_tests.jl
```diff
@@ -0,0 +1,151 @@
...
+    templates_labels = ["IC", "LC"]
```
awage, line 55: It's for plotting purposes?

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,292 @@
...
+stable limit cycle is the standard deviation of a dimension in `X` (zero for the node, 
+nonzero for the limit cycle). The vectors of features are then used to identify to which 
+attractor each trajectory belongs (i.e. in which basin of attractor each initial condition is in).
+The method thus relies on the user having at least some basic idea about what attractors
+to expect, in contrast to [`basins_of_attraction`](@ref).
+
+The algorithm of[^Stender2021] that we use has two methods to do this. 
+In the **supervised method**, the attractors are known to the user, who provides one
+initial condition for each attractor in ``\\mathcal{S}`` using the optional `attractors`
+argument. `attractors` then is a `Dataset` containing the initial conditions.
```
awage, line 105: `attractors` is a `Dataset` whose rows represent a single initial condition for each attractor.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,292 @@
...
+
+
+"""
+`featurizer` receives an initial condition and returns its extracted features in a vector.
+It integrates the initial condition, applies the `feature_extraction` function and returns
+its output. The type of the returned vector depends on `feature_extraction`'s output.
+"""
+function featurizer(ds, u0, feature_extraction; T=100, Ttr=100, Δt=1,
+    extract_params=NamedTuple(), diffeq=NamedTuple(), kwargs...)
+    u = trajectory(ds, T, u0; Ttr=Ttr, Δt=Δt, diffeq) #TODO: maybe starting an integrator
```
awage, line 190: Is it possible to introduce some sort of callback such than the solver stops as soon as the feature are stable? It should improve the performance a lot.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,292 @@
...
+
+The output `fs` is a dictionary whose keys are the labels given to each attractor, and the values
+are their respective fractions. The label `-1` is given to any initial condition whose
+attractor could not be identified. The `class_labels` output is an array of size `N`
+containing the label of each initial condition given in `ics`.
+
+
+## Keyword arguments
+### Integration
+* `T, Ttr, Δt` : Propagated to [`trajectory`](@ref) with `T=100, Ttr=100` as default.
```
awage, line 65: Δt = 1

src/basins/basins_fractions.jl
Datseris: we already had some discussion on this. A problem is that some features need entire trajectories to compute, some others do not. For example, standard deviation of a trajectory. 

@KalelR perhaps we can completely skip saving a trajectory. We can use https://github.com/joshday/OnlineStats.jl for means and standard deviations instead. For Lyapunov exponents we already have such an algorithm. On the downside, I don't see how we can apply this approach to fractal dimensions.

Datseris: if go this approach, we don't use the `EnsembleSimulation` from DiffEq. Instead, we need to make our own struct.

Datseris:
After the initial comments here are addressed, we need to continue in two ways:
- Continue on this branch and use the `EnsembleProblem` for parallelization.
- Start a new branch and use `integrator` and `OnlineStats`. In this case the the trajectory is _not_ saved in memory. This can also efficiently compute the Lyapunov exponents.

I must admit, I am not sure if this is the way to go. It is certainly more complicated. But we should test and benchmark both approaches nevertheless. I will do one of the two branches, so @KalelR pick whatever you feel more comfortable and do it, and I'll do the other option. I believe you should go with `EnsembleProblem`, as in that interface you already have the trajectory saved, as is currently done, so it will be easier. You need to create a function that maps the diff eq `sol` object to the feature vector.

src/basins/basins_fractions.jl
awage: I see. The problem is that there is no universal feature. For a specific attractor feature you may need the whole trajectory. On the other hand `EnsembleSimulation` is a great tool for bStab.  It's a tough design decision (I leave it to you both!!). 

I think there is a lot of possible research to do on automatic feature extraction. But that's for later.

Datseris: There are definitly pros and cons. For example, in the second suggestion with OnlineStats, we can actually implement a termination protocol so that users don't have to guess for how long to evolve trajectories for. It is evolved as long as necessary until the measures converge. This is not possible with `EnsembleProblem` which needs a predefined amount of type to run and save trajectory. 

We will see! Once Kale has addressed existing comments, we decide on a system to benchmark. Let's use a multistable one from our basins paper. Then we implement the `EnsembleProblem` method, and I do a second PR for the other method and see how it goes.

We might even decide to keep both methods in the end, who knows...

KalelR:
Hey, that's awesome! I will address all your comments then start working on the `EnsembleProblem` interface as soon as I have time (I'm quite busy with my PhD for the next few days).

src/basins/basins_fractions.jl
SebastianM-C: Sorry for the intervention, but I think that `EnsembleProblem` doesn't necessarily impose a fixed time. You can use an infinite `tspan` and terminate the integration when desired. See https://diffeq.sciml.ai/stable/features/callback_functions/#Example-2:-Terminating-an-Integration

Datseris: Correct. It adds yet another callback however. We will now have three callbacks total. One for the termination, one for the saving the output and one for the calculation of the feature vector. 

The process becomes more and more convoluted with this `EnsembleProblem` I must admit, so the performance needs to be much, much better than the much clearer `@threads` over integrators, in order for DynamicalSystems.jl to go with it. Let's see...

awage:
Hi! I have checked out this branch and made some comparison of the estimation with the current bSTAB implementation and the `basins_of_attraction` tracking mode (or whateever it will be called). 

There are some interesting figures that I have taken from the test suite: 

```
# Duffing oscillator (5000 samples): 

@btime S, class_labels = basin_fractions(ds, feature_extraction, ics, ic_templates; T=Texec, Ttr=Ttr, Δt=Δt)
17.778 s (200302212 allocations: 5.99 GiB)

@btime for (k,y0) in enumerate(ics)
    bas_stat[k] = ChaosTools.get_color_point!(bsn_nfo, integ, y0)
end
 1.433 s (733198 allocations: 27.21 MiB)

# Forced pendulum: 

# println("Test No1. Supervised") (1000 Samples)
@btime S, class_labels = basin_fractions(ds, feature_extraction, ics, ic_templates; T=Texec, Ttr=Ttr, Δt=Δt)
10.745 s (205122781 allocations: 6.05 GiB)

@btime for (k,y0) in enumerate(ics)
    bas_stat[k] = ChaosTools.get_color_point!(bsn_nfo, integ, y0)
end
154.337 ms (207515 allocations: 7.62 MiB)

```
The basins fractions are similar in both implementation but the numbering of attractors is different. 

There is a lot of allocation for bSTAB and some profiling is needed. Tracking the attractors is faster in both cases but I think bSTAB can do much better than the actual implementation. If you want the full code I can push it in this branch but it will duplicate #237. My guess is that the implementation with `basins_of_attraction` is a lower bound for bSTAB. 

The good point about the tracking method is that you forget about the features and you automatically get the attractors. But on the other hand you may have to choose wisely the grid. Another good thing is that the interface of `basins_of_attraction` takes care of everything.

Datseris:
It is important to clarify the following: the bSTAB method does not need an _exact_ specification/finding of the attractors, which is why it uses the clustering method. So the attractors do not have to be found exactly; the clustering algorithm takes care of grouping the feature vectors into "cloud points that correspond to attractors".

The current PR is certainly far from optimized, but that is known, we are waiting for Kale on this front (no pressure, no worries). This approach for bSTAB also is the only one that is parallelizable, which is worth considering, while our tracking method is not parallelizable I believe (?).

> If you want the full code I can push it in this branch but it will duplicate #237. 

No let's please not do that yet. The bSTAB method has clear design based on the paper that defines it. We should keep it that way. Anything that builds upon it should be certainly in a separate PR, and certainly after the "original" bSTAB method is implemented. This makes for a clear comparison and makes potential improvements clear to see if they exist.

I'll be having a closer look in #237 shortly.

awage:
Thanks for the clarification. I am intending to modify the bSTAB implementation but rather to propose an extension. I find the method very interesting but there is a blind spot: the definition of the feature. It is not obvious to me how to find the adequate feature and the automatic identification of attractors maybe solve this problem. 

Regarding parallelization: I have to think seriously about this. 

@KalelR : No rush! I was just exited to share these results.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,300 @@
...
+  bSTAB paper
+* `diffeq = NamedTuple()` : other parameters for the solvers of DiffEqs
+* `num_samples` : Number of sample initial conditions to generate. Used only if `ics` is a
+`Function`.
+* `seed_sampling = 1` : The seed for the random number generator used in generating the initial
+conditions. Used only if `ics` is a `Function`.
+### Feature extraction and classification
+*`clust_method_norm=Euclidean()` : metric to be used in the clustering in the supervised
+    method; 
+* `extract_params = NamedTuple()`  : optional parameters for the `feature_extraction` function.
```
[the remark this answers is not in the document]
KalelR: But what if the user needs to use `basin_fractions` in a loop? It could be that the featurizer relies on some parameters that would change for each iteration. I thought that then passing `extract_params`would be a clean to allow this.

src/basins/basins_fractions.jl
```diff
@@ -0,0 +1,292 @@
...
+j-th feature. To do this, it  calls the other `featurizer` method, made for just one array of ICs.
+`ics` should contain each initial condition along its rows.
+"""
+function featurizer_allics(ds, ics::Dataset, feature_extraction::Function;  kwargs...)
+    num_samples = size(ics, 1) #number of actual ICs
+    feature_array = [Float64[] for i=1:num_samples]
+    for i = 1:num_samples #TODO: implement parallelization, if necessary
+        ic, _ = iterate(ics, i)
+        feature_array[i] = featurizer(ds, ic, feature_extraction; kwargs...)
+    end
```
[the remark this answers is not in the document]
KalelR: I wrote the loop in full form, instead of the single line, to allow for later paralelization through `@Threads` if we wanted. Given the other plans for parallelization, this may not be necessary any longer and I can do this change if you prefer.

test/basins/basins_fractions_tests.jl
KalelR: Yup!

test/basins/basins_fractions_tests.jl
KalelR: It was, not really needed anymore. I removed it.

KalelR:
> So from my end, I suggest the following. If the user provides ics::Function, we only return the fs dictionary. Nothing else. If instead we get ics::Dataset then we return fs, ics, class_labels. Returning only class_labels is useless if the initial conditions are not returned as well.

Good point. What is the best way to do this? Should I define two `basin_fractions`, one with each return type?

Also, why return `ics` if it was the user who passed it?

Datseris:
yeah you're right no reason to return `ics` in this case.

KalelR:
Hey guys. I'll start looking at `EnsembleProblem` now. Only two small issues are still open from all the initial comments: 
1. Whether the `feature_extraction` function should receive `extract_params`. I think this could be useful for calling `basin_fractions` in a loop. What do you think @Datseris?
2. Shortening `featurizer_allics` loops to a single line. I think we can do this once we decide `@Threads` definitely won't be used.

Datseris:
What are `extract_params` ?

src/basins/basins_fractions.jl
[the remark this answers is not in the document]
KalelR: @Datseris `extract_params` is here.

Datseris:
@KalelR no extract parameters please. These can be wrapped into a closure. `f(u) = f(u, extract_params)`.

Datseris:
@KalelR please also prepare a simple benchmarking script that uses two systems, one is the magnetic pendulum and the other the Lorenz84 in the parameter regime with coexistinence of three attractors. this happens for parameters `F = 6.886, G = 1.347, a = 0.255, b = 4.0`

KalelR:
> @KalelR please also prepare a simple benchmarking script that uses two systems, one is the magnetic pendulum and the other the Lorenz84 in the parameter regime with coexistinence of three attractors. this happens for parameters `F = 6.886, G = 1.347, a = 0.255, b = 4.0`
> 
> EDIT: I will use that script to measure my parallel threaded integrator approach.

Done. I've added the script to the basins test directory. Please let me know if there is better place to put it in.

KalelR:
Hey, I just pushed a benchmark file comparing integration + feature extraction using two parallelization methods: `@Threads` and `EnsembleProblem`. The results were:
### 25 workers, N=50000, T = 2000
- Threads:    23.125 s (1077057193 allocations: 31.03 GiB)
- Ensemble (Distributed):     106.806 s (1176707851 allocations: 44.56 GiB)
- Emseble (Threads):   33.503 s (1242096027 allocations: 41.51 GiB) [It prints a long error message saying "concurrency violation detected". But also outputs the feature values similar to the other methods. Don't know what is happening.]

### 25 workers, N=1000, T=2000
- Threads:   256.846 ms (21540932 allocations: 635.44 MiB)
- Ensemble (Distributed):   2.179 s (23598717 allocations: 916.71 MiB)
- Ensemble (Threads):   343.171 ms (24875192 allocations: 852.93 MiB

I have no clue why the `EnsembleThreads()` method errored for N=50000. But even if we trust its result (which is similar to the other results), `EnsembleProblem` is still losing to `@Threads`. 

It could very well be that my implementation is bad. If someone could please take a look to see if it can be improved, then we can trust these benchmarks better, and I could also do some other ones.

Datseris:
I will be taking a long look, don't worry.

Datseris:
@KalelR please help me by telling me exactly which file I should be analyzing, `basin_fractions_benchmark.jl` or `basin_fractions_featurizer_benchmark.jl`.

Datseris:
@KalelR this is very weird:
```julia
    @Threads.threads for i = 1:num_samples 
        ic, _ = iterate(ics, i)
```
why didn't you write just `ic = ics[i]`?

Datseris:
@SebastianM-C the timings from @KalelR show that looping over our own `trajectory` with `@threads` is faster than using `EnsembleProblem` with threading from DiffEq. So that already kind of settles the discussion of whether to go with that approach. I am now testing my own idea of not even bothering to calculate a solution, but rather pre-initialize as many integrators as threads.

Datseris:
I've added another version. It pre-initializes integrators, re-inits them over a `@threads` loop. Oddly, it is as fast as the version that loops over trajectory. A bit slower actually. No real clue why, because we are saving creating an integrator from scratch. This is for `N=1000`. Increasing to `2000` makes the difference even worse.

---

I'll now work on the version that _does not save the trajectory_ and instead uses OnlineStats.jl.

Datseris:
Ah, alas. I am saddened. The OnlineStats.jl version is _slower_. So I guess, the OnlineStats.jl would only win if we wanted to compute trajectories so large, they wouldn't fit in memory. This of course doesn't make sense in our application, we can instead increase `Ttr`, which doesn't increase memory consumption. 

Timings:

```julia
  899.035 ms (33630634 allocations: 791.36 MiB) # Threading over `trajectory` calls
  1.433 s (32027366 allocations: 1.67 GiB)       # Threading over integrators and OnlineStats.jl
```
Josh Day, creator of onlinestats.jl, also confirmed in a question that typically onlinestats.jl will be slower by about an order of magnitude.

Datseris:
So, I guess @KalelR you should go ahead and implement the version based on threading `trajectory` on the actual source files, polish as many things as you can and tag me for a final review!

Datseris:
oh, and we should rename this function. `basin_fractions_clustering` is a more appropriate name, as in #237 we are implementing an entirely new method to calculate basin fractions.

KalelR:
> Ah, alas. I am saddened. The OnlineStats.jl version is _slower_. So I guess, the OnlineStats.jl would only win if we wanted to compute trajectories so large, they wouldn't fit in memory. 

Or in situations where some ics have very large transients and others don't, I guess. Terminating once the asymptotic state is reached could still be faster in more extreme scenarios. But that's not really the focus here.

KalelR:
> Well, actually, one of the major performance bottlenecks of the method is that you do not know when to terminate the time evolution. If we can add convergence criteria to the means and lyapunov exponents, then being able to terminate in advance is super advantageous. I'm not sure how to write a convergence criterion for the `Mean` however.

Yeah. That could be something to work on in the future.

Datseris:
> Yeah. That could be something to work on in the future.

It is a breaking change to change the interface like that. So a bit more careful consideration must be done.

test/basins/tipping_points_tests.jl
```diff
@@ -11,10 +11,10 @@ using Test
     basins_after, = basins_of_attraction((xg, yg), ds; Δt = 1., idxs = 1:2,
     diffeq=(;:reltol => 1e-9), show_progress = false)
     @testset "Basin fractions" begin
-        fs = basin_fractions(basins_before)
+        fs = basin_fractions_clustering(basins_before)
```
Datseris, line 14: What's going on here is certainly not correct: you are supposed to rename only your own method, not the previous version of `basin_fractions` that takes as input an array

KalelR: Oh, that was stupid. My bad.

Datseris:
@KalelR this is very nice. Thank you so much for making such a huge PR. Normally I'd merge this and tag a release. I just want to hold it back for a moment though. I am thinking whether I can still add the convergence criterion that terminates iteration on demand instead of saving the trajectory. But actually, you know what, it's okay, you are right, we can do this later. 

Would you please open a documentation PR at DynamicalSystems.jl docs implementing a really basic application of this function and also stating its docstring? You can obviously use what we have in the test suite already, or whatever else you fancy.

KalelR:
> @KalelR this is very nice. Thank you so much for making such a huge PR. 

Haha yeah, it took much longer than what I anticipated. Thank _you_ for everything! And sorry about my Julia skills haha, I'll study a bit more before taking another project.

> Normally I'd merge this and tag a release. I just want to hold it back for a moment though. I am thinking whether I can still add the convergence criterion that terminates iteration on demand instead of saving the trajectory. But actually, you know what, it's okay, you are right, we can do this later.

Up to you. There are also other improvements that could be made to this algorithm, like in the clustering part. And also even research around this, such as automatic feature detection. But the algorithm is already good enough to be released I think.

> Would you please open a documentation PR at DynamicalSystems.jl docs implementing a really basic application of this function and also stating its docstring? You can obviously use what we have in the test suite already, or whatever else you fancy.

Sure! This was the last thing on my list.

awage:
Very good job!! I will try your code when the final version is ready.

Datseris:
@KalelR I've changed the API a bit, improved the sampler function , which is now exported and will be listed in the documentation as its own thing. Please do have a go through my commits, I've made the docstring a bit smaller and made the naming of the source code a bit better, as I intend to add a second version using OnlineStats.jl. When you see these changes, please update the documentation Pr at DynamicalSystems.jl accordingly!

Datseris:
@KalelR I am a bit skeptical... the duffing oscillator has 5 attractors for these parameter values? Have you actually confirmed that? have you seen / plotted these attractors? seems too high of a number of attractors for this system...

Datseris (approved):

KalelR:
> @KalelR I am a bit skeptical... the duffing oscillator has 5 attractors for these parameter values? Have you actually confirmed that? have you seen / plotted these attractors? seems too high of a number of attractors for this system...

As I remember, yes. This is the second example given in the bSTAB paper. There, they plot trajectories for each attractor, showing they are indeed 5. I replicated their figures, so I have seen them by myself also. 

image

PR #221, opened 2021-10-03
Add `fixedpoints` function based on IntervalRootFinding.jl

Datseris:
See docstring!


Datseris:
@dpsanders this is good to go from my end. Perhaps you are interested in giving a review? Something I found odd is that there is always a warning of Non-unique fixed points. E.g, for the Lorenz test case, even though the dynamic rule is very smooth and everything is in order, `root.status != :unique` is not true. Any ideas why this is the case...? I'd expect that in such a system where the fixed points are very much separated in state space and the vector field is smooth there shouldn't be any problems.

Datseris:
@rveltz I am also tagging you here, just to let you know about this function that seems to be directly relevant for BifurcationKit.jl, at least for starting continuations for low dimensional systems.

PR #222, opened 2021-10-03
Fix and complement DyCA tests.

KalelR:
Complement tests are on the EEG seizure data. Also added the option to normalize the eigenvectors in DyCA. This way, the algorithm matches with the Python version.

### DIFF
--- a/src/dimreduction/dyca.jl
+++ b/src/dimreduction/dyca.jl
@@ -33,6 +33,8 @@ used for dimensionality reduction.
 
 Return the eigenvalues, projection matrix, and reduced-dimension data
 (which are just `data*proj_mat`).
+## Keyword Arguments
+* norm_eigenvectors=false : if true, normalize the eigenvectors
 
 ## Description
 Dynamical Component Analysis (DyCA) is a method to detect projection vectors to reduce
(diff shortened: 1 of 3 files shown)


src/dimreduction/dyca.jl
```diff
@@ -33,6 +33,8 @@ used for dimensionality reduction.
 
 Return the eigenvalues, projection matrix, and reduced-dimension data
 (which are just `data*proj_mat`).
+## Keyword Arguments
+* norm_eigenvectors=false : if true, normalize the eigenvectors
```
Datseris, line 37: Which version does the Python implementation have? We should have the same for consistency (true/false)

KalelR: On my machine, I used Python 3.8.10, but I don't know for the notebook. The only difference is that I had to pass `np.array(eeg_data)` to `DyCA()`, instead of just `eeg_data` (Pandas DataFrame).

Datseris: oh i was asking specifically for the norm-eigen vectors part

KalelR: Aahh okay, now I got it. Sorry hahaha. The python `eig()` function returns the normalized eigenvectors (https://numpy.org/doc/stable/reference/generated/numpy.linalg.eig.html), so its "default" (it actually only has this option) would be true.  I made the default here false for consistency with previous code, but I'm not sure if it's necessary. I'll change as you prefer.

Also, just a note, for the two algorithms to completely match, the python one would have to sort the eigenvalues/eigenvectors.

Datseris:
I've left the keyword as its default (false) for now. Thanks!

PR #93, opened 2019-10-09
WIP: Implementing the YIN algorithm

anoojpatel:
Initial Commit with basic code structure. Please let me know if there is any styling requirements I need to adhere too, or if my structure needs to be rearranged/relocated!

Closes #92

PR #223, opened 2021-10-04
YIN, a fundamental frequency estimator

KalelR:
Hey, I've finished the code that was already started in PR #93 and tested it. Included here is the whole code and the test I did, comparing plots the measured quantities. It matches perfectly with the available python test, using the audio file they provide.
I commented the code as it is a plot, and am not sure how to automatize. Also not sure if I should upload the audio file, or just provide a link to it.


src/periodicity/yin.jl
```diff
@@ -0,0 +1,167 @@
...
+    for τ in 1:τmax-1
+        for j in 1:(N-τmax)
+            df[τ+1] += (x[j] - x[j + τ]) ^ 2
+        end
+    end
+    return df
+end
+
+"""
+getPitch(
```
Datseris, line 40: `getPitch` is inappropriate name. it would be `get_pitch` in Julia syntax.

src/periodicity/yin.jl
```diff
@@ -0,0 +1,167 @@
...
+    cumulative_mean_normalized_difference_functionf, τ_min, τ_max, harmonic_threshold=0.1
+)
+
+Return the fundamental period of a frame,
+based on the cumulative mean normalized difference function (CMNDF).
```
Datseris, line 45: This is confusing. The function name says "get pitch", yet here you say "return the fundamental period of time". Pitch and time are inverse of each other. The function name should be improved so that its purpose becomes clear from its name.

src/periodicity/yin.jl
```diff
@@ -0,0 +1,167 @@
...
+        w_len = 512,
+        w_step = 256,
+        f0_min = 100,
+        f0_max = 500,
+        harmonic_threshold = 0.1,
+        diffference_function = differenceFunction_original
+    )
+
+Computes the Yin Algorithm.
+Returns fundamental frequency and harmonic rate.
```
Datseris, line 102: You need to have a `## Description` of what this algorithm is, and which also cites the paper.

src/periodicity/yin.jl
```diff
@@ -0,0 +1,167 @@
...
+"""
+    yin(
+        sig, sr;
+        w_len = 512,
+        w_step = 256,
+        f0_min = 100,
+        f0_max = 500,
+        harmonic_threshold = 0.1,
+        diffference_function = differenceFunction_original
+    )
```
Datseris, line 99: Please change this to `yin(sig, sr; kwargs...)`. Then explain position arguments directly after the call signature.

Datseris:
@KalelR I've left some comments.

1. Can you also incorporate `yin` as an option in the `estimate_period` function? After all, the inverse of the fundamental frequency is the period.
2. For the tests: what you should do is convert the wave file into a `.csv`. Then, upload this here via a pull request: https://github.com/JuliaDynamics/JuliaDynamics/tree/master/timeseries  It is a folder that we upload data necessary to test things so that we do not burden the code repo.

src/periodicity/yin.jl
KalelR: Done.

src/periodicity/yin.jl
KalelR: Yup, you are right. I had actually just naively translated from the python algorithm that was available. I have now update the code and changed this function significantly. The new name makes a lot more sense.

src/periodicity/yin.jl
KalelR: Sorry, but then how would I set the optional values for these parameters?

KalelR:
Reading the paper more carefully, I started to have some problems with the Python implementation that I had based the code on. So I rewrote a lot of the code, comparing closely with another Python implementation, done in the [Librosa library](https://librosa.org/doc/main/_modules/librosa/core/pitch.html#yin), which makes a lot more sense. I also compared the code to several others I found on the web, including the Matlab implementation from the authors of YIN. 
There are some minor differences between codes, which I don't think interfere with the end result too much. 

I've done some tests comparing the result from this code with the result from Librosa, and they all fit very well. Two example tests are done over chirps, which are artificially generated signals from 440 to 880 Hz. The results are:

librosa-chaostools-exponentialchirp
librosa-chaostools-linearchirp

KalelR:
> 2. For the tests: what you should do is convert the wave file into a `.csv`. Then, upload this here via a pull request: https://github.com/JuliaDynamics/JuliaDynamics/tree/master/timeseries  It is a folder that we upload data necessary to test things so that we do not burden the code repo.

This file's size is 2MB. It could actually be generated using SignalAnalysis.jl instead. What would be better?

Also, how would I include it in the test?

Datseris:
> Also, how would I include it in the test?

You download it with Base.download! 

> It could actually be generated using SignalAnalysis.jl instead. What would be better?

This would increase the test time more than downloading I think...? Depends on how much time install+precompile+data_production takes with this package.

src/period_return/period.jl
```diff
@@ -37,6 +37,9 @@ These methods are faster, but some are error-prone.
   default setting, this method will fail if there less than 10 periods in the
   signal). The keyword `ϵ = 0.2` (`\\epsilon`) means that `1-ϵ` counts as "1" for the AC.
 
+* `:yin` : The YIN algorithm. An autocorrelation-based method to estimate the fundamental
+period of the signal.
+
```
Datseris, line 42: can we add a reference to the paper here?

Datseris: (or a crossreference to another function that references the paper)

src/period_return/yin.jl
```diff
@@ -0,0 +1,176 @@
+export yin
+
+"""
+    yin(sig::Vector, sr::Int; w_len::Int = 512, f_step::Int = 256, f0_min = 100, f0_max = 500,
+        harmonic_threshold = 0.1, 
+        difference_function::Function = difference_function_original, kwargs...) -> F0s, frame_times
+
+Estimates the fundamental frequency (F0) of the signal `sig` using the YIN algorithm [^1].
```
Datseris, line 8: Notice that [^1] is no longer the way to add references. Do something like [^AuthorNameDate] and put this also at the end of the docstring.

src/period_return/period.jl
KalelR: Done.

src/period_return/yin.jl
KalelR: I changed it to [^CheveigneYIN2002], but I'm not sure what you mean by putting at the end of the docstring. It's already there in `yin.jl`.

KalelR:
> > Also, how would I include it in the test?
> 
> You download it with Base.download!
> 
> > It could actually be generated using SignalAnalysis.jl instead. What would be better?
> 
> This would increase the test time more than downloading I think...? Depends on how much time install+precompile+data_production takes with this package.

Ah ok, I'll update the test with this when the file is merged into JuliaDynamics :)

KalelR:
I think the test is failing because `DelimitedFiles` is not being found. Should I add it to this package?

test/yin_tests.jl
```diff
@@ -0,0 +1,36 @@
...
+# sig = convert(Array{Int64,1}, sig)
+
+
+# pitches, harmonic_rates, argmins, times = yin(sig, sr; w_len=1024, w_step=256, f0_min=70, f0_max=200, harmonic_threshold=0.85)
+
+# #Plot results, comparing with python's version. If we use the differenceFunction_original there, the results match perfectly!
+# harmonic_threshold=0.85
+# duration = length(sig)/float(sr)
+
+# using PyPlot
```
Datseris, line 17: this file is fully commented.

Datseris:
tests fail

test/yin_tests.jl
KalelR: My bad, shouldn't be there

Datseris:
alright we need a minor version bump in project.toml. Can you open a PR in DynamicalSystems.jl that adds this to the docs as well? So that I we don't forge to docmeunt that this exists. Just adding `yin` to the same `@docs` block that has `estimate_period` is enough.

KalelR:
Oh, I forgot to branch on DynamicalSystems.jl and pushed directly to master. I'm sorry (didn't even know I could do that! :s).
Should I redo it? Sorry again

Datseris:
Unfortunately I do not know how to limit access to only non-master branches (or even if it is possible), so yes you can do that. If you have push access you have it for all branches. Nothing we can do about this now other than verting the change and force pushing to master again. But it's okay I saw the commit and is what I'd merge anyways. Just be careful next time for something more substantial.

KalelR:
> Just be careful next time for something more substantial.

I will, no more pushing again right before lunch!

Datseris:
I'm a different person just before lunch, so I can understand.

PR #226, opened 2021-10-28
Update on lyapunov convergence

KalelR:
Updated  `lyapunovspectrum_convergence` for continuous systems with the `_buffered_qr` more performant version. A test for an in-place Kuramoto network of 300 units with random connectivity showed an improvement from 55 to 34 seconds, with the same result.

Allowed also passing a `DynamicalSystem` instead of just passing the integrator. 

Added `lyapunovspectrum_convergence` for discrete systems and a test for it using the logistic map.


Datseris:
thanks. Please increment patch version nuimber and I'll merge.

PR #228, opened 2021-11-09
Fixed mistake in return of estimate_period with YIN.

KalelR:
Period T0 should be 1/f0, not sr/f0. Now the estimated period for the [example](https://juliadynamics.github.io/DynamicalSystems.jl/dev/chaos/periodicity/#ChaosTools.estimate_period) using a FitzHugh-Nagumo oscillator is the correct value of around 91.07.


Datseris:
thanks, increment patch number!

PR #237, opened 2022-01-04
Add a new mode for basins_of_attraction 

awage:
This PR add a new mode to compute the basins of attraction. 

The idea is to get the structure `BasinsInfo` and to identify individual initial conditions. The algorithm can still identify new attractors and the structure holds all the important information. The basins are stored into a `SparseArray` since we may not need to compute all the basins. Changes are minimals and do not break anything. However there is maybe an issue with an AbstractArray declaration (I pointed this out in the code). 

The function `basins_of_attraction` takes care of all the initialization of the system. The usage is very simple: 

```
ds = Systems.henon_iip(zeros(2); a = 1.4, b = 0.3)
xg = yg = range(-2.,2.,length = 70)
grid = (xg,yg)
bsn_nfo, integ = basins_of_attraction(grid, ds; tracking_mode = true)
att = ChaosTools.get_color_point!(bsn_nfo, integ, [xg[1], yg[1]])
```

This is not intended for end-users but rather it to identify attractors in #219. This is experimental and we should test if it brings any benefit to the bStab method.  @KalelR, I will check out the branch and see how it can work with #219.


src/basins/basins_lowlevel.jl
```diff
@@ -1,8 +1,10 @@
 import ProgressMeter
 using Statistics: mean
+using SparseArrays
 
+# WARNING: THE TYPE HAVE CHANGED TO ABSTRACT ARRAY TO ALLOW Sparse Arrays
 mutable struct BasinInfo{B, IF, RF, UF, D, T, Q, K}
-    basin::Array{Int16, B}
+    basin::AbstractArray{Int16, B}
```
awage, line 7: I understand that it is bad for performances but I don't know how to resolve this without declaring another structure. It will be duplicating some code.

Datseris:
```
mutable struct BasinInfo{B, BA, IF, RF, UF, D, T, Q, K}
    basin::BA{Int16, B}
```
is the answer.

Datseris:
@awage I don't understand this PR yet... can you please give a high level description of what you are trying to do, like what changes from e.g., what we say in the paper? And how does the code work?

Here are some further comments and/or questions:

---

> The idea is to get the structure `BasinsInfo` and to identify individual initial conditions. 

So we still have an underlying grid to identify attractors via recurrences (Section II-A in our paper), and after we identify attractors we can feed any initial condition through the main loop of our algorithm. Yes, that sounds totally fine, and great.

---

I am assuming that the function `att = ChaosTools.get_color_point!(bsn_nfo, integ, [xg[1], yg[1]])` gets the index of the attractor of the initia lcondition `[xg, yg]`.

---

Why do we need SparseArrays, and how are they used? You are aware that Julia only has support for sparse matrices, not higher dimensional arrays, right?

---

The naming system of the current code really makes it difficult to read and understand if one only knows our paper. We should really be sticking to the same naming conventions we use in our paper, throughout the code. This means for example that:

- We shouldn't be using words such as "color" or "draw", as we are no longer thinking in terms of drawing a basin as the initial work of Yorke. In the paper we talk about the "label" of an attractor, and we "estimate" the basins (effortlessly as well ;) )
- Tracking is a name we don't use in the paper. Is it a completely different concept?

---

I see in your code example that you already call the full `basins_of_attraction` function with a "tracking mode". So I guess this "tracking mode" prepares some kind of structure

---

A totally different operation of a function that is decided by a keyword argument is not good API design I think. Whatever we do in this PR, it should be a new function name. Its goal its apparently not to calculate the basis of attraction of a system, but rather to prepare some kind of "structure" that maps initial conditions to attractors, right?

---

Final question: So I guess the benefit of all this approach is that you can use a much sparser grid to create this "tracking mode", which doesn't have as heavy dimensional limitations as the normally dense grid. But, I'm not sure if this really saves a lot. Because even in 10 dimensions, 21 points along each dimension are still way too much, yet it is barely dense enough to separate attractors in state space (or maybe it isn't I actually don't know).

awage:
First the high level description: 

The functionality that I propose to include in the library is to match a single initial condition with an attractor. The mode of operation is identical to the description of the paper, there is no change here. The difference is that the function gives back to the user the structure `BasinInfo` and the integrator object. With these two object you can call the low level functions of `basins_of_attraction` that identify and match the attractors. 

I have several ideas for the use of this mode: 
* You can evaluate initial conditions at arbitrary points (the grid for ic's and the grid for identification are now separate things).  
* Basins fraction estimation based on a sampling of the basins (maybe random walks are worth trying). 
* Basin entropy estimation with random sampling
* Identification of stable and unstable manifolds of saddles.

I had to change some details of the algorithm such as the numbering of unidentified cells from 1 to 0. It does not change anything else to the code and it has no impact on the results. The reason for this change is the possible use of a sparse array that are initialized with zeros instead of ones. **BUT**, I was not aware that `SparseArrays` only support matrices. It is a big bummer for me, I had hopes to be able to sample high dimensional state space. In this case there is absolutely no need for a sparse matrix and this part should be rolled back. 

The code works in the same way as `basins_of_attraction` does. But the user has to iterate the initial condition "manually". 

I will answer the other comments later in the day.

awage:
> So we still have an underlying grid to identify attractors via recurrences (Section II-A in our paper), and after we identify attractors we can feed any initial condition through the main loop of our algorithm. Yes, that sounds totally fine.
>

This is the main idea of this functionality. 


> I am assuming that the function `att = ChaosTools.get_color_point!(bsn_nfo, integ, [xg[1], yg[1]])` gets the index of the attractor from the initial condition `[xg, yg]`.
> 

This is the index of the basin or attractor if the initial condition is in a attractor cell.  


> Why do we need SparseArrays, and how are they used? You are aware that Julia only has support for sparse matrices, not higher dimensional arrays, right?
> 

I answered earlier. It seems that SparseArrays are not efficient in high dimension and use of Dictionary of keys is an alternative. There are some package that aim at this functionallity (like  [SimpleSparseArrays](https://github.com/jw3126/SimpleSparseArrays.jl)) but it is not in the standard library and rather experimental. I will leave this part out and leave it a side project. bSTAB seems a better tool for high dims systems.  

> The naming system of the current code really makes it difficult to read and understand if one only knows our paper. We should really be sticking to the same naming conventions we use in our paper, throughout the code. This means for example that:
>     * We shouldn't be using words such as "color" or "draw", as we are no longer thinking in terms of drawing a basin as the initial work of Yorke. In the paper we talk about the "label" of an attractor, and we "estimate" the basins (effortlessly as well ;) )
> 

Totally agree I will rename functions and variables accordingly. 

>     * Tracking is a name we don't use in the paper. Is it a completely different concept?

This is a bad name. I lacked of imagination the other day.  `ic_labeling` or `ic_identification` fits better. 
 
> I see in your code example that you already call the full `basins_of_attraction` function with a "tracking mode". So I guess this "tracking mode" prepares some kind of structure

The returned structure is the same than `basins_of_attraction`. 

> A totally different operation of a function that is decided by a keyword argument is not good API design I think. Whatever we do in this PR, it should be a new function name. Its goal its apparently not to calculate the basis of attraction of a system, but rather to prepare some kind of "structure" that maps initial conditions to attractors, right?

Totally agree. I will prepare a wrapper around basins of attraction with the same input options. The interface of `basins_of_attraction` is just perfect. 

 
> Final question: So I guess the benefit of all this approach is that you can use a much sparser grid to create this "tracking mode", which doesn't have as heavy dimensional limitations as the normally dense grid. But, I'm not sure if this really saves a lot. Because even in 10 dimensions, 21 points along each dimension are still way too much, yet it is barely dense enough to separate attractors in state space (or maybe it isn't I actually don't know).

I think we can use this functionality to sample basins up to dimension 6 or 7 (I tried it with the Kuramoto oscillator). But as I said in the previous message it will be useful for other purposes.

Datseris:
> the grid for ic's and the grid for identification are now separate things

I've been thinking about it a lot when we were writing the paper. But I have a crucial challenge to this whole idea: **What do we gain by having the grid for basins and grid for attractors different?** Does it _actually_ help anywhere? The limitation of `basins_of_attraction` is memory. If either grid is larger, well, we don't really care about the smaller one. The larger grid is what limits everything.

Yeah, I'm not sure I follow your motivation for these grids being different. What I totally understand is having this separate functionality, which maps an initial condition to an attractor. But in the end, it uses 1 grid, no matter what this grid is, as in the same grid both the attractors and their basins are stored. If we have any basin points, we can use them even in random sampling. A trajectory that hits 60 basin cells in a row can still be mapped to the attractor as with anything else. So all in all, you would just have a coarse grid, but I still see no reason for this grid to be separate for basins and for attractors.

---

> Basin fraction estimation with random sampling

Yes, this can me used with the bSTAB method, but the bSTAB method itself remains unafected. The design of bSTAB is: "the user provides a function that given an initial condition it returns a vector of features". Well, the users can use this special tracking mode as this function and the vector of features will infact have only one feature, the attractor label. 

So this means that we shouldn't need to actually do anything in #219 .

---

> the numbering of unidentified cells from 1 to 0.

Besides the fact that SparseArrays cannot be used in the end here, I think it is a good idea to keep this change. 0 is rather special, yet it is unused in our algorithm. 1 on the other hand is "the next free odd number", i.e., the next new attractor label.

awage:
> What do we gain by having the grid for basins and grid for attractors different? Does it actually help anywhere? 

Suppose you have attractors in the square [-5, 5]x[-5, 5] and you are interested only in a smaller part of this region you can use this method with a second grid or a random sampling of the smaller region. One grid stores only the attractors while the basins are stored in your custom grid. There is already a mode for this operation in `basins_of_attraction` but it needs two pass. In fact both functionalities are very similar, the difference is that you have the automatic detection of attractors in one case. 

Also, in the attractor grid the basins are **not updated** since the initial conditions are different. You don't have the speed improvement of hitting several basins in a row. 

The other benefit that I had in mind was using the sparse matrix for attractors only to avoid memory limitations. But lets forget about this part. 

I know that these are not strong reasons to have to separate grids. Anyway this functionality can be useful.

Datseris:
I'll need a bit of time before I can return here (more than a week probably), so just letting you know in advance. It is good to think about this thoroughly and let it simmer as much as possible before merging anything.

awage:
All right! I will leave it for now and have a though about parallel computing.

awage:
I leave here some ideas and results on the parallelization of the code.

* It is possible to get multithreading working with the second mode of operation. I used the tools developed in this PR: 

```
using DynamicalSystems

function multi_threaded_basins(bsn_nfo, integ, grid)
	nthreads = Threads.nthreads()
	bsn = zeros(Int16, map(length, grid))
	bsn_nfo_v = [deepcopy(bsn_nfo) for i in 1:nthreads]
	integ_v = [deepcopy(integ) for i in 1:nthreads]
	I = CartesianIndices(bsn)
	Threads.@threads for ind in I
		u0 = ChaosTools.generate_ic_on_grid(grid, ind)
		bsn[ind] = get_label_ic!(bsn_nfo_v[Threads.threadid()], integ_v[Threads.threadid()], u0)
	end
	return (bsn .- 1) .÷ 2
end

F = 0.27; ω = 0.1;  # smooth boundary
res = 500
ds = Systems.duffing(zeros(2), ω = ω, f = F, d = 0.15, β = -1)
xg = yg = range(-2.2, 2.2, length = res)
grid = (xg, yg)
@time basins, att = basins_of_attraction(grid, ds; T=2*pi/ω, diffeq = (;reltol = 1e-9, alg = Vern9()), show_progress = false)
bsn_nfo, integ = ic_labelling(ds; attractors = att, T=2*pi/ω, diffeq = (;reltol = 1e-9, alg = Vern9()))
@time bsn2 = multi_threaded_basins(bsn_nfo, integ, grid)
```
Timing for two threads: 
30.383229 seconds (2.00 M allocations: 137.854 MiB, 0.16% gc time)
14.088567 seconds (4.52 M allocations: 370.467 MiB, 0.35% gc time)

I think it is also possible to use `Distributed` and `pmap` to achieve the same. 

* I think it is possible to get a parallel version of the "normal" mode of operation. It is almost the same code as before but each thread operates on a part of the complete grid. After the computation the attractors must be matched between threads (the label may be changed from one worker to the other). The results are encouraging but need more work. 

Everything is at a early stage still.

Datseris:
Yes, the first method is certainly parallelizable as there aren't any race conditions. There isn't any shared array that different initia lconditions would write onto.

I don't think it is possible to parallelize the normal version. "each thread operates on a part of the complete grid", but you don't get to decide which part of the grid a trajectory visits, that's something we don't know in advance and hence can't plan. Different initial conditions would write different entries into the basins array if each started first. For me this means the method isn't parallelizable. But I'm interested to see what you have coded so far.

awage:
Each thread have an independent copy of the grid. It is as dumb as computing _n_  smaller basins.  It is a waste of memory, but it is maybe the price to pay to get better in speed.

This code works but I have only tested it on 3 examples: 

```julia
function multi_threaded_basins(bsn_nfo, integ, grid)
	nthreads = Threads.nthreads()
	bsn = zeros(Int16, map(length, grid))
	bsn_nfo_v = [deepcopy(bsn_nfo) for i in 1:nthreads]
	integ_v = [deepcopy(integ) for i in 1:nthreads]
	I = CartesianIndices(bsn)
	Threads.@threads for ind in I
		if bsn_nfo_v[Threads.threadid()].basins[ind] == 0
			u0 = ChaosTools.generate_ic_on_grid(grid, ind)
 			bsn_nfo_v[Threads.threadid()].basins[ind] = get_label_ic!(bsn_nfo_v[Threads.threadid()], integ_v[Threads.threadid()], u0)
		end
	end

	# Remove attractor labels from basins
	for k in 1:nthreads
		ind = iseven.(bsn_nfo_v[k].basins)
		bsn_nfo_v[k].basins[ind] .+= 1
		bsn_nfo_v[k].basins .= (bsn_nfo_v[k].basins .- 1) .÷ 2
	end

	# Match labels between basins.
	for k in 2:nthreads
		match_attractors!(bsn_nfo_v[k-1].basins, bsn_nfo_v[k-1].attractors, bsn_nfo_v[k].basins, bsn_nfo_v[k].attractors)
	end

	# Reconstruct the whole basins
	for k in 1:nthreads
		for ind in I
			if bsn_nfo_v[k].basins[ind] != 0
				bsn[ind] = bsn_nfo_v[k].basins[ind]
			end
		end
	end
	return bsn
end


# Continuous system
ω = 0.1617; F = 0.395 # fractal boundary
res = 200
ds = Systems.duffing(zeros(2), ω = ω, f = F, d = 0.15, β = -1)
xg = yg = range(-3., 3., length = res)
grid = (xg, yg)
bsn_nfo, integ = ic_labelling(ds; grid = grid, T = 2*pi/ω, diffeq = (;reltol = 1e-10, alg = Vern9()))
@time bsn2 = multi_threaded_basins(bsn_nfo, integ, grid)
# 44.148590 seconds (77.60 M allocations: 1.577 GiB, 0.63% gc time, 1.08% compilation time)

#@time basins, att = basins_of_attraction(grid, ds; T = 2*pi/ω, diffeq = (;reltol = 1e-10, alg = Vern9()), show_progress = true)
#  84.830231 seconds (78.47 M allocations: 1.689 GiB, 0.39% gc time, 2.17% compilation time)
```

The gain is roughly 2 on two threads.

Datseris:
@awage as a heads up please do not do parallelization additions in the _original_ basins implementation, otherwise this PR will be too complicated to review. Let's please keep this PR to (1) adding the new tracking mode/feature and (2) better naming for the algorithm variables in teh source code. Any parallelization for the original method should be done in a separate PR. Of course, parallelization of the _new_ method feel free to add in this PR.

src/basins/basins_lowlevel.jl
```diff
@@ -1,20 +1,20 @@
...
     get_projected_state::UF
     state::Symbol
-    current_att_color::Int
-    current_bas_color::Int
+    current_att_label::Int
+    visited_cell::Int
     consecutive_match::Int
     consecutive_lost::Int
-    prev_clr::Int
+    prev_lab::Int
```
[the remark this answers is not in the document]
awage: Ok, I will also update other variables like this one.

awage:
We are on the same page! Parallelization is always a delicate thing and it deserve another PR. I left the comments in this PR just to share some advances in this direction.

src/basins/basins_lowlevel.jl
```diff
@@ -131,125 +138,132 @@ is changed when the basins and attractors are returned to the user. Diverging tr
...
-    nxt_clr = (n[1]==-1  || isnan(u_full_state[1])) ? -1 : bsn_nfo.basin[n]
+    ic_label = (n[1]==-1  || isnan(u_full_state[1])) ? -1 : bsn_nfo.basins[n]
 
     # search attractors directly
     if !isnothing(bsn_nfo.search_trees)
+        bsn_nfo.consecutive_lost = (ic_label == -1 ? bsn_nfo.consecutive_lost + 1 : 0);
+        if norm(u_full_state) > horizon_limit || bsn_nfo.consecutive_lost > mx_chk_lost
+            return -1
+        end
+        # THIS RETURN IS TO BE DISCUSSED: DESIGN CHOICE IF WE WANT THE FSM IN THE SECOND MODE
```
awage, line 156: This change is to be discussed. In the second mode of `basins_of_attraction` do we need the FSM? It is only useful for the `:lost` state. We cannot guarantee a correct mode of operation of the FSM in this second mode. So I propose to skip the FSM but we should check if the trajectory is not lost somewhere without hitting and attractor.

awage: Also I checked that it improves the speed, for the Henon system: 

```julia
@time basins2, att2 = basins_of_attraction(grid, ds; attractors = att, show_progress = true )
#  780.273 ms (720079 allocations: 44.75 MiB) PR
#  1.561 s (1793985 allocations: 77.53 MiB) on Master. 
```

Datseris:
I've finished working on #219 finally. That was a big project. When I have free time later I will work here. My conclusion from working on #219, is that whatever method we have here, it should be a different function entirely than what #219 does. In #219 the clustering and approximate features are important. In our method, they aren't. Attractors are identified exactly and hence one does not need to be saving trajectories.

One important question: how is the basin fraction calculation happening in this PR? Do we (1) first find the attractors by letting the machine run for long times, or are the attractors found on the fly _while_ calculating fractions? I.e. a list of attractors is being kept that increases as the simulation runs and some new initial conditions that go to a different attractor are identified. My guess is that the latter would be better for performance?

awage:
> I've finished working on #219 finally. That was a big project. When I have free time later I will work here. My conclusion from working on #219, is that whatever method we have here, it should be a different function entirely than what #219 does. In #219 the clustering and approximate features are important. In our method, they aren't. Attractors are identified exactly and hence one does not need to be saving trajectories.

Exactly, this is why we evolve the trajectory step by step until we decet the attractor or the basin. 

 
> One important question: how is the basin fraction calculation happening in this PR? Do we (1) first find the attractors by letting the machine run for long times, or are the attractors found on the fly _while_ calculating fractions? I.e. a list of attractors is being kept that increases as the simulation runs and some new initial conditions that go to a different attractor are identified. My guess is that the latter would be better for performance?

The fractions are not computed as such, but it can be implemented straightforwardly. To answer you question the attractors are detected on the fly using the finite state machine with the normal mode. But the second mode can also be used if the attractors have been supplied.  

The function `get_label_ic!` match the initial condition to an attractor stored in a structure `BasinsInfo`. All the operations are the same as computing the entire basin using `basins_of_attraction` **but** the information on the basins in `BasinsInfo` is not updated. The reason is that the grid and the initial conditions do not match necessarily. 

Here is a summary of the changes in this PR: 

* renaming of some variables for clarity of the algorithm. 
* The unlabelled initial conditions are now 0 by default instead of 1. 
* The second mode of operation that compare distances from state point to attractors have been modified such that the finite state machine is skipped. A side effect is that the performances improved. 
* Function `ic_labelling` to initialize the structure `BasinsInfo` and the integrator. This is `basins_of_attraction` under the hood and the two mode of IC labelling are possible (on the fly attractor detection and distance matching). 
* Function  `get_label_ic!` that match the IC to an attractor. This function can be used to compute the basins fractions in parallel given a bunch of ICs. 
* Some minor changes in `estimate_basins!`

Datseris:
Okay, I've had some time here now. I propose a different API which makes more sense.

Define a struct
```julia
struct AttractorMapper{B<:BasinInfo, I, A}
    basin_info::B
    integ::I
    attractors::Dict{Int8, A}
end
```
An instance of `AttractorMapper` is created using the code of `basins_of_attraction` with this new keyword `ic_lab_mode = true`.
(notice that `basin_info` has the `grid` inside it)

then the labelling interface should be:

```julia
mapper = AttractorMapper(.....)
label = mapper(u0) # for a given initial condition it gives you the label. Literally is `get_label_ic!`
```

Importantly, this new functionality **should not be part of `basins_of_attraction`, especially not via a keyword argument.**. Let's keep things simple in the front end API. Of course, the source code of `AttractorMapper` is like 20 lines or so, as it re-uses everything from `basins_of_attraction`, but the point is: for a user, this new mode / mapping functionality should just stay separate. And for us, as developers, the cleaner we can make this separation, the better. If we need to copy 5 lines of code from `basins_of_attraction` to actually initialize the `AttractorMapper`, then that's okay. We really just need to create the BasinsInfo structure and the integrator. I really want to avoid having multiple functionalities in one function name. It will make our work harder down the line. 

**In fact, I'd even argue that generating the BasinInfo and the integrator should be a function in itself, that both `basins_of_attraction` and this new `AttractorMapper` will call.**

@awage if you are unfamiliar with making Function Like Objects (which is what this `AttractorMapper` is), have a look here:  https://docs.julialang.org/en/v1/manual/methods/#Function-like-objects , or just ask me questions.


---

Everything else in the PR is okay!

Datseris:
Actually @awage let me do the changes I suggested, you've done already a lot of work on this :)

Datseris:
@awage I've done the changes I outlined. Please do have a look and let me know if you like this code. I'm still not sure about the name `AttractorMapper`, perhaps we can find a better one :) 

Oh, by the way does `get_label_ic!` return the "correct" labels, 1, 2, 3, 4 or does it return the "only odd number" labels?

awage:
Nice! Now the two functions are clearly separated. Functions-like objects are a good way to attach a method to type. I will keep this trick in mind. 

I changed some small details and fixed the test. 

Maybe the last thing I would add is a convenience function to get the attractors from the mapper.

src/basins/basins_attractor_mapping.jl
```diff
@@ -25,17 +25,19 @@ end
 
 function AttractorMapper(ds;
         # Notice that all of these are the same keywords as in `basins_of_attraction`
-        grid = nothing, attractors = nothing,
+        grid = 0, attractors = nothing,
```
Datseris, line 28: What's the purpose of using `0` instead of `nothing`?  I'd really recommend to use `nothing` instead of `0`. `nothing` is a singleton type, but `0` isn't. `nothing` is also the de-facto default value in Julia for "not giving a value".

awage: It is a dirty temporary fix since the other parameters are calling length(grid) it raises an exception. `length(nothing)` does not exists.

Datseris: you can set `grid` to a dummy grid _inside_ the `AttractorMapper` function if the given `grid` is nothing, but the keyword default should remain `nothing`.

Datseris: ah okay you did that already.,

src/basins/basins_attractor_mapping.jl
```diff
@@ -0,0 +1,53 @@
...
+    end
+    if isnothing(complete_state)
+        complete_state = zeros(eltype(get_state(ds)), length(get_state(ds)) - length(grid))
+    end
+    bsn_nfo, integ = basininfo_and_integ(ds, attractors, grid, Δt, T, SVector(idxs...), complete_state, diffeq)
+    return AttractorMapper(bsn_nfo, integ, kwargs)
+end
+
+function (mapper::AttractorMapper)(u0; kwargs...)
+    lab = get_label_ic!(mapper.bsn_nfo, mapper.integ, u0; mapper.kwargs...)
```
Datseris, line 51: @awage I am asking here again: does this `lab` correspond to the labelling the user gets from `basins_of_attraction`, and hence is always 1, 2, 3, 4 (the integers), or is this labelling the "odd numbers only" that we also describe in the paper and labels attractors? We need to clarify this, and convert it appropriately if it is the latter.

src/basins/basins_attractor_mapping.jl
```diff
@@ -0,0 +1,53 @@
...
+It has the same keywords as [`basins_of_attraction`](@ref),
+so see there for how attractors are identified and what keywords you can use.
+**One of keywords `grid` or `attractors` is mandatory for `AttractorMapper`.**
+
+`AttractorMapper` has two modes of operation: **unsupervised**, with automatic detection of
+attractors and **supervised**, where the attractors are known.
+Both modes are in fact the same process that occurs in [`basins_of_attraction`](@ref),
+depending on whether the `attractors` keyword is given or not.
+"""
+struct AttractorMapper{B<:BasinsInfo, I, K}
```
Datseris, line 20: @awage can you brainstorm a couple of alternative names for this object? The problem with this name is that it is too generic. One could also make an `AttractorMapper` via the featurizing approach of the bSTAB method... So perhaps we could have `AttractorMapperRecurrences` and `AttractorMapperFeatures`?

awage: Hi, I will be busy these days. I'll try to catch up next week.

src/basins/basins_attractor_mapping.jl
awage: The labelling is the same that is used in `basins_of_attraction`. The conversion is done on line 52.

src/basins/basins_attractor_mapping.jl
awage: We can leave the name `AttractorMapper` or `BasinMapper` (the later is maybe more appropriate) and dispatch on the arguments. There would be three modes for this functionality: 
* Searching with the recurrences on the grid. 
* Matching the attractors within a distance ε (given the list of attractors). 
* Using the feature function and do some clustering. 

Other names can be `BasinTagger`, `BasinTag`, `BasinId`, `AttractorId`. The last two ones are shorter so you can form `BasinIdFeatures` or `BasinTagFeature`.

Datseris: @awage you know what, this is a brilliant idea. We could be merging interfaces instead of having new ones. Currently the featurizing approach also has the code that "maps" to attractors based on distance from attractors (so caalled "supervised approach). And you have the same here. This is duplication.

Okay, let me work on a unified interface, I have an idea of how to do this so that it is flexible and allows all three options you outlined. 

With this API, I like the `AttractorMapper` name more, as the type of an argument to the mapper decides what more it does!!!

Datseris: I'm working on this now. But I rrealized, we can't use the featurizing+clustering approach for this, because to my understanding, this approach can only happen once all points are calculated and _then_ the clustering algorithm is called. @KalelR can confirm this.

By the way @KalelR , it seems to be that the supervised version of the featurizing approach is useless: the basins via recurrences code also has the "suppervised" version, where one directly gives attractors to the function. Is there any functional difference between the supervised version with the featurizer and the supervised version implemented in this PR?

awage:
> I've done large changes and everything is broken now, I just don't have the strength to code more for tonight. Will summarize stuff soon. Please don't push any commit to this code until I finish the changes :)

You are in charge. If you need any help please tell.

Datseris:
Okay @awage and @KalelR , I've finished the design here. Haven't corrected tests, but doesn't matter for now. Please read the new design and provide feedback. The new design is in the following files: `attractor_mapping.jl` first. Then, the `attractor_mapping_proximity.jl` and `_recurrences.jl`. Finally, the `basin_fractions_clustering.jl`. 

The benefits of this approach:

* Unified api for calculating basins of attraction irrespectively of the method
* Significant elimination of code duplication.
* Simplification of the clustering algorithm docstring
* It all makes more sense from a pure dynamical systems perspective: initial conditions are mapped to attractors with some methodology. The methodology is just an instance of the abstract type `AttractorMapper`.

If you approve this new design I'll move on to:

1. Change the source code of `basins_of_attraction` to initialize and use an `AttractorsViaProximity` mapper and re-use its source code.
2. Fix tests.

Please comment on the names: So far I have all these names `AttractorsViaWhatever`. I don't like them very much but couldn't think of anything better anyways.

awage:
Nice! I will be able to start working on Monday on this code. Also I will start the interface for `StroboscopicMap`.

Datseris:
Please focus on the StroboscopicMap for now as I'm not done yet here!

Datseris:
@KalelR just for reference, featurizing the Henon map seems to create problems:
```
        f(A, t) = [mean(A[:, 1]), mean(A[:, 2])]
        mapper = AttractorsViaFeaturizing(ds, f; Ttr = 100)
        henon_fractions_test(mapper)
```
errors with 
```
ERROR: ArgumentError: radius NaN must be ≥ 0
Stacktrace:

_dbscan(kdtree::NearestNeighbors.KDTree{SVector{2, Float64}, Euclidean, Float64}, points::Matrix{Float64}, radius::Float64; min_neighbors::Int64, min_cluster_size::Int64) at [C:\Users\m300808.julia\packages\Clustering\tt9vc\src\dbscan.jl]()
```
I am assuming this is because trajectories diverge to infinity and therefore some feature values become `Inf`, which makes your optimal radious become `NaN`, which makes DBSCAN fail. I am not sure how to fix this, i.e., what kind of featurizing functions I should define that do not fail...

Datseris:
Note to self: todo: pretty printing for attractor mappers.

Datseris:
Note to self: todo: autimatic estimation of ε in `Proximity` version as the minimum distance between all attractors.

KalelR:
> I am assuming this is because trajectories diverge to infinity and therefore some feature values become Inf, which makes your optimal radious become NaN, which makes DBSCAN fail. 

Yes, you're right. I just tested this to make sure. DBSCAN itself just ignores the Inf values, the error was that the optimal radius `ε_optimal` was NaN. 

> I am not sure how to fix this, i.e., what kind of featurizing functions I should define that do not fail...

Assigning a new value to Inf features is one way, but it's not ideal I think. If the reassigned value is too large, the optimal radius ε will also become large, because of the way we initially guess viable radii:
     `feat_ranges = maximum(features, dims=2)[:,1] .- minimum(features, dims=2)[:,1]`
    `ϵ_grid = range(minimum(feat_ranges)/200, minimum(feat_ranges), length=200)`

which might lead to nonoptimal results.

Maybe instead we should instead disconsider all NaN features, and assign them to a same cluster?

> I think the clustering method is just incapable of detecting infinity

What do you think is the ideal way to deal with infinities? I would think assigning them to the same attractor makes sense, doesn't it? We can do that.

> It just considers it as another attractor, and hence we should mention this in the docstring.

For sure.

Datseris:
@awage I've pushed the code that computes `ε` of proximity version automatically as half the minimum distance of points across attractors (if an ε is not given). You can have a look now for why proximity and recurrences give wrong results.

test/basins/attractor_mapping_tests.jl
```diff
@@ -97,9 +97,7 @@ using Statistics
         # Deterministic test, should be tested with exact accuracy
         fs, labels = basin_fractions(mapper, ics; show_progress = false)
         @test sort!(unique(labels)) == [1,2,3]
-        @test fs[1] == 0.223
-        @test fs[2] == 0.191
-        @test fs[3] == 0.586
+        @test sort!([fs[i] for i in 1:3]) == [0.165, 0.193, 0.642]
```
Datseris: That's okay, but we still need to test that the method not only finds the correct fractions, but also attributes them to the correct attractors.

Datseris:
So, the bSTAB algorithm is _much_ worse than we thought. It is impossible for it to get any decent mapping of the fractions for the complex Lorenz84 case, where half the time it only finds two clusters, and many times it also attributes -1 to some initial conditions. We should point this out in a next paper. Devise a method/way to quantify the "sensitivity" of each algorithm, like, if we change parameters, how badly does it perform, etc. etc,..

Datseris:
Damn, this PR is an insane amount of work. I'll need several days to finish it. I have planned out everything to do thankfully, now only coding it remains.

awage:
>  It is impossible for it to get any decent mapping of the fractions for the complex Lorenz84 case, 

An extension of this method would be to automatically find the feature function that maximizes the distances between attractors in the feature space. Maybe some fancy adaptive filtering method. But this is easier said than done.  

If you need some help for tests or coding I am available.

Datseris:
I am currently re-writting the test suite entirely, because it was simply not maintainable to write tests for each invidual case. Once I am done with that, I Will ask you to add a call to `basins_of_attraction` in the test suite for all tests I've written and make sure that `basins_of_attraction` works as before, with all possibliities as before. It is crucial that this PR does _not_ introduce breaking changes, only deprecations. Code that worked before should still work now... I'll let you know once I am done. Almost there......

Datseris:
@awage I have written the tests in a generalizable way. I am very happy with the way I've written the tests, but there is still much work to do and I would appreciate any help because I need to turn to and release Agents.jl 5.0. The tests are in the file `attractor_mapping_tests.jl`. At the end of the file I outline the TODOs remaining:

```
TODO: Add a call to `basins_of_attraction` within the `test_basin_fractions` function.
TODO: Tests for  projected system (magnetic) and poincare map (thomas cyclical)
TODO: Tests of Lorenz84 fail, significant differences between recurrence and proximity
TODO: Tests of Duffing fail, significant differences between recurrence and proximity
```

anything you can hack is great help, I'll continue working on this when I get some more time!

awage:
OK! I'll start to work on this.

src/basins/attractor_mapping_featurizing.jl
```diff
@@ -143,8 +143,13 @@ function extract_features(mapper::AttractorsViaFeaturizing, ics::Union{AbstractD
...
 function extract_features(mapper::AttractorsViaFeaturizing, u0::AbstractVector{<:Real})
-    u = trajectory(mapper.ds, mapper.total, u0;
-    Ttr=mapper.Ttr, Δt=mapper.Δt, diffeq=mapper.diffeq)
+    if isdiscretetime(mapper.ds)
+        u = trajectory(mapper.ds, mapper.total, u0;
+                Ttr = mapper.Ttr, Δt = mapper.Δt)
+    else
+        u = trajectory(mapper.ds, mapper.total, u0;
+                Ttr = mapper.Ttr, Δt = mapper.Δt, diffeq = mapper.diffeq)
+    end
```
Datseris, line 152: What's the point of this block? In discrete systems `diffeq` is propagated nowhere. If `diffeq` was propagated somewhere in a discrete system, that is a bug that needs to be fixed in DynamicalSystemsBase.jl. This code block should be reverted to the original one.

test/basins/attractor_mapping_tests.jl
```diff
@@ -111,21 +111,19 @@ end
...
 
     function lorenz84_featurizer(A, t)
-        x1 = ChaosTools.Entropies.genentropy(Dataset(A[:, 2:3]), 0.1)
-        mini, maxi = minmaxima(A)
-        x2 = maxi[1] - mini[1]
-        x3 = maxi[3] - mini[3]
-        return [x1, x2, x3]
+        a,b = ChaosTools.data_boxing(Dataset(A), 0.1, 3)
+        x1 = length(b)
+        return [x1*0.1]
```
Datseris, line 116: Oh my GOD this is so much smarter!!! Nice!!!

src/basins/attractor_mapping_featurizing.jl
awage: I forgot about that. The program was throwing an error when using the stroboscopic map because the `diffeq` keyword is not defined in the `discrete_trajectory` function (in `DynamicalSystemsBase`). I don't know what is the correct way to do it. Is it ok to propagate the keyword even if it is not defined for this special integrator?

awage:
The unsupervised method of clustering features gives me a lot of trouble. Now I have a strange error with the Lorenz84 system. I leave this part for now and focus on the Poincaré map tests.

src/orbitdiagrams/poincare.jl
```diff
@@ -270,6 +274,10 @@ function DynamicalSystemsBase.reinit!(pmap::PoincareMap, u0)
 	reinit!(pmap.integ, u0)
 	return
 end
+function DynamicalSystemsBase.reinit_on_the_plane!(pmap::PoincareMap, u0)
+	reinit!(pmap.integ, u0)
+	return
+end
```
awage, line 280: We need some function that initiates the values on the plane for `basins_of_attraction` and `attractor_mapping_recurrence`. I don't if it is possible to dispatch `reinit!` depending on the length of `u0`. Maybe there is a better solution. I don't know

Datseris: If a user uses a Poincare Map, the grid must already be defined "on the plane", otherwise it doesn't make sense I think, right? But we still need to face the problem of how to transform the initial condition on the plane to an initial condition for the integrator, as we are missing one dimension... Correct?

Datseris: I think I've made this possible with my latest commit. Can you give it a shot?

awage: Yes, it's the point. With the new interface the integrator takes more responsability. The function `basins_of_attraction` just calls `step!` and `reinit!` without knowing much about the dynamical system. The `complete_and_reinit!` is know in the integrator side.

Datseris: sure, but this fixes is it then, at least in the case where the hyperplane is `(j, r)`. I didn't know how to do the generic case yet.

src/orbitdiagrams/poincare.jl
```diff
@@ -231,7 +231,7 @@ next_state_on_psos = step!(pmap)
 """
 function poincaremap(
 		ds::CDS{IIP, S, D}, plane, Tmax = 1e3;
-	    direction = -1, u0 = get_state(ds),
+	    direction = -1, u0 = zeros(D),
```
Datseris, line 234: I'm confused by this change. Can you justify why the initial condition should be zeros instead of the state of the system...?

src/orbitdiagrams/poincare.jl
```diff
@@ -303,13 +308,21 @@ end
 
 
 function DynamicalSystemsBase.get_state(pmap::PoincareMap)
+	if pmap.integ.t < pmap.Tmax
+		return pmap.proj_state[pmap.diffidxs]  # Possibly breaking change
+	else
+		return pmap.proj_state[pmap.diffidxs]*NaN
+	end
+end
+function get_full_state(pmap::PoincareMap)
```
Datseris, line 317: We have defined the Poincare map to always return the full system change, and I think it makes sense this way. It is a problem to change it now in this respect, and as you've noticed, it is indeed a breaking change. The worst part is that it is not possible to do this for a generic plane anyways. 

I think you need to write an internal function `_get_projected_state` that for all other integrators it simply returns `get_state` while for Poincare Map it gives you the projected state.

awage:
The `PoincareMap` integrator code is becoming ugly. There is an ambiguous choice about the state of the integrator: is it the full state of the dynamical system or just the state on the plane. There is surely a simpler approach but right now I am too tired to think straight.

src/orbitdiagrams/poincare.jl
awage: Forget about that, it was a change while debugging. I'll revert this.

src/orbitdiagrams/poincare.jl
awage: It is also my opinion. I was too tired to find another solution. Maybe we can consider encapsulate a `PoincareMap` into a `ProjectedSys`. The Poincaré map would be a special case of projected system.

awage:
I think I found the solution. Maybe it is too complicated: I use a projected integrator to wrap a Poincaré Map. 

This way there are minimal changes to the Poincaré Map integrator interface. It seems to work but I don't know if there is a regression because of this integrator within an integrator.

awage:
The interface is not intuitive but it works: 

```julia 
    pmap = poincaremap(ds, (3, 0.0), 1e6;
        rootkw = (xrtol = 1e-8, atol = 1e-8), diffeq=(reltol=1e-9,)
    )
    psys = projected_integrator(pmap, 1:2, [0.])
```

I think it is possible to create a better integration of both integrators together. Maybe with a keyword in the Poincaré Map interface such as `projected = true` so that the integrator is defined on the plane?

Datseris:
Unfortunately I do not agree with combining or connecting the projected integrator with the Poincare map. They are fundamentally different things. The Poincare map is not projecting the state into a lower dimensional plane, but rather it intersects it with it. This is why the Poincare map has the fundamental property of 1-to-1 correspondance with the continuous dynamics, with a projected system in general does not. The solution here is a simple low level function that for all integrators returns `get_state` while for the poincare map it returns the "intersected" state. This changes nothing in no integrators and is a 5 lines of code change.

Datseris:
@awage can you please revert the changes you did to the Poincare map, and starting from there we see what we is missing and what we need to do in code? It is really difficult for me to do this via the commit history, because there was also a lot of testing around. Generally speaking, if you have an idea to do anything breaking, it is always better to first discuss it, and then commit code after we have settled on a design. This will likely save time of re-thinking or re-implementing something.

awage:
> Unfortunately I do not agree with combining or connecting the projected integrator with the Poincare map. They are fundamentally different things. The Poincare map is not projecting the state into a lower dimensional plane, but rather it intersects it with it. This is why the Poincare map has the fundamental property of 1-to-1 correspondance with the continuous dynamics. A projected system does not have this property. Hence, I think it is more intuitive if these two things are separated.
> 
> As for the code, one possible solution is a simple low level function that for all integrators returns `get_state` while for the poincare map it returns the "intersected" state. This changes nothing in no integrators and is a 5 lines of code change.

Ok, I understand your argument. Although I still find the idea of reducing the dimension appealing. I though I was on the right track this is why I pushed the commits. I will revert to where you left it.

Datseris:
I've pushed some commits I was working on a different branch (idenepdentn improvements).

Alright, lets reorganie ourselves. So, at the moment, the "problem" we have with the Poincare map is that the grid is D-1 dimensions, but the state of the map is D dimensional, right? My proposal is to just define a new function `_get_state_on_grid`, which for all integrators is just `get_state`, but for the poincare map it returns the state on the plane, which has one less dimensio0n. This internal function `_get_state_on_grid` is just used in the AttractorsViaRecurrences code.

This has the advantange of not breaking anything. What do you think of this approach @awage ?

Datseris:
Just letting you know that I've tested this locally @awage . I changed absolutely nothing, and only added one more `reinit!` method to the `PoincareMap` tht works for a given state with D-1 dimensions. _Everything works, basin_fractions, etc._, even though I had to write no extra code.

awage:
> Just letting you know that I've tested this locally @awage . I changed absolutely nothing, and only added one more `reinit!` method to the `PoincareMap` tht works for a given state with D-1 dimensions. _Everything works, basin_fractions, etc._, even though I had to write no extra code.

That's wizardry and julia black magic. Thanks! I will try to have a look this weekend.

Datseris:
> That's wizardry and julia black magic. Thanks! I will try to have a look this weekend.

Wait until I push the commits first. Doing some small debugging at the moment. Will probably not have time to finish this today because i'm in london now, but will do it whenever I have some spare time. Soon soon.

awage:
> AttractorsViaRecurrences

We will need a function of this kind. Another possibility is to get the index of the variables that correspond to the grid (D-1 indices in the case of the Poincaré Map). The only place where these indices are needed is in the function `basin_cell_index` where you get the correspondence between the coordinates on the grid and the index of the basins matrix. 

These indices can be stored in the `BasinsInfo` struct. What do you think?

Datseris:
Sorry, I don't understand where you are referring to. We will need a function of "what" kind? And for what purpose?

awage:
>This internal function `_get_state_on_grid` is just used in the AttractorsViaRecurrences code.
I was referring to this function. If we use the indices we don't need this function for `AttractorsViaRecurrences`.

Datseris:
Give me a second to push some commits, I am almost done. Then we regroup and recap one more time and see what is remaining. By the way, `trajectory(pmap::PoincareMap)` works just by itself :D Simply because we defined this `trajectory_discrete` stuff. Haha I'm so happy when things just magically work.

awage:
I think it will depend on where you want to put the functionality. The key functions are: 
* `reinit!` : this function goes from the grid to the dynamical system state space. 
* `get_state_on_grid`: this function goes from the state space to the grid. 

You can integrate these functions in the `AttractorsViaRecurrences` mapper or in the integrator. This is a design choice.

Datseris:
@awage I've finished with my changes. Poincare map works without a problem, and as I said, I actually only had to write two small functions for it and nothing was a breaking change. Please also have a look at the new `basins_of_attraction` function inside the `attractor_mapping.jl` file. It makes the previous version completely obsolete by allowing the user to fierstr initialize an attractor mapper, and then get the basins.

What todos remain:

- [ ] Fix tests for Poincare map, strobscopic map, and projected system (i.e. make them have "passing values". All code is running correclty)
- [ ] Add a call to `basins_of_atttraction` inside the `test_basins` function
- [ ] DECISION: For the vcase of projected systems, what do we store as attractors? Do we store the projection of the attractors, or do we store the "original" attractors...?

awage:
This is really good stuff! I will work on the test suite on Monday and include the `basins_of_attraction` call. 

Regarding the decision on the projected case: the expected behavior is to store the full state as `basins_of_attraction` was doing. For example, in the magnetic pendulum you get the attractors in the state space. It is a valuable information.

Datseris:
Okay, then we would have to make this change in the "store attractors" part, by defining a function `_get_full_state` that for all integrators is just `get_state` but for projected integrator it gives the full state. Also note that the BasinInfo constructor needs to be changed so that `D` is the length of `_get_full_state`.

Datseris:
@KalelR we need a bit of your help here. The lorenz84 system does not seem to work no matter what we do. I am a bit curious, where did you find the code for the `optimal_radius` function from? For the lorenz84 system (see currently pushed code in the test file `attractor_mapping_tests.jl`, I get the following features:

image

I'm just finding it impossible to believe that a clustering algorithm wouldn't find three clusters here. So something is wrong. I also get: 

```
julia> ChaosTools.optimal_radius_dbscan(features, 10)
1.4214954693580018
```

So, something tells me our implementation is wrong because the features I show in the picture pasted above are simple too clearly divided into three clusters to believe that a clustering algorithm wouldn't work. Can you please check and also mention where is this optimal radius coming from, and why wouldn't the Clustering.jl package already have an algorithm to deduce such radious? I thought DBSCAN worked out of the box, it could find the clusters without further input from the user besides the scatter points.

test/basins/attractor_mapping_tests.jl
```diff
@@ -114,10 +114,8 @@ end
     expected_fs_raw = Dict(2 => 0.165, 3 => 0.642, 1 => 0.193)
 
     function lorenz84_featurizer(A, t)
-        a, b = ChaosTools.data_boxing(A, 0.1, 3)
-        g = genentropy(A, 0.1)
-        x1 = length(b)
-        return [x1*0.1, g]
+        g = exp(genentropy(A, 0.1; q = 0))
+        return [g, g]
```
Datseris, line 118: @awage FYI

Datseris: this `g` and `length(b)` from before are the same quantity. g is massively faster to compute

Datseris:
Actually @KalelR the current tests for Lorenz84 gives error `ArgumentError: reducing over an empty collection is not allowed` in the optimal radious code. :(

Datseris:
So I have currently commented out the tests of the clustering version and I added a warning saying that the code is bugged if someone initializes the mapper. We really need to test this more rigorously but the problem is this massive PR really needs to be merged soon (and also I am presenting it in two days :D ) . Since this new design works nicely, and all other tests pass, I'll open a new PR at the DynamicalSystems repo and then merge the present PR and tag.

awage:
Nice work! Just a question: the Duffing tests fails on my machine. The fractions I get are: `2 => 0.509, 1 => 0.491`. Maybe it is a random number generator issue. 

I will check the old tests for basins. I saw some tests failing.

Datseris:
For the old `basins_of_attraction`, we only need to make sure the deprecations are working. Because it has been now superseeded by the new `basins_of_attraction(mapper::AttractorMapper, grid)`. For the tests that give different results in your machine: okay, no worries, we should simply increase the error tolerance level then.

Datseris:
This is because `complete_state` for a `projected_integrator` must return the full state, not just the missing elements. This is because now a projection is any generic function, it doesn't have to be just "take the first two coordinates".

To fix this you simply need to define a wrapper inside the old `basins_of_attraction` that just creates a new function `complete_state_2 = make complete state from the old complete_state that gives fewer elements`. and use `complete_state_2` in the `projected_integrator` that is initializes within the old `basins_of_attraction`.

awage:
> To fix this you simply need to define a wrapper inside the old `basins_of_attraction` that just creates a new function `complete_state_2 = make complete state from the old complete_state that gives fewer elements`. and use `complete_state_2` in the `projected_integrator` that is initializes within the old `basins_of_attraction`.

Do you want me to push this change?

Datseris:
if you have coded it sure go ahead

awage:
By the way, the tests for mapper seem broken. It has something to do with `labels` in `basin_fractions`

Datseris:
@awage you still get exactly the same optimization in the new interface as long as you give the same grid to both `AttractorsViaRecurrences` and `basins_of_attraction`. Right? or have I misunderstood somethinbg?

awage:
I narrowed down the problem to `extract_attractors` in attractor_mapping_featurizing.jl. 

There is a problem with the type of `ics`. It not recognized as a function.

awage:
> @awage you still get exactly the same optimization in the new interface as long as you give the same grid to both `AttractorsViaRecurrences` and `basins_of_attraction`. Right? or have I misunderstood somethinbg?

No, you don't have this optimization: 

```julia
function (mapper::AttractorsViaRecurrences)(u0)
    # Low level code of `basins_of_attraction` function
HERE ->   lab = get_label_ic!(mapper.bsn_nfo, mapper.integ, u0; mapper.kwargs...)
    # Transform to integers indexing from odd-even indexing
    return iseven(lab) ? (lab ÷ 2) : (lab - 1) ÷ 2
end
``` 
The matrix `bsn_nfo.basins` is not updated. That's why the old interface is still usefull. But we can think of an adaptation for the case where the grids matches.

Datseris:
okay this is a problem that we have to fix because the old file will be deleted in the future. It shouldn't be hard to fix this. I'll think about it.

Datseris:
@awage it is a bit weird if we store the full attractors in the projected system. I understand it is more beneficial this way, but unfortunately, if a user uses the Proximity version, they have to give projected attractors. This makes it weird that in one algorithm for finding attractors we would store full states, and in the other we would store projected states. Maybe we should consider storing projected states in every case? In the end of the day the user only needs to apply the projection function in all attractor points to get the "full" attractor.

awage:
> Maybe we should consider storing projected states in every case? In the end of the day the user only needs to apply the projection function in all attractor points to get the "full" attractor.

I have no problem with this option if its more convenient. I mean, if it was possible to have the full state, better. But the design should be coherent also.

awage:
Wow!! Hats off!

KalelR:
> @KalelR we need a bit of your help here. The lorenz84 system does not seem to work no matter what we do. I am a bit curious, where did you find the code for the `optimal_radius` function from? For the lorenz84 system (see currently pushed code in the test file `attractor_mapping_tests.jl`, I calculated the features for all given predetermined initial conditions get the following features (x axis is boxing length, y axis is attractor entropy):
> 
> image
> 
> we see the three attractors of the system: fixed point in lower left corner, periodic orbit in high middle, and chaotic attractor in high right. However, our clustering algorithm actually only finds two clusters. It groups the top two clusters into one. I also get:
> 
> ```
> julia> ChaosTools.optimal_radius_dbscan(features, 10)
> 1.4214954693580018
> ```
> 
> So, something tells me our implementation is wrong because the features I show in the picture pasted above are simply too clearly divided into three clusters to believe that a clustering algorithm wouldn't work. Can you please check and also mention where is this optimal radius coming from, and why wouldn't the Clustering.jl package already have an algorithm to deduce such radious? I thought DBSCAN worked out of the box, it could find the clusters without further input from the user besides the scatter points.


Hey, sorry for the late reply!

I got `optimal_radius` from the [code](https://github.com/TUHH-DYN/bSTAB) made available by the authors of the bSTAB paper. In the [specific code for the radius](https://github.com/TUHH-DYN/bSTAB/blob/master/bSTAB-M/utils_bSTAB/classify_solution.m) there they actually comment that "finding the optimal epsilon can be challenging" haha. 
There are alternative methods that might work better, but I'd only be sure trying. 

The `optimal_radius` was indeed throwing an error, which was rather silly. It is easily fixable with an additional check. 


Could you send me the `featurizer` you used to generate that figure? I can't replicate it, but I'm not sure if I followed your discussions. I tried using
```
function featurizer(A, t)
    a, b = ChaosTools.data_boxing(A, 0.1, 3)
    g = genentropy(A, 0.1)
    x1 = length(b)
    return [x1*0.1, g]
end
```
and got 
image

The algorithm finds 2 clusters here also, but that is reasonable for the features.

Datseris:
@KalelR can you please try with featurizer _currently_ in the test suite? The tests are not run but the code is there: https://github.com/JuliaDynamics/ChaosTools.jl/blob/master/test/basins/attractor_mapping_tests.jl#L123-L126 

this gives for sure three very clearly different clusters.

Datseris:
> The optimal_radius was indeed throwing an error, which was rather silly. It is easily fixable with an additional check.

Can you do a PR for this?

KalelR:
> @KalelR can you please try with featurizer _currently_ in the test suite? The tests are not run but the code is there: https://github.com/JuliaDynamics/ChaosTools.jl/blob/master/test/basins/attractor_mapping_tests.jl#L123-L126
> 
> this gives for sure three very clearly different clusters.

Did you plot g vs log(g)? Why return `[g,g]` and not `[g]`?

g vs g is like this:
image


The full relevant code is:
```
F = 6.886; G = 1.347; a = 0.255; b = 4.0
ds = Systems.lorenz84(; F, G, a, b)
u0s = [
    1 => [2.0, 1, 0], # periodic
    2 => [-2.0, 1, 0], # chaotic
    3 => [0, 1.5, 1.0], # fixed point
]
diffeq = (alg = Vern9(), reltol = 1e-9, abstol = 1e-9)
M = 200; z = 3
xg = yg = zg = range(-z, z; length = M)
grid = (xg, yg, zg)
expected_fs_raw = Dict(2 => 0.165, 3 => 0.642, 1 => 0.193)

function featurizer(A, t)
    g = exp(genentropy(A, 0.1; q = 0))
    return [g, g]
end
mapper = ChaosTools.AttractorsViaFeaturizing(ds, featurizer; diffeq, Ttr = 100)
sampler, = statespace_sampler(Random.MersenneTwister(1234); min_bounds = minimum.(grid), max_bounds = maximum.(grid))
ics = Dataset([sampler() for i in 1:1000])
feature_array = ChaosTools.extract_features(mapper, ics; show_progress=true)
feature_array_s = ChaosTools.extract_features(mapper, sampler; show_progress=true, N=1000)
```

Datseris:
Yeah you can just use `[g]` no difference there. Okay in this figure the clusters aren't as clear. I guess it depends on initial conditions and `N`, the number of trajectories. Only for the lorenz84, let's increase N to 10000, and probably return not only `g`, but also the `maximum(A[:, 1])`.

KalelR:
Increasing N does not change much. These are the results with `maximum(A[:, 1])`
image

Using g and the "period" estimated via `estimate_period(A[:,1], :zc)` the separation becomes better
image

`:zc` might not be the best estimator, but it works for this case. And like you said, the method only finds two clusters for the automatically found `optimal_radius`. Taking a closer look, the estimated radius is too large.  To see this, I normalized the data, so it fell within [0,1] (which we should we probably do always, since DBSCAN is usually using Euclidian distances). 
image
Any radius below 0.5 or so works fine, and clustering returns  `(2 => 0.193,   3 => 0.165,  1 => 0.642)`. Above that, it groups the two upper clouds into a single cluster, and returns `(2 => 0.358,  1 => 0.642)`.  And this transition is pretty sudden: I guess as soon as a few points belonging to the two clouds are connected, then everything will be, because of how DBSCAN works.

So the problem really is with the estimation of the `optimal_radius`, and I don't know a good solution at the moment. I'd need to think more. Maybe just finding a better estimation method, clustering algorithm, distance metric, or something...

I can open a PR fixing the small issue that was present in `optimal_radius`, but this does not solve the full problem yet of a good estimation. What do you suggest?

Datseris:
please open a PR fixing the small issue that was present in `optimal_radius`. Using the period was a smart idea. I think we should remove the warning of the method not working, and also uncomment the tests in the test suite in your PR.

Datseris:
@KalelR could you please provide the simple fix for `optimal_radius` so that it doesn't error? You mentioned it was really simple to fix, but I still don't know how, so it would be good if you could open a PR for it or simply provide the code here and I can do the update.,

KalelR:
Sure, I'm very sorry for taking so long. I hadn't seen you had responded 9 days ago already!

PR #239, opened 2022-01-10
Fix nearest neighbors search of attractors

### LINKED ISSUE #238: Unexpected time while using the second mode for basins_of_attraction
The computation of basins with the second mode can take a long time depending on the radius parameter ε. 

```
res = 600
ds = Systems.henon(zeros(2); a = 1.4, b = 0.3)
xg = yg = range(-2.,2.,length = res)
grid = (xg,yg)
@time basins, att = basins_of_attraction(grid, ds; show_progress = false)
@time basins, att = basins_of_attraction(grid, ds; attractors = att, show_progress = false )
@time basins, att = basins_of_attraction(grid, ds; attractors = att, ε = 1e-6, show_progress = false )
```

Results are: 

```
@time basins, att = basins_of_attraction(grid, ds; show_progress = false)
 7.703854 seconds (43.50 M allocations: 1.376 GiB, 4.25% gc time, 62.09% compilation time)

@time basins, att = basins_of_attraction(grid, ds; attractors = att, show_progress = false )
123.987298 seconds (286.80 M allocations: 19.574 GiB, 2.83% gc time, 1.35% compilation time)

@time basins, att = basins_of_attraction(grid, ds; attractors = att, ε = 1e-6, show_progress = false )
  1.055172 seconds (4.02 M allocations: 180.549 MiB, 4.70% gc time, 54.13% compilation time)

```

Default radius is 1e-3. My guess is that the nearest neighbor search finds a lot of points for this radius and there are too much memory allocation in basins_low_level.jl : 

```
        for (k, t) in bsn_nfo.search_trees # this is a `Dict`
            idxs = isearch(t, u_full_state, WithinRange(ε))
            if !isempty(idxs)
                nxt_clr = 2*k + 1
                break
            end
        end
```

Is there a way to just stop after finding one index instead of having all of them?

awage:
Fix to reduce the number of allocations in the second mode of operation in `basins_of_attraction`. 

Two new variables have been added to BasinsInfo. 

New default value ε = 1e-5

Closes #238

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ChaosTools"
 uuid = "608a59af-f2a3-5ad4-90b4-758bdf3122a7"
 repo = "https://github.com/JuliaDynamics/ChaosTools.jl.git"
-version = "2.6.4"
+version = "2.6.5"
 
 [deps]
 Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
--- a/src/basins/basins_lowlevel.jl
+++ b/src/basins/basins_lowlevel.jl
@@ -133,7 +133,7 @@ and the trajectories staying outside the grid are coded with -1.
 function _identify_basin_of_cell!(
         bsn_nfo::BasinInfo, n::CartesianIndex, u_full_state;
         mx_chk_att = 2, mx_chk_hit_bas = 10, mx_chk_fnd_att = 100, mx_chk_loc_att = 100,
-        horizon_limit = 1e6, ε = 1e-5,
+        horizon_limit = 1e6, ε = 1e-3,
         mx_chk_lost = isnothing(bsn_nfo.search_trees) ? 20 : 1000,
     )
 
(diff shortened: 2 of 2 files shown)


Datseris:
Why so small ε though?

awage:
Nothing wrong with 1e-3. I was just annoyed by this behavior. 

I solved the problem. There was a break statement in the loop once the attractor was found. The correct way to do is to return the label of the basin. For some reason the program was doing something else. 

closes issue #238 

Performances are smashing, for a 2000x2000 grid: 

```
@time basins, att2 = basins_of_attraction(grid, ds; attractors = att, show_progress = true )
 17.078572 seconds (19.98 M allocations: 862.074 MiB, 1.05% gc time)
```
On my old laptop.

Datseris:
I always wondered why that `break` was there instead of `return` and thought we do some kind of cleanup of the indexing, but guess it was just a mistake xD will merge and tag once tests pass

PR #240, opened 2022-01-11
allow silencing warnings in `fixedpoints`

Datseris:


Datseris:
@KalelR dyca test fail once again. I think we need to download one of the timeseries stored here: https://github.com/JuliaDynamics/JuliaDynamics/tree/master/timeseries (the Roessler one) and perform the test on these timeseries because everytime we change computer and/or random number generation in jUlia version tests fail again. 

(this is unrelated with this PR, but I am commenting them out once more)

KalelR:
> @KalelR dyca test fail once again. I think we need to download one of the timeseries stored here: https://github.com/JuliaDynamics/JuliaDynamics/tree/master/timeseries (the Roessler one) and perform the test on these timeseries because everytime we change computer and/or random number generation in jUlia version tests fail again.

With the Rossler timeseries, the Embedded Rossler tests reach up to 7 wanted eigenvalues (with this condition: `vec(0.999 .< abs.(eigenvals) .< 1.0) .& vec(imag(eigenvals) .== 0)`). They are very close to 1 (up to 9 decimal places). 

I can fix the tests by simply increasing the possible acceptable length to 7. I'm not sure this is quite correct, though. What do you think @Datseris ?

Datseris:
You're using the timeseries at the folder online right? I'm a bit lost, what does the "wanted" eigenvalues refer to? The roessler system is 3D and its chaotic attractor is practically 2D. What does this 7 mean?

KalelR:
> You're using the timeseries at the folder online right? I'm a bit lost, what does the "wanted" eigenvalues refer to? The roessler system is 3D and its chaotic attractor is practically 2D. What does this 7 mean?

For an input time-series in R^N, Dyca should return N eigenvectors in R^N as well, each with its associated eigenvalue.
The number of eigenvalues approximately 1 should be the number of linear differential equations of the dynamical system generating the time-series.

So for the Rossler system we should have two eigenvalues equal to 1, and one different. This is indeed what we see.

For the Embedded Rossler system, I don't know what is expected. In the original paper, they analyse a Rossler "embedded in a 25-dimensional space with additional multiplicative Gaussian noise" . It's not clear to me what they did, to be honest. But they were expecting 2 eigenvalues close to 1, as I understood. The test that is currently written returns more than that, 7 at most. It then fails because it was accepting only up to 4. But I don't think it is the same test that they did, so I'm not sure what the correct number should be.

To be honest, I find the paper to be badly written, and so I can't really "read between the lines" with the little knowledge I have of embedding stuff. I think it'd take some time and testing to really find out what they mean. It might be better to instead try different tests, with some projection plots for instance. I could do it once I have more time. 

So for now, the first test (on the pure Rossler) is working fine, the second isn't, and I don't know how it should be fixed.

Datseris:
Yeah, I'm also extremely unhappy with both the method and the paper. Seems like it really doesn't work well. In anycase, you can safely stop worrying about this for now, I think I'll just remove the embedded DyCa tests.

KalelR:
Yeah, it's a bit of a shame. It seemed like really cool method, but I'm not convinced anymore about it. 
Maybe in the future I'll come back to work on it. If you are okay with this, I'll remove the embedded tests as you said and commit my changes to the first test.

Datseris:
I have already done the changes, don't worry about this anymore!

PR #242, opened 2022-01-13
Update the docstring of `lyapunov_from_data`

onkyo14taro:
Added a statement recommending the addition of random noise when the data is strongly discretized. The description I added is as follows:

> If the data values are very strongly discretized (e.g., integers or floating-point numbers with very small bits), the coarse quantization is likely to result in a zero distance between the two different points compared. If there is at least one pair of points with zero distance, the return value of `lyapunov_from_data` (the estimate of the maximum Lyapunov exponent) will be `-Inf`. If you want to avoid the artifact due to discretization, you can add noise to your data as follows: `data = data .+ 1e-15randn(length(data))`.

### DIFF
--- a/src/chaosdetection/lyapunovs.jl
+++ b/src/chaosdetection/lyapunovs.jl
@@ -557,6 +557,14 @@ If however the `Metric` is `Cityblock()`, calculate
 the absolute distance of *only the first elements* of the `m+k` and `n+k` points
 of `R` (distance ``d_F`` in ref.[^Skokos2016], useful when `R` comes from delay embedding).
 
+If the data values are very strongly discretized (e.g., integers or
+floating-point numbers with very small bits), the coarse quantization is likely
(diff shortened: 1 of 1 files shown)


Datseris:
Thanks @onkyo14taro but I now realize that this is an inappropriate place to put this. This is not specific to lyapunov exponents, but in fact it is relevant for _any_ kind of analysis you do after delay embedding. I'd suggest that you instead made a pull request directly in the documentation of DynamicalSystems.jl. I.e., edit this file: https://github.com/JuliaDynamics/DynamicalSystems.jl/blob/master/docs/src/embedding/reconstruction.md  after the `embed` documentation codeblock. You can also make it a nice tip by writing in the style:

```
!!!! tip "Embedding discretized data values"
    Each sentence starts with four spaces.
    Lalala discretized values
```

onkyo14taro:
Thank you for your review. I will place it in the location you indicated.

PR #246, opened 2022-02-09
Remove some duplicate code from bSTAB implementation

awage:
I was trying to understand the code more in detail and I found a way to reduce the codebase without affecting the performances. 

Have a look a the change. If you think it is not worth it or too risky changing something that works you can close this PR.

### DIFF
--- a/src/basins/basin_fractions_clustering.jl
+++ b/src/basins/basin_fractions_clustering.jl
@@ -19,7 +19,7 @@ describing the trajectory. Initial conditions are sampled from `ics`, which can 
 be a `Dataset` of initial conditions, or a 0-argument function `ics()` that spits out
 random initial conditions. See [`statespace_sampler`](@ref) to generate such functions.
 
-The output `fs` is a dictionary whose keys are the labels given to each attractor, and the 
+The output `fs` is a dictionary whose keys are the labels given to each attractor, and the
 values are their respective fractions. The label `-1` is given to any initial condition
 whose attractor did not match any of the clusters, see description below.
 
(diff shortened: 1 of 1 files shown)


src/basins/basin_fractions_clustering.jl
```diff
@@ -111,48 +111,34 @@ and returns their extracted features in a matrix, with the j-th column containin
...
-        ic = ics()
+    Threads.@threads for i = 1:N
+        ic = _get_next_ic(ics,i)
         feature_array[i] = extract_features(ds, ic, featurizer; kwargs...)
         show_progress && next!(progress)
     end
     return reduce(hcat, feature_array)
 end
 
+_get_next_ic(ics::Function, i) = ics()
```
Datseris, line 131: Maybe name this just `_get_ic`?

src/basins/basin_fractions_clustering.jl
```diff
@@ -111,48 +111,34 @@ and returns their extracted features in a matrix, with the j-th column containin
 j-th feature. To do this, it  calls the other `extract_features` method, made for just one array of ICs.
 `ics` should contain each initial condition along its rows.
 """
```
Datseris, line 113: Let's also completely remove the documentation string of this function.

Datseris (approved):
Yes, perfect, I love this kind of code clarity. Only 2 minor comments.

awage:
Test fails here but it passed on my computer. I will fix it.

awage:
It seems that the test fails sometimes. Even on the master branch. Is there a way to set the random generator seed? Another way is to use more samples but tests are getting very long. I rerun the test and it passes the second time.

awage:
It seems that the test fails sometimes. Even on the master branch. Is there a way to set the random generator seed? Another way is to use more samples but tests are getting very long. I ran the test again and it has succeeded  the second time.

Datseris:
> It seems that the test fails sometimes.

Which tests? Yes, there is a way to set random seeds.

awage:
Sometime `Basin Fractions Clustering Pendulum` fails and also `Correlation Dim`. 

For `Basin Fractions`the results are almost within the bounds but due to the sample it sometimes steps out the interval.

Datseris:
I'll have a look in the other PR.

PR #257, opened 2022-05-19
Recurrence mapper with Sparse matrix

awage:
This is an experimental code to allow to track attractors in high dimensional dynamical systems.

There are some problems that I don't know how to solve: 
* conditional parametric types to allow `SparseArray` or an `Array` for the basins matrix. 
* how to handle the choice of a Sparse Matrix: automatically or with the propagation of a keyword.
* I am not satisfied with the dependence on the `SparseArrayKit.jl`. It is not a mature package but it does the work. Maybe there are better alternatives. 
 
I have a added a test file just for demonstration of the capability of the algorithm. It works in dimension 100 without a flinch...

### DIFF
--- a/src/basins/attractor_mapping_featurizing.jl
+++ b/src/basins/attractor_mapping_featurizing.jl
@@ -21,8 +21,9 @@ end
 
 """
     AttractorsViaFeaturizing(
-        ds::DynamicalSystem, featurizer::Function, clusterconfig::ClusteringConfig;
-    kwargs...)
+        ds::DynamicalSystem, featurizer::Function, 
+        clusterconfig = ClusteringConfig(); kwargs...
+    )
 
 Initialize a `mapper` that maps initial conditions to attractors using the featurizing and
 clustering method of [^Stender2021]. See [`AttractorMapper`](@ref) for how to use the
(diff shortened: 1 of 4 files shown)


src/basins/attractor_mapping_recurrences.jl
```diff
@@ -1,3 +1,5 @@
+using SparseArrayKit
```
Datseris, line 1: SparseArrayKit is used but it is not in Project.toml, hence this won't actually work.

src/basins/attractor_mapping_recurrences.jl
```diff
@@ -144,7 +146,8 @@ end
 # Definition of `BasinInfo` and initialization
 #####################################################################################
 mutable struct BasinsInfo{B, IF, D, T, Q}
-    basins::Array{Int16, B}
+    #basins::Array{Int16, B}
+    basins::SparseArray{Int16, B}
```
Datseris, line 150: This is not the way to go for type stability. The type of the array needs to be parameterized by its own. In `BasinInfo{..}` add one more type parameter `A <: AbstractArray`. Then you can write `basins::A{Int16, B}` in the field.

Datseris: In addition, `A` should be a type parameter before `B` in the sequence `{A, B, IF...}`

Datseris (commented):
I think we need to be a bit careful with how much we mix the methods. In the sparse version it is important that we DO NOT actually assing basin values in the sparse array in the full `basins_of_attraction` function. Otherwise it defeats the purpose.

Datseris:
can confirm, seems to work, but we have to come up with examples that aren't trivial. Seems to me that this setting of the Kuramoto has a huuuuuge amount of fixed points as its attractors?

awage:
The Kuramoto oscillator has a lot of symmetries and many attractors are basically the same with phases switched. It has no practical interest for this method. I will look for more suitable examples this week.

awage:
@Datseris some wonky stuff happened here. I was updating my local branch to the upstream branch and github merged the all stuff without asking. The strange thing is that it didn't merged any commits. I will reopen a new PR with the changes already made from the branch stored in my computer.

PR #263, opened 2022-07-20
add dt reset in reinit

IBArbitrary:
This is in response to issue [#187](https://github.com/JuliaDynamics/DynamicalSystems.jl/issues/187) from DynamicalSystems.jl package. It has been a while since the original issue poster responded, so I stepped-up, as a similar issue is also impeding my work. Cheers!


Datseris:
Hello thanks! However you also need to add a test in the test folder that checks exactly what the user reported in #187 . This is called a "regression test": whenever a bug is fixed, a test that tests against this bug should be added.

Datseris:
Notice that now all functionality related with attractors/basins lives in Attractors.jl. You need to reopen this pull request there: https://github.com/JuliaDynamics/Attractors.jl

PR #266, opened 2022-08-05
Generic type for basins in `basins_of_attraction`

KalelR:
A solution to the problem of arbitrary labels in the featuring supervised method


KalelR:
Sorry if I messed something up with the commits here, don't know why it's ahead of main...

This change initializes `basins` with `Any`type, which then allows it to be filled with the generic labels for the supervised method. In all other cases however it will still be returned as an Array of `Any`, when it could be the more specific `Int` type.
Is that a problem?

Datseris:
@KalelR I suspect the reason the git is a mess is because you are not branching off `main`. you are branching off other branches that are later merged into `main`. So git doesn't register the commit differences. After a PR is merged, best practice for you, assuming your PRs come from **branches of the origin**, and not from your own fork: (1) checkout main locally, (2) sync main with the remote main, that has your PR merged in, (3) start a new branch after you have updated your local main.

src/basins/attractor_mapping.jl
```diff
@@ -134,7 +134,7 @@ the partitioning happens directly on the hyperplane the Poincaré map operates o
 corresponding to the state space partitioning indicated by `grid`.
 """
 function basins_of_attraction(mapper::AttractorMapper, grid::Tuple; kwargs...)
-    basins = zeros(Int16, map(length, grid))
+    basins = Array{Any}(undef, map(length, grid))
```
Datseris, line 137: No, we can't do that. This is terrible for performance. You need to identify the type correctly from `mapper` here. Define a convenience function `basin_key_type(mapper) = Int16`. For the proximity and featurizer methods, define a new method that actually uses the `tempaltes` or the `attractors` field of these mappers, and uses the same key type as these dictionaries.

Datseris (changes requested):

Datseris:
Also, by trying to update the this PR to `main`, I removed your commit in the test file that removed the `sort!` function, sorry about that! Please do go ahead and restore this change!

src/basins/attractor_mapping.jl
KalelR: Ok, good idea. I have one question on how to implement this though.

In this case the labels for both  `Proximity`and the supervised `Featurizing` are going to be generic, dependent on the user input. But for both we are currently using `-1` to label unclustered points. How do we return a  `-1` with a generic type? For example, if the labels were strings, this would be easy: `string(-1)`. But what about `Symbol` or something else?  I don't know how we'd convert to generic types.

Datseris: Shit. I have not thought about this.

Datseris:
@KalelR okay, I resetted after a nice sleep and rest. My idea to allow generic types was stupid. It takes much more effort from our side. From a user perspective, they can always define a new dictionary mapping their labels to integers. Our code should always operate on integers like before. Could you please update this pr to **make sure our code uses only integers**? I am really sorry to make you do double work, but I thought this through and what we are trying to do doesn't really align with the principles of DynamicalSystems.jl, where things should really only be done one way and the user should always convert to this one way.

KalelR:
> @KalelR okay, I resetted after a nice sleep and rest. My idea to allow generic types was stupid. It takes much more effort from our side. From a user perspective, they can always define a new dictionary mapping their labels to integers. Our code should always operate on integers like before. Could you please update this pr to **make sure our code uses only integers**? I am really sorry to make you do double work, but I thought this through and what we are trying to do doesn't really align with the principles of DynamicalSystems.jl, where things should really only be done one way and the user should always convert to this one way.

That's alright. It would have been a nice feature but unfortunately it is not useful enough for the amount of effort needed.
I now resetted the code for `basins_of_attraction` so that `basins` is of `Int` type. I am keeping the previous addition to `cluster_features_distances`, with the correct replacing of labels. With this, `basins_fractions` still works with arbitrary labels in the `templates` (in the supervised method). And `basins_of_attractions` will also return the correct labels for any `Int` labels the user passes. Should we state in the documentation that the labels **must** be of `Int` type, to avoid confusion? They only really need to be for  `basins_of_attraction`.

Datseris:
Yes, state in the doc that labels must be of type `Int`.

Datseris:
Also please add a default value for clustering that was like our previous default keywords for clustering. I think you just need to write
```
AttractorsViaFeaturizing(ds::GeneralizedDynamicalSystem, featurizer::Function,
    cluster_config::ClusteringConfig = ClusteringConfig()
```
in the function definition.

src/basins/attractor_mapping_featurizing.jl
```diff
@@ -62,8 +62,8 @@ representation of features.
     stability of multi-stable dynamical systems](https://doi.org/10.1007/s11071-021-06786-5)
 """
 function AttractorsViaFeaturizing(ds::GeneralizedDynamicalSystem, featurizer::Function,
-    cluster_config::ClusteringConfig; T=100, Ttr=100, Δt=1, diffeq = NamedTuple(),
-    attractors_ic::Union{AbstractDataset, Nothing}=nothing)
+    cluster_config::ClusteringConfig = ClusteringConfig(); T=100, Ttr=100, Δt=1,
+    diffeq = NamedTuple(), attractors_ic::Union{AbstractDataset, Nothing}=nothing)
```
Datseris, line 66: @KalelR **remember what I tell you about documentation strings!!! Don't push changes that are not reflected to the docstring**.  You can do the call signature like this:
```
    AttractorsViaFeaturizing(
        ds::DynamicalSystem, featurizer::Function, 
        clusterconfig = ClusteringConfig(); kwargs...
    )
```

KalelR: goddamn you are right,  sorry for messing up the doc again, super bad habit I have of not paying enough attention to it

PR #269, opened 2022-08-10
Recurrence method for finding attractors with sparse matrix 

awage:
This is a new PR with the same code as #257. I will keep working on the tests for this new mode.


test/basins/attractor_mapping_tests.jl
```diff
@@ -161,7 +161,7 @@ end
         1 => [-0.8, 0],
         2 => [1.8, 0],
     ]
-    expected_fs_raw = Dict(2 => 0.511, 1 => 0.489)
+    expected_fs_raw = Dict(2 => 0.509, 1 => 0.491)
```
Datseris, line 164: If we constantly have to change these numbers to ensure tests pass, then we aren't actually writing good tests. We should write tests that we can be confident in without having to change the tests themselves. I think we need to allow the testing function to have a margin of error when checking the fractions.

awage: I have been thinking about this behavior and I have no ready response. Anyway I will increase the error margin for this example.

awage:
The code is ready. There is still the problem with the duffing oscillator test that I don't know how to track. It is maybe a problem with the dependencies. I will spend some time next week to dig this problem.

Project.toml
```diff
@@ -20,6 +20,7 @@ Neighborhood = "645ca80c-8b79-4109-87ea-e1f58159d116"
 ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Roots = "f2b01f46-fcfa-551c-844a-d8ac1e96c665"
+SparseArrayKit = "a9a3c162-d163-4c15-8926-b8794fbefed2"
```
Datseris, line 23: You need to add a compatibility version here as well, later down in the Project.toml.

src/basins/attractor_mapping_recurrences.jl
```diff
@@ -143,8 +154,8 @@ end
 #####################################################################################
 # Definition of `BasinInfo` and initialization
 #####################################################################################
-mutable struct BasinsInfo{B, IF, D, T, Q}
-    basins::Array{Int16, B}
+mutable struct BasinsInfo{B, IF, D, T, Q, A <: AbstractArray}
+    basins::A # A always have type parameterization {Int16, B}. Sparse or Array
```
Datseris, line 158:
```suggestion
mutable struct BasinsInfo{B, IF, D, T, Q, A <: AbstractArray{Int16, B}}
    basins::A # sparse or dense
```
if we can write the same thing with proper code instead of comments, we should prefer the code.

test/basins/attractor_mapping_tests.jl
```diff
@@ -154,6 +157,7 @@ end
     diffeq = (alg = Vern9(), reltol = 1e-9, abstol = 1e-9)
     T = 2π/1.0
     ds = stroboscopicmap(ds, T; diffeq)
+    # ds = stroboscopicmap(ds, T)
```
Datseris, line 160: Delete, or replace. We shouldn't have commented but runnable code hanging around.

Datseris (commented):
Great! I only have very minor comments. One thing I am thinking: do you think it's worth defining
```
function AttractorsViaRecurrencesSparse(...)
```
which just propagates `sparse = true` to the previous version?

---

There is one more thing: if `sparse` is enabled, the specialized version of `basins_of_attraction(mapper::AttractorsViaRecurrences)` needs to throw an error.

PR #274, opened 2022-08-20
Finalize recurrences seeded continuation

Datseris:
Will fix #273. Opening the PR already so it is easier.

PR #275, opened 2022-08-30
Pr more continuation tests

awage:
This is a PR to the PR #274 

Added a new discrete system: Newton algorithm to find complex roots in the plane. I changed the polynomial to move the roots locus smoothly on the plane. 

Added the second order coupled kuramoto problem: It works but it is slow and difficult to define good tests.


src/basins/mapping/attractor_mapping_recurrences.jl
```diff
@@ -312,8 +312,8 @@ function get_label_ic!(bsn_nfo::BasinsInfo, integ, u0; safety_counter_max = Int(
         # within the same grid cell. In such a case, when starting on the second attractor
         # the trajectory will forever reset between locating a new attractor and recurring
         # on the previously found one...
-        bsn_nfo.unsafe || (bsn_nfo.safety_counter += 1)
-        if bsn_nfo.unsafe || (bsn_nfo.safety_counter ≥ safety_counter_max)
+        !bsn_nfo.unsafe && (bsn_nfo.safety_counter += 1)
+        if !bsn_nfo.unsafe && (bsn_nfo.safety_counter ≥ safety_counter_max)
```
Datseris, line 316: FYI these changes are equivalent, the code doesn't change

awage:
I think that we can ditch this PR, the tests with the Newton method are not necessary.

PR #277, opened 2022-09-08
Add Ttr keyword to mapping recurrences.

awage:
This PR adds the keyword `Ttr` to `AttractorsViaRecurrences`. It allows to skip a transient time before starting the recurrence routine. This is useful for high dimensional systems where the sparse matrix operations are slow.


Datseris:
nice and straightforward

PR #278, opened 2022-09-18
Fix return times function for continuous systems

Datseris:
This PR aims to:

1. Improve source code clarity of return times for discrete systems
1. Overall better organize tests for return times
1. Fix return times for continuous systems, which never really worked well and was also super slow. To this end we have
    * Re-written the source code from scratch.
    * Offer two algoritms: linear interpolation, or integrator-based interpolation with minimization for finding closest point and then rootfinding for finding the crossings.
    * Linear interpolation works only for spheres (in the future one could contribute code for rectangle crossing I guess.
    * High precision interpolation uses Optim.jl and Roots.jl.
    * Serious tests are now also done for continuous systems.
1. Performance increase for continuous systems: TODO; write by how much.

Only todo left is to make it work for hyperrectangles in continuous systems.


src/rareevents/mean_return_times/mrt_continuous.jl
```diff
@@ -0,0 +1,455 @@
...
+    # use Optim.jl to find minimum of the function
+    f = (t) -> evaluate(metric, integ(t), u0)
+    # Then find minimum of `f` in limits `(tprev, t)`
+    optim = Optim.optimize(
+        f, integ.tprev, integ.t, Optim.Brent();
+        store_trace=false, abs_tol = method.abstol, rel_tol = method.reltol,
+    )
+    tmin, dmin = Optim.minimizer(optim), Optim.minimum(optim)
+    return tmin, dmin
+end
```
Datseris, line 191: @KalelR here is the code snippet

PR #289, opened 2022-10-17
Use set_state! in lyapunov rescale!

### LINKED ISSUE #213: Re-write source code of `lyapunov` to use `set_state!`
The source code of `lyapunov` was one of the very first functions of DynamicalSystems.jl. It does _not_ use the `set_state!` function for parallel integrators. That is a mistake (and in fact made us miss a bug).

One should re-write the `rescale!` function of the code to simply generate the new reference state and the just call `set_state!`.

xlxs4:
Closes #213. From the issue:

> we can do some optimization so that we have a dummy preinitialized r that we carry around and modify its value in place so that we don't create a new r all the time

Open to any pointers.

---

Tests pass locally:

image

### DIFF
--- a/src/chaosdetection/lyapunovs/lyapunov.jl
+++ b/src/chaosdetection/lyapunovs/lyapunov.jl
@@ -157,26 +157,7 @@ function λdist(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}) where {Alg, I
     return norm(integ.u[1] - integ.u[2])
 end
 
-# Rescales:
-function rescale!(integ::MinimalDiscreteIntegrator{true, Vector{S}}, a) where {S<:SVector}
-    integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
-end
-function rescale!(integ::MinimalDiscreteIntegrator{true, Vector{S}}, a) where {S<:Vector}
-    @. integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
-end
-function rescale!(integ::AbstractODEIntegrator{Alg, IIP, M}, a) where {Alg, IIP, M<:Matrix}
-    for i in 1:size(integ.u)[1]
-        integ.u[i, 2] = integ.u[i,1] + (integ.u[i,2] - integ.u[i,1])/a
-    end
-    u_modified!(integ, true)
-end
-function rescale!(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}, a) where {Alg, IIP, S<:Vector}
-    @. integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
-end
-function rescale!(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}, a) where {Alg, IIP, S<:SVector}
-    integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
+function rescale!(integ, a)
+    r = get_state(integ, 1) .+ (get_state(integ, 2) .- get_state(integ, 1)) ./ a
+    set_state!(integ, r, 2)
 end


src/chaosdetection/lyapunovs/lyapunov.jl
```diff
@@ -157,26 +157,7 @@ function λdist(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}) where {Alg, I
...
-function rescale!(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}, a) where {Alg, IIP, S<:Vector}
-    @. integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
-end
-function rescale!(integ::AbstractODEIntegrator{Alg, IIP, Vector{S}}, a) where {Alg, IIP, S<:SVector}
-    integ.u[2] = integ.u[1] + (integ.u[2] - integ.u[1])/a
-    u_modified!(integ, true)
+function rescale!(integ, a)
+    r = get_state(integ, 1) .+ (get_state(integ, 2) .- get_state(integ, 1)) ./ a
+    set_state!(integ, r, 2)
```
Datseris, line 162: The problem here is that this allocates an extra vector every time it's called, as it first creates a new vector and then sets the state of the vector instead of modifying it in place.

what happens if instead of `r = ...` you do `get_state(integ, 2) .= ...`

Datseris:
Ah. also, there is an ongoing PR that tries to fix the same thing but turned out to be more complicated: https://github.com/JuliaDynamics/ChaosTools.jl/pull/214

src/chaosdetection/lyapunovs/lyapunov.jl
Datseris: (you will need two versions here, one for the in place and one for the out of place, because out of place can't do `.=`)

Datseris: So, for the out-of-place, the code you have written now is perfectly fine. You just need one more version for the in-place.

xlxs4: Makes perfect sense. From what I can gather, the out-of-place is needed when `u` is immutable (`SVector`). What should I dispatch on?

Datseris: It's probably simpler to define `rescale!` that calculates `iip = isinplace(integ)` and then has an `if` clause that calls a `rescale_oop(...)` and a `rescale_iip!(...)`. The `iip` is compile-time deduced so this if clause will be compiled away.

You should nevertheless benchmark `lyapunov` before and afte this Pr to be sure that we are not having any performance deficits.

xlxs4: So I'm guessing something like:
```julia
function rescale!(integ, a)
    iip = isinplace(integ)
    if iip
        rescale_iip!(integ, a)
    else
        rescale_oop(integ, a)
    end
end
function rescale_iip!(integ, a)
    get_state(integ, 2) .= get_state(integ, 1) .+ (get_state(integ, 2) .- get_state(integ, 1)) ./ a
end
function rescale_oop(integ, a)
    r = get_state(integ, 1) .+ (get_state(integ, 2) .- get_state(integ, 1)) ./ a
    set_state!(integ, r, 2)
end
```

I haven't figured out yet exactly what goes wrong, but:

1. I'm getting some `MethodError: no method matching inplace(XYZIntegrator)` errors, for example in https://github.com/JuliaDynamics/ChaosTools.jl/blob/main/test/chaosdetection/lyapunov_exponents.jl#L119 it's `isinplace(MinimalDiscreteIntegrator)`, or `isinplace(SimpleATsit5Integrator)` in other tests. I guess I can do something like
    ```julia
    function isinplace(integ::AbstractODEIntegrator{Alg, IIP, S}) where {Alg, IIP, S}
        return IIP
    end
    function isinplace(integ::MinimalDiscreteIntegrator{IIP, S}) where {IIP, S}
        return IIP
    end
    ```
    ?
2. I don't think `isinplace` works as intended currently, as there are some
    ```
    setindex!(::SVector{3, Float64}, value, ::Int) is not defined.
    Hint: Use `MArray` or `SizedArray` to create a mutable static array
    ```
    errors, it seems like `iip` evaluates to `True`.
    
What am I missing?

Datseris: For your (1), yes. For (2), hm, are you sure? Can you push the code that leads to this problem?

xlxs4: Of course

Datseris: I am resolving this conversation for now, as the original comment has been addressed.

Datseris:
I'll have a look but I can't promise when (same thing with the #214 ) , because I have many projects going on :( But the code here looks like a simpler change so I think I'll get to this quickly enough!

PR #297, opened 2023-02-09
Add base function for extreme events

SergeyNovak777:
I added base functions for extreme events, but don't know where paste `using Packages`.  It paste in Base_function.jl or ChaosTools.jl ?

### DIFF
--- a/src/ChaosTools.jl
+++ b/src/ChaosTools.jl
@@ -46,6 +46,8 @@ include("chaosdetection/expansionentropy.jl")
 include("chaosdetection/partially_predictable.jl")
 include("chaosdetection/testchaos01.jl")
 
+include("eva/Base_function.jl")
+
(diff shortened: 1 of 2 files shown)


Datseris:
But at the moment these functions do not do anything else besides counting how often a timeseries exceeds its 3 std value. I don't think this functionality is advanced enough to be included here. If you use it as a basis for more advanced algorithms, then you should put them in this PR as well.

Datseris:
I don't challenge that this is not used in the literature. Of course, finding when a timeseries exceeds a threshold is used. What I challenge is whether it is of any use to put this in a package. It seems to me that this code is simple enough that a user is better off writing it down themselves than trying to find it in a package.

Can we built on this code? Is there anything more advanced we can do once we have points the timeseies exceeds a threshold?

PR #312, opened 2023-08-16
update `predictability` to DynamicalSystems.jl v3.0.

rusandris:
Hi everyone!
I've started to update `predictability` to DynamicalSystems.jl v3.0, this being the first batch of changes. It seems to be working, all tests pass. There are some other TODOs here in there in the comments, so further improvements can be made regarding code readibility/design etc. Let me know what you think.

### DIFF
--- a/src/chaosdetection/partially_predictable.jl
+++ b/src/chaosdetection/partially_predictable.jl
@@ -80,9 +80,7 @@ function predictability(ds::CoreDynamicalSystem;
         ν_threshold = 0.5, C_threshold = 0.5
     )
 
-    #error("This function has not yet been updated to DynamicalSystems.jl v3.0.
-    #Please consider a Pull Request :) (very easy!)")
-
+	reinit!(ds)
     rng = Xoshiro()
     λ_max < 0 && return :REG, 1.0, 1.0
     samples = sample_trajectory(ds, Ttr, T_sample, n_samples)
@@ -94,9 +92,7 @@ function predictability(ds::CoreDynamicalSystem;
     # Calculate cross-distance scaling and correlation scaling
     distances = Float64[] # Mean distances at time T for different δ
     correlations = Float64[] # Cross-correlation at time T for different δ
-    # TODO: p_integ is type unstable, the rest of the code should be moved
-    # into a separate function. It's also good todo for more readable code...
-    p_sys = ParallelDynamicalSystem(ds, samples[1:2])
+    
     for δ in δ_range
         # TODO: some kind of warning should be thrown for very large Tλ
         Tλ = log(d_tol/δ)/λ_max
@@ -173,7 +169,7 @@ function sample_trajectory(ds::DynamicalSystem,
     samples = typeof(current_state(ds))[]
     while current_time(ds) < Ttr + T_sample
         step!(ds, rand(D_sample))
-        push!(samples, current_state(ds))
+        push!(samples, deepcopy(current_state(ds)))
     end
     samples
 end
--- a/src/chaosdetection/partially_predictable.jl
+++ b/src/chaosdetection/partially_predictable.jl
@@ -92,7 +92,7 @@ function predictability(ds::CoreDynamicalSystem;
     # Calculate cross-distance scaling and correlation scaling
     distances = Float64[] # Mean distances at time T for different δ
     correlations = Float64[] # Cross-correlation at time T for different δ
-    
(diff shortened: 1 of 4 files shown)


test/chaosdetection/partially_predictable.jl
```diff
@@ -4,15 +4,14 @@ using Random
...
-diffeqmaxit = (maxiters = 1e9,)
 println("\nTesting Partially predictable chaos...")
 
 @testset "Predictability Lorenz" begin
     @testset "strongly chaotic" begin
         Random.seed!(12)
         lz = Systems.lorenz(ρ=180.70)
         @time chaos_type, ν, C = predictability(lz;
-            λ_max = 1.22, diffeq=diffeqmaxit, T_max = 1e3, n_samples = 100,
+            λ_max = 1.22, T_max = 1e3, n_samples = 100,
```
Datseris, line 14: Here you removed the `diffeqmaxit` but you don't give it as the `diffeq` keyword when creating the Lorenz system. SInce you are on this, please **do not use `Systems.`**. We have moved away from using the systems module for predefined systemsin the test suite. Please make the lorenz system from scratch by writing down the equations function and creating a `CoupledODEs`. I nthe creation you can also give `diffeqmaxit`.

src/chaosdetection/partially_predictable.jl
```diff
@@ -74,15 +77,15 @@ function predictability(ds::CoreDynamicalSystem;
...
         λ_max::Real = lyapunov(ds, 5000), d_tol::Real = 1e-3, T_multiplier::Real = 10,
         T_max::Real = Inf, δ_range::AbstractArray = 10.0 .^ (-9:-6),
-        ν_threshold = 0.5, C_threshold = 0.5,
+        ν_threshold = 0.5, C_threshold = 0.5
     )
 
-    error("This function has not yet been updated to DynamicalSystems.jl v3.0.
-    Please consider a Pull Request :) (very easy!)")
+    #error("This function has not yet been updated to DynamicalSystems.jl v3.0.
+    #Please consider a Pull Request :) (very easy!)")
```
Datseris, line 84: delete all comments that don't help wit making the sourc code readable

Datseris (commented):
yeap, seems straightforward change.

Datseris:
also increment patch version in project toml

Datseris:
oh yerah you also have to actually enable the test file: https://github.com/JuliaDynamics/ChaosTools.jl/blob/main/test/runtests.jl#L25

test/chaosdetection/partially_predictable.jl
rusandris: I tried to use user-defined systems instead of `Systems` and now it's broken. 
Can you tell me what the correct way is to `reinit!` an iip `ParallelDynamicalSystem`? This doesn't work:
```
@inbounds function lorenz_rule!(du, u, p, t)
    σ = p[1]; ρ = p[2]; β = p[3]
    du[1] = σ*(u[2]-u[1])
    du[2] = u[1]*(ρ-u[3]) - u[2]
    du[3] = u[1]*u[2] - β*u[3]
    return nothing
end

u0 = [0, 10.0, 0]
p0 = [10, 28, 8/3]
diffeq = (maxiters = 1e9,abstol = 1.0e-6, reltol = 1.0e-6)

lz = CoupledODEs(lorenz_rule!, u0, p0; diffeq)
p_sys = ParallelDynamicalSystem(lz, [u0,u0 .+ 0.1])
reinit!(p_sys, [u0,u0 .+ 0.1])
```
The docs say `reinit!` takes in a vector of states (like states) for u. But it only works when using `hcat`
`reinit!(p_sys, hcat(u0,u0 .+ 0.1))`
but that doesn't work for oop systems.

Datseris: Thanks, this was a bug. The fix is here: https://github.com/JuliaDynamics/DynamicalSystemsBase.jl/pull/181 

once merged you can continue with this PR!

rusandris:
Hi! It seems like it's close to completion with your additional bugfixes. I've also added a `reinit!` at the first line so we don't run into problems when `predictability` needs to run multiple times (it is a mutating function). There are also some little things here and there, let me know if it's okay.

rusandris:
Also enabled the tests in `test/run_tests.jl`. Also I noticed that I messed up the commit message. Is it editable easily afterwards?

PR #325, opened 2024-03-15
Improve UPO storage in `periodicorbits.jl`

JonasKoziorek:
I noticed that while using periodicorbits algorithm with parameter roundtol that is high (eg. 8 ), it detects many duplicate UPOs.
MVE:
```
using ChaosTools
using PredefinedDynamicalSystems: henon
begin
    ds = henon()
    xs = LinRange(-3.0, 3.0, 60)
    ys = LinRange(-10.0, 10.0, 60)
    seeds = [SVector{2}(x,y) for x in xs for y in ys]
    o = 5
    indss, signss = lambdaperms(dimension(ds))
    λ = 0.005
    results1 = sort!(periodicorbits(ds, o, seeds, λ, indss, signss; roundtol = 8))
end
```
Output:
```
10-element Vector{SVector{2, Float64}}:
 [-1.13135448, -0.33940634]
 [-1.13135448, -0.33940633]
 [-1.13135447, -0.33940636]
 [-1.13135447, -0.33940635]
 [0.63135447, 0.18940633]
 [0.63135447, 0.18940634]
 [0.63135448, 0.18940634]
 [0.63135448, 0.18940635]
 [0.63135449, 0.18940636]
 [0.63135449, 0.18940636]
 ```
My proposed change solves this:
```
using ChaosTools
using PredefinedDynamicalSystems: henon
begin
    ds = henon()
    xs = LinRange(-3.0, 3.0, 60)
    ys = LinRange(-10.0, 10.0, 60)
    seeds = [SVector{2}(x,y) for x in xs for y in ys]
    o = 5
    indss, signss = lambdaperms(dimension(ds))
    λ = 0.005
    results2 = sort!(periodicorbits(ds, o, seeds, λ, indss, signss; spacetol = 1e-7))
end
```
Output:
```
julia> results2
2-element Vector{SVector{2, Float64}}:
 [-1.1313544826122608, -0.33940632522716674]
 [0.6313544679502884, 0.18940632554538597]
 ```

I have implemented `FP` as a binary tree. This way,  new fixed points are inserted into `FP` only if they are `spacetol` away from the remaining fixed points in `FP`. Thus duplicates are avoided for a good choice of `spacetol`.

Additionally, binary tree removes the need for linear search through `FP` even though this doesn't seem to be a bottleneck.

Let me know if there is something to fix. This is my first PR :).

### DIFF
--- a/src/periodicity/custombintree.jl
+++ b/src/periodicity/custombintree.jl
@@ -1,0 +1,35 @@
+using DataStructures: RBTree
+import Base: <, ==
+
+struct DummyStructure{T}
+    value::T
+    eps::Float64
+end
+
+function <(a::DummyStructure, b::DummyStructure)
+    abs_diff = norm(a.value - b.value)
+    return abs_diff > a.eps && a.value < b.value
+end
+
+function ==(a::DummyStructure, b::DummyStructure)
+    return norm(a.value - b.value) <= a.eps
+end
+
+function tovector(tree::RBTree{DummyStructure{T}}) where T
+    len = length(tree)
+    type = typeof(tree).parameters[1].parameters[1]
+    vect = Vector{type}(undef, len)
+    index = 1
+    _tovector!(tree.root, vect, index)
+    return vect
+end
+
+function _tovector!(node, vect::Vector{T}, index::Int64) where T
+    if !isnothing(node.data)
+        index = _tovector!(node.leftChild, vect, index)
+        vect[index] = node.data.value
+        index += 1
+        index = _tovector!(node.rightChild, vect, index)
+    end
+    return index
+end
--- a/src/periodicity/periodicorbits.jl
+++ b/src/periodicity/periodicorbits.jl
@@ -1,5 +1,6 @@
 using Combinatorics: permutations, multiset_permutations
 using Random: randperm
+using DataStructures: RBTree
 
 export lambdamatrix, lambdaperms, periodicorbits
 
@@ -30,11 +31,10 @@ a random permutation will be chosen for them, with `λ=0.001`.
    the next one is `≤ disttol` then it has converged to a fixed point.
 * `inftol = 10.0`: If a state reaches `norm(state) ≥ inftol` it is assumed that
    it has escaped to infinity (and is thus abandoned).
-* `roundtol::Int = 4`: The found fixed points are rounded
-   to `roundtol` digits before pushed into the list of returned fixed points `FP`,
-   *if* they are not already contained in `FP`.
-   This is done so that `FP` doesn't contain duplicate fixed points (notice
-   that this has nothing to do with `disttol`).
+* `spacetol = 1e-8`: A detected fixed point is stored if it is at least `spacetol` 
+   far from all other stored fixed points `FP`. Distance is measured by euclidian norm.
+   This limits the number of duplicate fixed points in the output `FP`. If you are 
+   getting duplicate fixed points, decrease this value.
 
 ## Description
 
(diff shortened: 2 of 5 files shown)


src/periodicity/custombintree.jl
Datseris, line 1: This file has confused me a bit... So Datastructures provides a binary tree. What is the `DummyStructure` supposed to do? WHy do we need this?

src/periodicity/periodicorbits.jl
```diff
@@ -67,15 +67,16 @@ function periodicorbits(
         maxiters::Int = 100000,
         disttol::Real = 1e-10,
         inftol::Real = 10.0,
-        roundtol::Int = 4
+        spacetol::Real = 1e-8,
```
Datseris, line 70: Right, notice that this is a breaking change right now. You have removed a keyword, so users that were using this code will be getting an error. They should be getting a warning instead.

Do `roundtol = nothing`
and inside the function have a warning code:
```
if !isnothing(roundtol)
warn("`roundtol` keyword has been removed in favor of `spacetol`")
end
```

by the way, why the name `spacetol`? This is an absolute tolerance right? perhaps `abstol` is better?

src/periodicity/custombintree.jl
JonasKoziorek: Typically binary tree will store value to the left child if it is less than root and to the right side if it is more. I want to store value to the left only if it at least `DymmyStructure.eps` away from the root. Hence duplicate periodic points that are close closer than `DummyStructure.eps` won't be stored. DummyStructure is needed for `==` and `<` overload so that it can be used in the binary tree from Datastructures. Perhaps there is a more elegant way to do it. I agree that it is confusing.

src/periodicity/periodicorbits.jl
JonasKoziorek: You are right. I added warning. `abstol` would stand for absolute tolerance. I renamed it as `absintol` for absolute intolerance, which is what is actually is: dropping results in `absintol` neighborhood.

src/periodicity/custombintree.jl
```diff
@@ -0,0 +1,35 @@
+using DataStructures: RBTree
+import Base: <, ==
+
+struct DummyStructure{T}
```
Datseris, line 4: Okay, please add comments here explaining why this exists (copy paste the explanation you gave in the GitHub discussion)

src/periodicity/periodicorbits.jl
```diff
@@ -30,11 +31,9 @@ a random permutation will be chosen for them, with `λ=0.001`.
    the next one is `≤ disttol` then it has converged to a fixed point.
 * `inftol = 10.0`: If a state reaches `norm(state) ≥ inftol` it is assumed that
    it has escaped to infinity (and is thus abandoned).
-* `roundtol::Int = 4`: The found fixed points are rounded
-   to `roundtol` digits before pushed into the list of returned fixed points `FP`,
-   *if* they are not already contained in `FP`.
-   This is done so that `FP` doesn't contain duplicate fixed points (notice
-   that this has nothing to do with `disttol`).
+* `absintol = 1e-8`: A detected fixed point isn't stored if it is in `absintol`
```
Datseris, line 34: Alright, then rename this to `abstol`. it is the same as `abstol` in the `isapprox` function so it will make it easier to approach for a user.

src/periodicity/periodicorbits.jl
```diff
@@ -85,17 +89,20 @@ function periodicorbits(
...
-        roundtol::Int = 4
+        absintol::Real = 1e-8, # absolute intolerance
     )
+    if !isnothing(roundtol)
+        warn("`roundtol` keyword has been removed in favor of `absintol`")
+    end
 
-    FP = typeof(current_state(ds))[]
+    type = typeof(current_state(ds))
+    FP = RBTree{DummyStructure{type}}()
```
Datseris, line 99: Here in the FP write a couple of lines of code why you use the tree. The same explanation you gave to me.

Datseris (approved):
All in all good, I have some standind comments for adding comments in the source code to increase clarity for the non obvious additions.

Datseris (approved):
Perfect. Please rename `DummyStructure` to `NumberWithEpsRadius`, which is a much more indicative name of the data structure. Then I'll merge. Would be nice to add a changelog entry and bumb the _minor_ version in project toml. We will release this along with your other PR.

PR #326, opened 2024-03-17
Support for fixed points of n-th order in `fixedpoints`

JonasKoziorek:
I have added support for computing fixed points of n-th iteration of a discrete dynamical system.

### DIFF
--- a/src/stability/fixedpoints.jl
+++ b/src/stability/fixedpoints.jl
@@ -111,15 +111,13 @@ function to_root_J(Jf, ds::DeterministicIteratedMap, p, o::Int)
     c = Diagonal(ones(typeof(current_state(ds))))
     d = dimension(ds)
     u -> begin
-        trajectory = Vector{Any}(undef, o) # has to be any to work with IntervalRootFinding
-        trajectory[1] = u
+        J = Jf(u, p, 0.0)
+        x = ds.f(u, p, 0.0)
         for i in 2:o
-            trajectory[i] = ds.f(trajectory[i-1], p, 0.0) # trajectory from DynamicalSystemsBase won't work because of IntervalRootFinding
+            J = Jf(x, p, 0.0) * J
+            x = ds.f(x, p, 0.0)
         end
-        reverse!(trajectory)
-        jacobians = map(x -> Jf(x, p, 0.0), trajectory)
-        jacob_of_composition = reduce(*, jacobians)
-        return jacob_of_composition .- c
+        return J .- c
     end
 end
 
--- a/src/stability/fixedpoints.jl
+++ b/src/stability/fixedpoints.jl
@@ -37,14 +37,29 @@ as a start of a continuation process. See also [`periodicorbits`](@ref).
   see the docs of IntervalRootFinding.jl for all possibilities.
 - `tol = 1e-15` is the root-finding tolerance.
 - `warn = true` throw a warning if no fixed points are found.
+- `order = nothing` search for fixed points of the n-th iterate of 
+  [`DeterministicIteratedMap`](@ref). Must be a positive integer or `nothing`.
+  Select `nothing` or 1 to search for the fixed points of the original map.
+
+## Performance notes
+
+Setting `order` to a value greater than 5 can be very slow. Consider using 
+more suitable algorithms for periodic orbit detection, such as 
+[`periodicorbits`](@ref).
+
 """
 function fixedpoints(ds::DynamicalSystem, box, J = nothing;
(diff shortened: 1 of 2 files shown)


src/stability/fixedpoints.jl
```diff
@@ -83,15 +98,30 @@ end
...
 end
-# TODO: Estimate periodic orbits of period `o` for discrete systems.
-# What's left is to create the Jacobian for higher iterates.
+
+# this can be derived from chain rule : J(f(g(x))) = Jf(g(x))*Jg(x)
+function to_root_J(Jf, ds::DeterministicIteratedMap, p, o::Int)
+    c = Diagonal(ones(typeof(current_state(ds))))
+    d = dimension(ds)
+    u -> begin
+        trajectory = Vector{Any}(undef, o) # has to be any to work with IntervalRootFinding
```
Datseris, line 114: Can you elaborate on this? Why do you need `any` instead of `typeof(current_state(ds))`?

src/stability/fixedpoints.jl
```diff
@@ -83,15 +98,30 @@ end
...
+        trajectory[1] = u
+        for i in 2:o
+            trajectory[i] = ds.f(trajectory[i-1], p, 0.0) # trajectory from DynamicalSystemsBase won't work because of IntervalRootFinding
+        end
+        reverse!(trajectory)
+        jacobians = map(x -> Jf(x, p, 0.0), trajectory)
```
Datseris, line 120: I am not sure I understand the necessity of this `trajectory`. Why make a `trajectory` container and loop over, instead of initializing a Jacobian matrix and multiplying in the loop? `J *= Jf(x, p, 0)` and `x = ds.f(x, p, 0)` would do the trick in the loop.

test/stability/fixedpoints.jl
```diff
@@ -38,6 +38,27 @@ end
...
+    tol = 1e-8
+    for x0 in fp
+        set_state!(ds, x0)
+        step!(ds, o)
+        xn = current_state(ds)
+        if isapprox(x0, xn; atol = tol)
+            @test true
+        else
+            @test false
+        end
```
Datseris, line 59:
```suggestion
        @test isapprox(x0, xn; atol = tol)
```
this is much simpler

src/stability/fixedpoints.jl
JonasKoziorek: When `to_root_J` is used in the `IntervalRootFinding.roots`, lambda function input `u` is of `IntervalBox{n, Float64}` type. To avoid `any`, `typeof(u.v)` can be used but the lambda function won't work if `u` is `SVector`. I will change `any` to `typeof(u.v)` because `to_roo_J` is meant to be used primarily in `roots`.

JonasKoziorek: I fixed it.

src/stability/fixedpoints.jl
JonasKoziorek: Good idea. I simplified the function.

src/stability/fixedpoints.jl
```diff
@@ -83,15 +98,28 @@ end
...
-# TODO: Estimate periodic orbits of period `o` for discrete systems.
-# What's left is to create the Jacobian for higher iterates.
+
+# this can be derived from chain rule : J(f(g(x))) = Jf(g(x))*Jg(x)
+function to_root_J(Jf, ds::DeterministicIteratedMap, p, o::Int)
+    c = Diagonal(ones(typeof(current_state(ds))))
+    d = dimension(ds)
+    u -> begin
+        J = Jf(u, p, 0.0)
+        x = ds.f(u, p, 0.0)
```
Datseris, line 115: Note that this code doesn't work if the map is in-place. Can you please add two multiple dispathc functions on `ds::DeterministicIteratedMap{false}` and `ds::DeterministicIteratedMap{true}` (false = out of place = what you have now).

Please also add a test for hte out of place configuration.

Datseris: for  the in place version you need to do `ds.f(u, uprev, p, 0.0)` to update the current state based on previous state. Jacobian function is also in-place I think.

JonasKoziorek: Function `fixedpoints` supports only oop systems.

https://github.com/JuliaDynamics/ChaosTools.jl/blob/2602883f4e4fa08ad49484e273ec9ca05132efe4/src/stability/fixedpoints.jl#L55-L57

I don't understand what you mean by a test for out of place configuration. All the systems in `test/stability/fixedpoint.jl` are oop.
I think that `IntervalRootFinding.roots` doesn't work with in-place systems.

PR #330, opened 2024-03-26
Changelog update

JonasKoziorek:
As promised, I have updated the changelog and bumped the minor version.


Datseris (approved):

Datseris:
Very good! How is the GSOC application progressing?

Issue #337, opened 2024-10-13
Non-autonomous version of Lyapunov exponent (EADP)


rusandris:
In non-autonomous systems with parameter drift, long time averages are less useful to assess chaoticity.

There are several different approaches to address this, one is related to the ensemble approach. Here, a new quantity called the Ensemble-averaged pairwise distance (EAPD) is used. An analog of the classical largest Lyapunov exponent can be extracted from the EAPD) function: the local slope can be considered an instantaneous Lyapunov exponent.

See for example [here](https://www.sciencedirect.com/science/article/pii/S0370157324003235?via%3Dihub) , section 3.3 :
> The function of the EAPD in continuous time is defined as
$\rho(t) = \langle ln r(t) \rangle$
where the average is taken over an ensemble of trajectories. To any member of the ensemble a test particle is ordered at an initial distance $r_0$ . Quantity $r(t)$ is the dimensionless phase space distance between a test particle and an ensemble member at time $t$.

I was wondering if we could use `ParallelDynamicalSystem` for handling large ensembles of trajectories of the drifting system.
The test particles would just be a copy of the ensemble (can be even inside the same `ParallelDynamicalSystem` or separately).  For estimating the slope, we could use the already existing tools from FractalDimensions.jl [FractalDimensions.slopefit](https://juliadynamics.github.io/DynamicalSystemsDocs.jl/fractaldimensions/dev/#FractalDimensions.slopefit).

As far as I know, there isn't a function for this right now. Let me know if this is a good idea, and whether ChaosTools is the right package. (I've seen [CriticalTransitions](https://juliadynamics.github.io/CriticalTransitions.jl/dev/) has some functionality for rate-dependent phenomena but seems more niche).

Datseris:
Isn't this what https://juliadynamics.github.io/DynamicalSystemsDocs.jl/chaostools/stable/lyapunovs/#ChaosTools.local_growth_rates does? If you pass in a drifting system, does it work? 

(if yes, we can add it to the docs as an example and also mention in the function docstring it can be used for EADP)

rusandris:
> If you pass in a drifting system, does it work?

Unfortunately not, or at least not directly. To calculate EAPD at each period (time step), one needs to make sure that all ensemble members are stepped at the same time and that states as well as time are updated between distance calculations. 
[...]

PR #345, opened 2025-03-28
Instantaneous Lyapunov exponent for systems with parameter drift

rusandris:
This is a non-autonomous version of Lyapunov exponent for systems with parameter drift, based on [Dániel Jánosi, Tamás Tél, 2024](https://www.sciencedirect.com/science/article/pii/S0370157324003235?via%3Dihub) .

In non-autonomous systems with parameter drift, long time averages are less useful to assess chaoticity.

There are several different ways to address this, one is related to the ensemble approach. 
Here, a new quantity called the Ensemble-averaged pairwise distance (EAPD) is used. An analog of the classical largest Lyapunov exponent can be extracted from the EAPD function `ρ(t)` : the local slope can be considered an instantaneous Lyapunov exponent.

This was previously discussed in #337 .

Example (low-precision version of Fig.11 a) from the article above):
```julia

using DynamicalSystems
using LinearAlgebra
using StatsBase
using Plots,LaTeXStrings


function duffing_drift(u0 = [0.1, 0.25]; ω = 1.0, β = 0.2, ε0 = 0.4, α=0.00045)
   return CoupledODEs(duffing_drift_rule, u0, [ω, β, ε0, α])
end

@inbounds function duffing_drift_rule(x, p, t)
           ω, β, ε0, α = p
           dx1 = x[2]
           dx2 = (ε0+α*t)*cos(ω*t) + x[1] - x[1]^3 - 2β * x[2]
           return SVector(dx1, dx2)
end

duffing = duffing_drift()
duffing_map = StroboscopicMap(duffing,2π)

init_states = randn(1000,2) #initalize ensemble
ρ,times = EAPD(duffing_map,init_states,100;Ttr=20,sliding_param_rate_index=4); #calc EAPD
lyapunov_instant(ρ,times;interval=1:20) #slope of first 20 time steps

pl = plot(times,ρ,ylabel=L"\rho(t)",lc=:gray10,lw=2,xlabel=L"t",legend=false,guidefontsize=20,dpi=300)

```

### DIFF
--- a/src/chaosdetection/lyapunovs/EAPD.jl
+++ b/src/chaosdetection/lyapunovs/EAPD.jl
@@ -1,14 +1,42 @@
 export ensemble_averaged_pairwise_distance,lyapunov_instant
 
-#slopefit to ρ(t) curve -> instantaneous Lyapunov exponent
+"""
+    lyapunov_instant(ρ,times;interval=1:length(times)) -> λ(t)
+
+Convenience function that calculates the instantaneous Lyapunov exponent by taking the slope of 
+the ensemble-averaged pairwise distance function `ρ` wrt. to the saved time points `times` in `interval`.
+"""
 function lyapunov_instant(ρ,times;interval=1:length(times))
-    _,s = linreg(times[interval], ρ[interval]) #return estimated slope and confidence intervals 
+    _,s = linreg(times[interval], ρ[interval]) #return estimated slope  
     return s
 end
 
-#Ensemble-averaged pairwise distance (EAPD) -> ρ(t),t
+"""
+    ensemble_averaged_pairwise_distance(ds,init_states::StateSpaceSet,T;kwargs...) -> ρ,t
+
+Calculate the ensemble-averaged pairwise distance function `ρ` for non-autonomous dynamical systems 
+with a time-dependent parameter. Time-dependence is assumed to be linear (sliding). To every member 
+of the ensemble `init_states`, a perturbed initial condition is assigned. `ρ(t)` is the natural log 
+of phase space distance between the original and perturbed states averaged over all pairs, calculated 
+for all time steps up to `T`. 
+
+This function implements the method described in 
+https://doi.org/10.1016/j.physrep.2024.09.003.
+
+## Keyword arguments
+
+* `sliding_param_rate_index = 0`: index of the sliding parameter
+* `initial_params = deepcopy(current_parameters(ds))`: initial parameters 
+* `Ttr = 0`: transient time used to evolve initial states to reach 
+    initial autonomous attractor (without sliding)
+* `perturbation = perturbation_uniform`: if given, it should be a function `perturbation(ds,ϵ)`,
+   which outputs perturbed state vector of `ds` (preferrably `SVector`). If not given, a normally distributed 
+   random perturbation with norm `ϵ` is added.
+*  `Δt = 1`: step size 
+*  `ϵ = sqrt(dimension(ds))*1e-10`: initial distance between pairs of original and perturbed initial conditions 
+"""
 function ensemble_averaged_pairwise_distance(ds,init_states::StateSpaceSet,T;sliding_param_rate_index=0,
-    initial_params = deepcopy(current_parameters(ds)),Ttr=0,perturbation=perturbation_uniform,Δt = 1,ϵ=sqrt(2)*1e-10)
+    initial_params = deepcopy(current_parameters(ds)),Ttr=0,perturbation=perturbation_uniform,Δt = 1,ϵ=sqrt(dimension(ds))*1e-10)
 
 	set_parameters!(ds,initial_params)
     N = length(init_states)
@@ -51,7 +79,6 @@ function ensemble_averaged_pairwise_distance(ds,init_states::StateSpaceSet,T;sli
     #set back time to t0 = 0
     reinit!(pds,current_states(pds))
 
-	#function barrier here?
     #calculate EAPD for each time step
     ensemble_averaged_pairwise_distance!(ρ,times,pds,T,Δt)
     return ρ,times
--- a/src/chaosdetection/lyapunovs/EAPD.jl
+++ b/src/chaosdetection/lyapunovs/EAPD.jl
@@ -1,25 +1,14 @@
 export ensemble_averaged_pairwise_distance,lyapunov_instant
 
 """
-    lyapunov_instant(ρ,times;interval=1:length(times)) -> λ(t)
-
-Convenience function that calculates the instantaneous Lyapunov exponent by taking the slope of 
-the ensemble-averaged pairwise distance function `ρ` wrt. to the saved time points `times` in `interval`.
-"""
-function lyapunov_instant(ρ,times;interval=1:length(times))
-    _,s = linreg(times[interval], ρ[interval]) #return estimated slope  
-    return s
-end
-
-"""
     ensemble_averaged_pairwise_distance(ds, init_states::StateSpaceSet, T, pidx;kwargs...) -> ρ,t
 
 Calculate the ensemble-averaged pairwise distance function `ρ` for non-autonomous dynamical systems 
-with a time-dependent parameter, using the metod described by [^Jánosi, Tél]. Time-dependence is assumed to be a linear drift. The rate of change
+with a time-dependent parameter, using the metod described by [^Jánosi,Tél2024]. Time-dependence is assumed to be a linear drift. The rate of change
 of the parameter needs to be stored in the parameter container of the system `p = current_parameters(ds)`,
 at the index `pidx`. In case of autonomous systems (with no drift), `pidx` can be set to any index as a dummy.
 To every member of the ensemble `init_states`, a perturbed initial condition is assigned.
-`ρ(t)` is the natural log of phase space distance between the original and perturbed states averaged 
+`ρ(t)` is the natural log of state space distance between the original and perturbed states averaged 
 over all pairs, calculated for all time steps up to `T`. 
 
 
@@ -39,11 +28,11 @@ over all pairs, calculated for all time steps up to `T`.
 In non-autonomous systems with parameter drift, long time averages are less useful to assess chaoticity.
 Thus, quantities using time averages are rather calculated using ensemble averages. Here, a new 
 quantity called the Ensemble-averaged pairwise distance (EAPD) is used to measure chaoticity of 
-the snapshot attractor/ phase space object traced out by the ensemble [^Jánosi, Tél].
+the snapshot attractor/ state space object traced out by the ensemble [^Jánosi, Tél].
 
 To any member of the original ensemble (`init_states`) a close neighbour (test) is added at an initial distance `ϵ`. Quantity `d(t)` is the 
-dimensionless phase space distance between a test particle and an ensemble member at time t .
-If `init_states`` are randomly initialized (far from the attractor at the initial parameter), and there's no transient, 
+state space distance between a test particle and an ensemble member at time t .
+If `init_states` are randomly initialized (far from the attractor at the initial parameter), and there's no transient, 
 the first few time steps cannot be used to calculate any reliable averages.
 The function of the EAPD `ρ(t)` is defined as the average logarithmic distance between original and 
 perturbed initial conditions at every time step: `ρ(t) = ⟨ln d(t)⟩`
@@ -70,7 +59,7 @@ ds = DeterministicIteratedMap(drifting_logistic, [0.1], p)
 ρ,times = ensemble_averaged_pairwise_distance(ds,init_states,100,2;Ttr=1000)
 ```
 
-[^Jánosi, Tél]: Dániel Jánosi, Tamás Tél, Physics Reports **1092**, pp 1-64 (2024)
+[^Jánosi,Tél2024]: Dániel Jánosi, Tamás Tél, Phys. Rep. **1092**, pp 1-64 (2024)
 
 """
 function ensemble_averaged_pairwise_distance(ds,init_states::StateSpaceSet,T,pidx;
@@ -152,4 +141,16 @@ function perturbation_normal(ds,ϵ)
     D, T = dimension(ds), eltype(ds)
     p0 = randn(SVector{D, T})
     p0 = ϵ * p0 / norm(p0)  
+end
+
+"""
+    lyapunov_instant(ρ,times;interval=1:length(times)) -> λ(t)
+
+Calculates the instantaneous Lyapunov exponent by taking the slope of 
+the ensemble-averaged pairwise distance function `ρ` wrt. to the saved time points `times` in `interval`.
+
+"""
+function lyapunov_instant(ρ,times;interval=eachindex(times))
+    _,s = linreg(times[interval], ρ[interval]) #return estimated slope  
+    return s
 end
--- a/src/chaosdetection/lyapunovs/EAPD.jl
+++ b/src/chaosdetection/lyapunovs/EAPD.jl
@@ -1,0 +1,87 @@
+export EAPD,lyapunov_instant
+
+#slopefit to ρ(t) curve -> instantaneous Lyapunov exponent
+function lyapunov_instant(ρ,times;interval=1:length(times))
+    s,_,_ = slopefit(times[interval], ρ[interval]) #return estimated slope and confidence intervals 
+    return s
+end
+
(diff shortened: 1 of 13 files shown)


src/chaosdetection/lyapunovs/EAPD.jl
```diff
@@ -0,0 +1,87 @@
+export EAPD,lyapunov_instant
+
+#slopefit to ρ(t) curve -> instantaneous Lyapunov exponent
+function lyapunov_instant(ρ,times;interval=1:length(times))
+    s,_,_ = slopefit(times[interval], ρ[interval]) #return estimated slope and confidence intervals 
+    return s
+end
+
+#Ensemble-averaged pairwise distance (EAPD) -> ρ(t),t
+function EAPD(ds,init_states::Matrix,T;sliding_param_rate_index=0,initial_params = deepcopy(current_parameters(ds)),Ttr,Δt = 1,ϵ=sqrt(2)*1e-10)
```
Datseris, line 10: Why is `init_states::Matrix`? In other functions we don't specify groups of initial conditions as a matrix. Rather we use a `StateSpaceSet`.

Datseris: The function `EAPD` is missing a docstring. You can add the same discription you have used in the github issues. I think before I review it i'd like a docstring that describes what each input argument means.

Datseris: function should also be named `eapd`, or better yet, something more extensive and descriptive. Write it out in full instead of the abbreviation.

src/chaosdetection/lyapunovs/EAPD.jl
```diff
@@ -0,0 +1,87 @@
...
+        ρ += log.(norm(states[i] - states[N+i]))
+    end
+    return ρ/N 
+
+end
+
+function perturbation(ds,ϵ)
+    D, T = dimension(ds), eltype(ds)
+    Q0 = randn(SVector{D, T})
+    Q0 = ϵ * Q0 / norm(Q0)
```
Datseris, line 86: These are uniform in state space, right? This is always desirable?

Datseris (commented):
Overall this looks fine, but cna you add the docstring and test before I do a proper review?

Datseris:
can you also post the pictures as the artivcle is not open access?

Datseris:
@rusandris is this PR still a draft or ready to review? Furthermore, are the test failures a result of this PR or unrelated?

src/chaosdetection/lyapunovs/EAPD.jl
rusandris: Normally distributed along each dimension. This is not always desirable (for example when the state vector can only contain positive values). This is the default, but one can choose a custom perturbation function through the `perturbation` keyword argument.

rusandris:
> can you also post the pictures as the artivcle is not open access?

I'm not sure if I can just post the screenshots from the original article here, but anyways here are my version of Fig.11.
Fig11

The slope values came out to be only approximately equal to the ones in the article (for an ensemble of N=10000), but trend-wise it seems to be okay.

rusandris:
> @rusandris is this PR still a draft or ready to review? Furthermore, are the test failures a result of this PR or unrelated?

I've added a test for the regular Henon map (autonomous) just to check if the instantaneous Lyapunov exponent coincides with the regular Lyap exp from the Benettin method (this is basically a sanity check that the time- and ensemble averages are the same).

Oops, I forgot to include my `EAPD.jl` tests in `runtests.jl` so the failures are definitely coming from somewhere else. I might need to look into those as well.

src/chaosdetection/lyapunovs/EAPD.jl
```diff
@@ -0,0 +1,116 @@
...
+
+## Keyword arguments
+
+* `sliding_param_rate_index = 0`: index of the parameter that gives the rate of change of the sliding parameter
+* `initial_params = deepcopy(current_parameters(ds))`: initial parameters 
+* `Ttr = 0`: transient time used to evolve initial states to reach 
+    initial autonomous attractor (without sliding)
+* `perturbation = perturbation_uniform`: if given, it should be a function `perturbation(ds,ϵ)`,
+   which outputs perturbed state vector of `ds` (preferrably `SVector`). If not given, a normally distributed 
+   random perturbation with norm `ϵ` is added.
```
Datseris, line 34: If the default value is normally distributed, why do we call it uniform? :D , maybe rename to `perturbation_normal`.

src/chaosdetection/lyapunovs/EAPD.jl
```diff
@@ -0,0 +1,116 @@
...
+    ensemble_averaged_pairwise_distance(ds,init_states::StateSpaceSet,T;kwargs...) -> ρ,t
+
+Calculate the ensemble-averaged pairwise distance function `ρ` for non-autonomous dynamical systems 
+with a time-dependent parameter. Time-dependence is assumed to be linear (sliding). To every member 
+of the ensemble `init_states`, a perturbed initial condition is assigned. `ρ(t)` is the natural log 
+of phase space distance between the original and perturbed states averaged over all pairs, calculated 
+for all time steps up to `T`. 
+
+This function implements the method described in 
+https://doi.org/10.1016/j.physrep.2024.09.003.
```
Datseris, line 24: Please make this a proper citation like in the docstrings of other functions of the library.

src/chaosdetection/lyapunovs/EAPD.jl
Datseris, line 1: Throughout this whole file, whenever there is a comma `,` it should be followed by a space. So all `a,b,c` needs to become `a, b, c`.

src/chaosdetection/lyapunovs/EAPD.jl
```diff
@@ -0,0 +1,116 @@
...
+
+## Keyword arguments
+
+* `sliding_param_rate_index = 0`: index of the parameter that gives the rate of change of the sliding parameter
```
Datseris, line 28: There is a parameter index but there is no rate given as a keyword...? I am really confused. What does this keyword do?

src/chaosdetection/lyapunovs/EAPD.jl
```diff
@@ -0,0 +1,116 @@
...
+function lyapunov_instant(ρ,times;interval=1:length(times))
+    _,s = linreg(times[interval], ρ[interval]) #return estimated slope  
+    return s
+end
+
+"""
+    ensemble_averaged_pairwise_distance(ds,init_states::StateSpaceSet,T;kwargs...) -> ρ,t
+
+Calculate the ensemble-averaged pairwise distance function `ρ` for non-autonomous dynamical systems 
+with a time-dependent parameter. Time-dependence is assumed to be linear (sliding). To every member
```
Datseris, line 18: I don't understand what this function is supposed to do just by reading the docstring. We need to clarify some things: is the input `ds` expected to be autonomous or NON- ? And, is it mandatory that the system is in a parameter drift scenario, or there can be arbitrary non-autonomous behavior? If it can only be a drift, does the parameter drift happens inside `ds` or do you create it inside this new function? If it happens inside the `ds`, why do you need `sliding_param_rate_index`? And then you set this parameter to 0...? 

Then inside the function I see this is done so that the `ds` reaches an attractor. But this is not said in the function. 

I think this function is non-trivial and therefore warrants a `## Description` section.

Datseris (commented):
The tests are good, but I am missing some clarity on what this function does.

I am also wondering why you do this "convergence to attractor" step first. Is it described in the paper?

src/chaosdetection/lyapunovs/EAPD.jl
rusandris: That's funny, I guess it was uniform, then I changed it for some reason and forgot to rename the function

src/chaosdetection/lyapunovs/EAPD.jl
rusandris: You're right, I didn't explain the assumptions I'd made about the function itself.
I assumed that the parameter drift happens inside `ds`, meaning that the user already has a  linearly drifting parameter in the definition.  Let's take the system from above as an example here:

```julia
function duffing_drift(u0 = [0.1, 0.25]; ω = 1.0, β = 0.2, ε0 = 0.4, α=0.00045)
   return CoupledODEs(duffing_drift_rule, u0, [ω, β, ε0, α])
end

@inbounds function duffing_drift_rule(x, p, t)
           ω, β, ε0, α = p
           dx1 = x[2]
           dx2 = (ε0+α*t)*cos(ω*t) + x[1] - x[1]^3 - 2β * x[2]
           return SVector(dx1, dx2)
end
```
where ` ε` parameter is replaced by the drifting term `(ε0+α*t)`. I also assume that the drift rate `α` is in `p`. This is a simple way to accomplish drifting without extending the phase space.

> I am also wondering why you do this "convergence to attractor" step first. Is it described in the paper?

Yes. You need some time to reach the snapshot attractor. 
`In practice, the convergence time is the time after which the averages of the phase space variables taken with respect to the ensembles are practically identical, although time dependent`.
If `init_states` are randomly initialized (far from the attractor at the initial parameter), and there's no transient, the first few time steps cannot be used to calculate any reliable averages.

Because of this, there is a `Ttr` transient time option, which is spent on converging to the attractor at `ε0`, before the drift even starts happening. In other words, the drift needs to be switched off (drift rate `α = 0`) for this step. In order to switch it off, I need to know which parameter I need to set to `0`, this is why `sliding_param_rate_index`
exists. This was the simplest way I could think of, but maybe there's a better way.

Now, what if you want to use this function in an autonomous setting (like I did in the tests with the Henon system)? Then you don't have a drift rate among the parameters, there is no need to switch off anything, in which case the default is `sliding_param_rate_index=0` and the function still works.
Let me know if this makes sense. 

As far as I know there is no standard way to create a drifting system. (Maybe [here](https://juliadynamics.github.io/CriticalTransitions.jl/dev/) but I'm not sure.)

Datseris: It appears to me that CriticalTransitions.jl discusses a "drift" system in the opening image but nowhere in the stable docs is this mentioned or implemented. So we can safely ignore that. 

Right, your approach is fine with me: we expect a parameter index `pidx` (mandatory input argument, and not optional keyword argument), and simply explain how this process works. that it is expected that `p[pidx]` multiplies time `t` and it is the "rate" parameter. For the non-drifiting henon map test just make a 3-parameter henon map where the 3rd parmaeter is not used anywhere. Then give `pidx = 3`.

Do you mind going ahead with adding the descriptions etc in the source code/docstring?

Datseris: cc @reykboerner or @ryandeeley I don't know if you are still working on CriticalTransitions.jl and what is the status for the drifting systems.

Datseris: Oh and @rusandris you shouldn't mess with the whole parameter container. Just do `original_rate = current_parameter(ds, pidx)` and then after setting it to zero do `set_parameter!(ds, pidx, original_rate)` to restore it.

ryandeeley: > cc @reykboerner or @ryandeeley I don't know if you are still working on CriticalTransitions.jl and what is the status for the drifting systems.

Hi, the status regarding "drifting systems" is that @raphael-roemer and I have developed code for passing a user-defined ```ds::CoupledSDEs``` (describing the system of interest in an _autonomous_ form) and a ```rp::RateProtocol``` (describing the desired time-dependent forcing, or "drifting", one wishes to apply to the system) to a function ```RateSystem()```, which returns a ```CoupledSDEs``` describing the corresponding non-autonomous system.  In particular the function ```RateSystem()``` creates a modified drift function ```f(u,p,t)``` from the information of ```ds.integ.f``` and ```rp```. The code is available in the ```RateSystem``` branch of CriticalTransitions.jl (https://github.com/JuliaDynamics/CriticalTransitions.jl/tree/RateSystem), specifically in the file ```/test/ratesystem/RateSystemDraft2.jl```, and the discussion surrounding this can be found in this pull request: https://github.com/JuliaDynamics/CriticalTransitions.jl/pull/117. Documentation hasn't been written yet, I suppose in an ideal world that'll be done soon. I'm happy to help with anything if you'd like to implement this and cannot understand things just from the scripts.

Datseris: Thanks, that's good to know. @rusandris does this method you are implementing rely on a _linear_ drift? I.e., it mandates a term `p*t` with `t` time? Or time dependence can be arbitrary?

For now we will proceed as is in this PR with `pidx`. In the future, when the drifting system is available as public API we can use it instead.

rusandris: Yes, it relies on linear drift, but it can be generalized further in the future if needed.

Datseris (commented):
Thanks @rusandris ; Let's finish this PR then! You need to:

1. Add the new method to the documentation.
2. Add a changelog entry.
3. Increase the minor version of the Project.toml file!

src/chaosdetection/lyapunovs/EAPD.jl
```diff
@@ -0,0 +1,155 @@
...
+
+
+## Description
+In non-autonomous systems with parameter drift, long time averages are less useful to assess chaoticity.
+Thus, quantities using time averages are rather calculated using ensemble averages. Here, a new 
+quantity called the Ensemble-averaged pairwise distance (EAPD) is used to measure chaoticity of 
+the snapshot attractor/ phase space object traced out by the ensemble [^Jánosi, Tél].
+
+To any member of the original ensemble (`init_states`) a close neighbour (test) is added at an initial distance `ϵ`. Quantity `d(t)` is the 
+dimensionless phase space distance between a test particle and an ensemble member at time t .
```
Datseris, line 45: dimensionless? why? it has the dimensions of the state space.

src/chaosdetection/lyapunovs/EAPD.jl
```diff
@@ -0,0 +1,155 @@
...
+
+"""
+    ensemble_averaged_pairwise_distance(ds, init_states::StateSpaceSet, T, pidx;kwargs...) -> ρ,t
+
+Calculate the ensemble-averaged pairwise distance function `ρ` for non-autonomous dynamical systems 
+with a time-dependent parameter, using the metod described by [^Jánosi, Tél]. Time-dependence is assumed to be a linear drift. The rate of change
+of the parameter needs to be stored in the parameter container of the system `p = current_parameters(ds)`,
+at the index `pidx`. In case of autonomous systems (with no drift), `pidx` can be set to any index as a dummy.
+To every member of the ensemble `init_states`, a perturbed initial condition is assigned.
+`ρ(t)` is the natural log of phase space distance between the original and perturbed states averaged
```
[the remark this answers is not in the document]
Datseris: (principle of 1 word per concept)

rusandris:
> Thanks @rusandris ; Let's finish this PR then! You need to:
> 1. Add the new method to the documentation.

I've added to `lyapunovs.md`, and it looks fine for a first version (I managed to build it locally), but I couldn't figure out how to render paper references properly yet lol

Datseris:
Sorry for being late, I had some personal life problems. I am thinking whether it is possible to combine this with the `RateSystem` of https://github.com/JuliaDynamics/CriticalTransitions.jl/pull/117  which creates a `RateSystem` that could be used directly here. But to do this, I must first ask @rusandris : does this functionality here only work / make sense if the forcing is _linear_? Or the methodology would work with any kind of forcing?

rusandris:
>  does this functionality here only work / make sense if the forcing is linear? Or the methodology would work with any kind of forcing?

In the original review paper the parameter drift was chosen to be linear for simplicity. The method itself - calculating the ensemble averaged distance between trajectory pairs - doesn't depend on this linearity. You just have to be careful when taking the slope of `ρ(t)`, as it only can be considered as Lyapunov exponent if the distance between pairs is typically small. And that depends on your step size, not the functional form of the drift.  I might be wrong but  that's how I understand it right now.

Looking at the code, the function `ensemble_averaged_pairwise_distance` does use the assumption that the drift is linear. Particularly, `pidx` parameter is for the index of the drifting rate. But that can be generalized if needed, especially if we want to make this compatible with the `RateSystem`.

Datseris:
I'll merge this in for now, as the rate system will take a while more. Thanks @rusandris !

Datseris (approved):

